PR #4, opened 2019-12-27
Removed all trailing whitespaces

masoudd:
No actual code changed. Only excess whitespace removed.


stephanemagnenat:
Glob2's coding convention was to keep indentation on empty lines, which is removed by this PR. It seems that this style is a bit out of fashion recently, but I am not sure whether it is worth doing this change. Is this change of indentation in this PR on purpose?

masoudd:
Oh sorry I was not aware this was intentional. Never mind then.

Issue #1, opened 2019-12-26
cppcheck error: iterator 'bi' used after element has been erased


randomvoids:
From richelbilderbeek
https://bitbucket.org/giszmo/glob2/issues/8/cppcheck-error-iterator-bi-used-after

When running cppcheck (see [Travis build log](https://travis-ci.org/richelbilderbeek/globulation2/builds/564008690)), it detects the following errors:

```
[Building.cpp:2783]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Building.cpp:2788]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Building.cpp:2821]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Building.cpp:2830]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Building.cpp:2840]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Building.cpp:2860]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Building.cpp:2878]: (error) Shifting signed 32-bit value by 31 bits is undefined behaviour
[Game.cpp:1560] -> [Game.cpp:1559]: (error) Iterator 'bi' used after element has been erased.
```

Focusing on the last error:
`
[Game.cpp:1560] -> [Game.cpp:1559]: (error) Iterator 'bi' used after element has been erased.
`

These are the lines it is about:
```
for (std::list<Building *>::iterator bi=teams[ti]->virtualBuildings.begin(); bi!=teams[ti]->virtualBuildings.end(); ++bi)
    if ((*bi)->posX==x && (*bi)->posY==y)
    {
        teams[ti]->virtualBuildings.erase(bi);                      //1559
        teams[ti]->myBuildings[Building::GIDtoID((*bi)->gid)]=NULL; //1560
        delete *bi;
        found=true;
        break;
    }
```

Changing the order to this fixes the cppcheck error:
```
for (std::list<Building *>::iterator bi=teams[ti]->virtualBuildings.begin(); bi!=teams[ti]->virtualBuildings.end(); ++bi)
    if ((*bi)->posX==x && (*bi)->posY==y)
    {
        teams[ti]->myBuildings[Building::GIDtoID((*bi)->gid)]=NULL;
        delete *bi;
        teams[ti]->virtualBuildings.erase(bi);
        found=true;
        break;
    }
```
[...]

PR #6, opened 2019-12-28
Update scons, add Travis, qmake build and tests, fix #2

### LINKED ISSUE #2: scons BoolOption() deprecated
Moved from [Bitbucket](https://bitbucket.org/giszmo/glob2/issues/7) to here:

When building Globulation 2 using scons, I get (as can be seen in [a Travis CI build log](https://travis-ci.org/richelbilderbeek/globulation2/builds/563958498#L1591)) to error:

```
scons: warning: The BoolOption() function is deprecated; use the BoolVariable() function instead.
```

This build is run on Ubuntu 16.04 (Xenial), so BoolOption is deprecated already for some time.

Would it be OK to replace BoolOption by BoolVariable to get a cleaner build? If yes, I’ll submit a Pull Request

richelbilderbeek:
Hi Globulation 2 maintainers,

As per [this Issue](https://github.com/Globulation2/glob2/issues/2) I submit this Pull Request.

Changes:

 * scons now uses `makeVariable` and `Variable` instead of the deprecated `makeOptions` and `Options`
 * This is checked by a Travis CI script: upon a commit, the build process is started on a virtual machine, so one can see if a commit breaks a build
 * The build is also re-done by a Qt Creator project file, `src/globulation2.pro`
 * The build is also re-done by a Qt Creator project file, `src/globulation2_test.pro` that runs the cpptests in the test folder. Although these tests seem rather trivial, my new addition may upon up the path to write more tests in the future

I volunteer to maintain the `.travis.yml` and `.pro`(and `.pri`) files.

Enjoy :+1:


richelbilderbeek:
I've also added a Build badge in README.md.

The GitHub user with the name `Globulation2` does need to activate Travis CI [here](https://travis-ci.org/Globulation2/glob2) (you can log in from GitHub), as it is not activated yet.

scripts/update
```diff
@@ -0,0 +1,12 @@
...
+#
+# Update the original glob2 source code
+#
+# Usage:
+#
+#   ./scripts/update
+#
+
+cd glob2
+hg pull
```
stephanemagnenat: Why hg, as we transitioned to git?

scripts/install_deps
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
stephanemagnenat: Should we somehow name this script in relation with the type of system it installs deps on, such as `install_deb_deps`?

stephanemagnenat:
I am curious, what is the rationale for the qmake-based build parallel to the scons one?

scripts/update
richelbilderbeek: You are right, It's an old update script ...

scripts/install_deps
richelbilderbeek: I agree with the reasoning. Go ahead and rename, you're the boss here.

richelbilderbeek:
> I am curious, what is the rationale for the qmake-based build parallel to the scons one?

To show that a scons and qmake build are equally fine.

richelbilderbeek:
> Your PR currently has a lot of commits, some of them fixing things of previous commits, would it be ok for you to rebase them into a small sets of commits (e.g. one that fixes scons, one that adds Travis, one that adds Qt build, etc.)?

I would be open to do so, but I have no experience with rebasing (I feel it's not worth it). Due to this, I'd worry I'd mess up things. 

If you'd insist, I can just redo all the steps by hand (of course, I prefer to leave things as they are). If so, let me know, and I'll do :+1:

Thanks for your review!

scripts/update
stephanemagnenat: Ok, can you do the deletion, so that I do not delete the wrong thing?

stephanemagnenat:
I can do the rebase if you want. Note that it is actually not that spooky, there are a lot of [introductions online](https://www.freecodecamp.org/news/an-introduction-to-git-merge-and-rebase-what-they-are-and-how-to-use-them-131b863785f/).

That's great if you maintain the travis and the pro files.

test/TestsRunner.cpp
```diff
@@ -29,6 +29,10 @@
 #include <cppunit/ui/text/TestRunner.h>
 #include <cppunit/XmlOutputter.h>
 
+//Globulation2 has one global
+#include "GlobalContainer.h"
+GlobalContainer *globalContainer = new GlobalContainer;
```
stephanemagnenat: Why do we need to create the `GlobalContainer` here? To see whether it fails? A comment would be helpful!

richelbilderbeek: It resolves an unresolved external error. 

I felt a comment like `// make it link` a bit unhelpful. The comment I did write was a left-over from `// Globulation2 has one global :-(` (I prefer my locals), which I think is unhelpful as well. Go ahead and remove it :+1:

scripts/update
richelbilderbeek: Done :+1:

scripts/install_deps
richelbilderbeek: If you want, I'll rename :+1:

test/TestsRunner.cpp
richelbilderbeek: (of course, I'd be happy to remove it myself)

richelbilderbeek:
> That's great if you maintain the travis and the pro files.

Just to re-iterate: I'll happily do so :+1:

scripts/install_deps
randomvoids: Can you rename it please

richelbilderbeek: Done :+1:

stephanemagnenat:
@randomvoids for me this PR looks ok, if it is ok for you I can rebase to cleanup its messy history and merge.

randomvoids:
> > I am curious, what is the rationale for the qmake-based build parallel to the scons one?
> 
> To show that a scons and qmake build are equally fine.

Is there a reason why we need both though? I would prefer to not have to maintain 2 build systems if we can avoid it. Is qmake better than scons or ?

richelbilderbeek:
@randomvoids:

> Is there a reason why we need both though? I would prefer to not have to maintain 2 build systems if we can avoid it. Is qmake better than scons or ?

AFAICS, `qmake` is just as fine as `scons`. I used `qmake` because I am comfortable with it (unlike `scons`). Thanks to `qmake`, I could also setup a `qmake` project to run the tests in the `tests` folder (something the `scons` build does not (?) do). Additionally, I can open up the project in my favorite IDE called Qt Creator.

I can imagine the writer of the `scons` setup prefers scons and volunteers to maintain it, like I would enjoy maintaining the qmake project files.

There are some options that may be considered (and I volunteer to implement each any decision):
 * I can setup the Travis CI file to either `scons` or `qmake` depending on the branch name. 
 * Or perhaps more fun: I can setup the Travis CI file to use two compilers (GCC and clang) and use `scons` for the first and `qmake` for the second compiler.

&TLDR; both are equally fine, I am just more comfortable with qmake

stephanemagnenat:
A "problem" with qmake is that it adds a dependency on Qt, which feels a bit wrong just to use its build system. If it is in addition to scons, why not, if it is the main one, I would be careful before changing.



buildhttps://en.m.wiktionary.org/wiki/bikeshedding system.

richelbilderbeek:
> [...] qmake is that it adds a dependency on Qt, which feels a bit wrong just to use its build system. If it is in addition to scons, why not.

Cool! 

You may overlook some benefits though:

 * Adding a Qt Creator project file (those `.pro` files) allows people to use the Qt Creator IDE to develop on Globulation 2
 * As I am comfortable with `qmake`, there will be more Pull Requests filled with goodies. For example, adding `cppcheck` to the Travis build. Note that @randomvoids already made  [an Issue](https://github.com/Globulation2/glob2/issues/1) due to an error spotted by cppcheck.

Again: I agree to keep `scons` the primary build system and I do volunteer to maintain the Travis and `.pro`/`.pri` files.

richelbilderbeek:
@stephanemagnenat: how is the rebase going?

stephanemagnenat:
I split this PR into two, the SCons fix is in master, the rest [in that branch](https://github.com/stephanemagnenat/glob2/tree/richelbilderbeek-master).

While doing it, I encountered several questions:
- What is the purpose of the scripts you added, as if I'm not mistaken they are not used?
- There seems to be duplications, for example cppunit is built by autoconf both in `.travis.yml` and in `scripts/install_cppunit`, but also by Qt in `src/cppunit.pri`, I do not understand why. They make the whole system hard to understand because one does not know what is necessary or not.

I would like these questions solved because merging the rest of the PR, because otherwise we import hard-to-maintain stuff that would confuse people entering the project, as they confuse me.

richelbilderbeek:
> What is the purpose of the scripts you added, as if I'm not mistaken they are not used?

 * `check_dead_links`: checks for dead links within a repo
 * `install_cppunit`: installs CppUnit
 * `install_deb_deps`: installs the dependencies (the Debian packages, that is)

> There seems to be duplications, for example cppunit is built by autoconf both in .travis.yml and in scripts/install_cppunit

This is a temporary duplication: I supplied a stand-alone `.travis.yml` file. I naively thought that that would give me the path of least resistance getting my Pull Request accepted. Ideally, all three scripts are called from the `.travis.yml` file, which will help shorten it.

> There seems to be duplications, for example cppunit is built [...] in .travis.yml [...], but also by Qt in `src/cppunit.pri` 

There is not: `install_cppunit` installs CppUnit, `cppunit.pri` adds the files to the project.

richelbilderbeek:
@stephanemagnenat: what is the status of this PR?

richelbilderbeek:
@stephanemagnenat: what is the status of this PR?

stephanemagnenat:
I think it is up to the maintainer to take the decision on whether to accept an additional build system. I did apply the part fixing Scons in 51ebae206deacb6e962f63914326b9955bfab676.

richelbilderbeek:
I'll take my loss then, submitting a Pull Request usually is a more fun experience.

PR #13, opened 2020-01-06
Work around segfault in YOGServerFileDistributor::update.

ProgVal:


src/YOGServerFileDistributor.cpp
```diff
@@ -103,6 +103,10 @@ void YOGServerFileDistributor::update()
 			i->get<0>()->sendMessage(fileInfo);
 			i->get<2>() = 1;
 		}
+		else if(i->get<2>() == 0) {
```
stephanemagnenat, line 106: I do not understand why this prevents the segfault. What does cause the segfault?

ProgVal: I wrote this code years ago so I don't remember (not that I understood it back then either). But the next block access `chunks[i->get<2>()-1]`, so it looks like it would be an issue if `i->get<2>()` is 0.

stephanemagnenat: Ok, seems reasonable and clearly better than what we have now, so I'll merge that.

PR #15, opened 2020-01-06
Update dependency on boost.

ProgVal:
I did this years ago and don't remember why. It may have
to do with making the server work with Debian 8 or 9.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -130,14 +130,24 @@ def configure(env, server_only):
         print "Could not find libboost_thread or libboost_thread-mt or boost/thread/thread.hpp"
         missing.append("libboost_thread")
     env.Append(LIBS=[boost_thread])
+
+    boost_system = ''
+    if conf.CheckLib("boost_system"):
+        boost_system="boost_system"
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -285,7 +295,7 @@ def main():
     env.Append(CPPPATH=['#libusl/src', '#'])
     env.Append(CXXFLAGS=' -Wall')
     env.Append(LINKFLAGS=' -Wall')
-    env.Append(LIBS=['SDL_net'])
+    env.Append(LIBS=['SDL_net', 'pthread'])
```
stephanemagnenat, line 298: I'm worried that this would break compilation on Windows.

stephanemagnenat:
I will wait for the SDL2 port to land, and then update Scons to Python3, and then see whether this change is still relevant.

stephanemagnenat:
@ProgVal I merged the update to Python3 and now Glob2 compiles on Ubuntu 20.04. Do you think this change is still necessary or desired?

ProgVal:
Just tried it and it builds and seems to work, thanks :)

The current git master is now running on the YOG server.

PR #16, opened 2020-01-06
Compile with -O3 instead of -O2.

ProgVal:


SConstruct
```diff
@@ -283,20 +283,20 @@ def main():
 
     env.Append(CPPPATH=['#libgag/include', '#'])
     env.Append(CPPPATH=['#libusl/src', '#'])
-    env.Append(CXXFLAGS=' -Wall')
+    env.Append(CXXFLAGS=' -Wall -O3 -fPIC')
```
stephanemagnenat, line 286: Why having this flag in any case, rather then below in the release only/

PR #30, opened 2020-09-03
Upgrade to SDL 2.0

### LINKED ISSUE #10: Upgrade to SDL 2
We should upgrade to the latest version

Quipyowert2:
Fixes #10.

### DIFF
--- a/libgag/include/KeyPress.h
+++ b/libgag/include/KeyPress.h
@@ -28,7 +28,7 @@ class KeyPress
 {
 public:
 	///Construct a KeyPress
-	KeyPress(SDL_keysym key, bool pressed);
+	KeyPress(SDL_Keysym key, bool pressed);
 	
 	///Clone a KeyPress except changing the pressed value
 	KeyPress(KeyPress key, bool pressed);
--- a/libgag/include/SDLGraphicContext.h
+++ b/libgag/include/SDLGraphicContext.h
@@ -316,7 +316,10 @@ namespace GAGCore
 		//! the minimum acceptable resolution
 		int minW, minH;
 		//! the pointer for iterating through mode list
-		SDL_Rect **modes;
+		SDL_DisplayMode **modes;
+		int totalmodes;
+		SDL_Window *window;
+		SDL_Renderer *sdlrenderer;
 		friend class DrawableSurface;
 		//! option flags
 		Uint32 optionFlags;
@@ -388,7 +391,7 @@ namespace GAGCore
 		//! Set the minimum acceptable resolution
 		virtual void setMinRes(int w = 0, int h = 0);
 		//! Set the caption of the window
-		virtual void setCaption(const char *title, const char *icon) { SDL_WM_SetCaption(title, icon); }
+		virtual void setCaption(const char *title, const char *icon) { SDL_SetWindowTitle(window, title); }
 		//! Begin listing of acceptable video mode, *not thread-safe*
 		virtual void beginVideoModeListing(void);
 		//! Get the next acceptable video mode in w,h, return false if end of list, *not thread-safe*
--- a/libgag/include/SDLGraphicContext.h
+++ b/libgag/include/SDLGraphicContext.h
@@ -315,8 +315,9 @@ namespace GAGCore
 	protected:
 		//! the minimum acceptable resolution
 		int minW, minH;
-		//! the pointer for iterating through mode list
 		SDL_DisplayMode **modes;
+		//! the pointer for iterating through mode list
+		SDL_DisplayMode **modesitr;
 		int totalmodes;
 		SDL_Window *window;
 		SDL_Renderer *sdlrenderer;
--- a/libgag/src/GraphicContext.cpp
+++ b/libgag/src/GraphicContext.cpp
@@ -18,7 +18,6 @@
 */
 
 #include <GraphicContext.h>
-#include <../../src/GlobalContainer.h>
 #include <Toolkit.h>
 #include <FileManager.h>
 #include <SupportFunctions.h>
@@ -55,7 +54,6 @@
 
 //extern "C" { SDL_PixelFormat *SDL_AllocFormat(int bpp, Uint32 Rmask, Uint32 Gmask, Uint32 Bmask, Uint32 Amask); }
 
-extern GlobalContainer *globalContainer;
 namespace GAGCore
 {
 	// static local pointer to the actual graphic context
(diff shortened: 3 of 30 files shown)


libgag/include/KeyPress.h
```diff
@@ -28,7 +28,7 @@ class KeyPress
 {
 public:
 	///Construct a KeyPress
-	KeyPress(SDL_keysym key, bool pressed);
+	KeyPress(SDL_Keysym& key, bool pressed);
```
stephanemagnenat, line 31: Why not passing by value, afaik `SDL_keysym` is small, no? If passed by reference, I would use a const reference.

libgag/include/SDLGraphicContext.h
```diff
@@ -296,7 +296,7 @@ namespace GAGCore
 	//! A GraphicContext is a DrawableSurface that represent the main screen of the application.
 	class GraphicContext:public DrawableSurface
 	{
-		static const bool verbose = false;
+		static const bool verbose = true;
```
stephanemagnenat, line 299: Why this change? Is it a left-over debug?

libgag/src/GraphicContext.cpp
```diff
@@ -1922,12 +1928,32 @@ namespace GAGCore
...
+			totalmodes += nummodes;
+		}
+		modes = new SDL_DisplayMode*[totalmodes+1];
+		modes[totalmodes] = 0;
+		for (int i =0;i < displays;i++) {
+			nummodes = SDL_GetNumDisplayModes(i);
+			for (int j = 0;j < nummodes;j++) {
+				if (SDL_GetDisplayMode(i, j, &mode))
+					std::cerr << "SDL_GetDisplayMode failed: " << SDL_GetError() << std::endl;
+				modes[i+j] = &mode;
```
stephanemagnenat, line 1948: That looks incorrect to me, it should rather be something like:
`int base_i = 0` in 1942,
then
`base_i += nummodes` in 1949
and
`modes[base_i + j]` here.

libgag/src/GraphicContext.cpp
```diff
@@ -1922,12 +1928,32 @@ namespace GAGCore
...
+			}
+		}
+		//modes = SDL_ListModes(NULL, SDL_FULLSCREEN|SDL_HWSURFACE);
 	}
 
 	bool GraphicContext::getNextVideoMode(int *w, int *h)
 	{
-		if (modes && (modes != (SDL_Rect **)-1))
+		if (modes && (modes != (SDL_DisplayMode **)-1))
```
stephanemagnenat, line 1956: We are leaking modes in this function, but we were already before... to be fixed after this PR is merged.

stephanemagnenat (changes requested):
Thank you very much for the upgrade! If you are available, please correct the small issue I pointed out. If you are not available, please tell us so that someone else can do it!

libgag/include/KeyPress.h
Quipyowert2: I changed it back to pass-by-value.

libgag/include/SDLGraphicContext.h
Quipyowert2: Yes, it was leftover from debugging.

libgag/src/GraphicContext.cpp
Quipyowert2: Yeah it's definitely incorrect the way I wrote it because it puts the address of a stack variable into a member variable. When getVideoModeListing returns the pointer is invalid. Also all the modes are being set to the same pointer.

libgag/src/GraphicContext.cpp
Quipyowert2: Fixed by using a different pointer for iterating.

stephanemagnenat:
Thanks a lot for the quick reaction! The PR does not compile in the CI, I think because the file `debian/control` still has the old SDL dependencies and is used to install the necessary packages.

libgag/src/GraphicContext.cpp
stephanemagnenat: Ok! In the future, we could probably simplify this whole mode getting code by using a `std::deque<SDL_DisplayMode>`, emplacing them on the back and eating them on the front. That would solve the leak (#31). But for now your changes look good!

libgag/src/GraphicContext.cpp
```diff
@@ -18,6 +18,7 @@
 */
 
 #include <GraphicContext.h>
+#include <../../src/GlobalContainer.h>
```
stephanemagnenat, line 21: This include breaks modularity, because `libgag` is not designed to depend on Globulation. I see that you use that for some wheel property. One proper way to do it is to put a `static bool` in `Screen` for example.

libgag/src/GraphicContext.cpp
```diff
@@ -240,13 +242,17 @@ namespace GAGCore
 	SDL_Surface *DrawableSurface::convertForUpload(SDL_Surface *source)
 	{
 		SDL_Surface *dest;
+		if (!_gc)
+		{
+			_gc = globalContainer->gfx;
```
stephanemagnenat, line 247: Why do we need this? As described above, `libgag` should not depend on the `globalContainer`, which is Glob2-specific.

libgag/src/GraphicContext.cpp
Quipyowert2: I removed this include and it compiled fine but when I add a `static bool scrollWheelEnabled` and change the scroll wheel code to use it I get an ```undefined reference to `GAGGUI::Screen::scrollWheelEnabled'``` error.

libgag/src/GraphicContext.cpp
Quipyowert2: I think that part was to stop the game crashing if `_gc` was NULL.

Quipyowert2:
With this pull request, the game works with and without OpenGL mode checked in my virtual machine (openSUSE 15.1) but the non-OpenGL mode in WSL1 (openSUSE 15.2 with VcXsrv) just prints a bunch of `Xlib:  extension "MIT-SHM" missing on display ":0".` and shows a blank screen. OpenGL in WSL1 works fine.

Quipyowert2:
I changed the dependencies in `debian/control` to the SDL2 dependencies and Travis is still fetching the 1.2 dependencies.

stephanemagnenat:
I do not know why Travis is still fetching the old dependencies, as I see no other place where the SDL package to be installed is listed. @krichter722 any idea?

Once this is solved, I'll merge that in. @Quipyowert2 is it ok for you if I squash these changes into a single commit?

stephanemagnenat:
My trial in #39 did not work because the packages are installed _outside_ the Docker. @krichter722 do you know how to install dependencies in the docker and not outside?

stephanemagnenat:
I think I got it, through the `DEPS` variable.

stephanemagnenat:
Ok, so the compilation in CI is now fixed in #39, but the port does not work for me (see #40), so we need fixing before merging otherwise we'll break master.

stephanemagnenat:
This has been merged as port of #39.

PR #33, opened 2021-02-15
Add upcoming Ubuntu 21.04 to Travis builds

krichter722:


krichter722:
I guess it's an issue with massive requests from the same IP addresses of Travis CI or a lot of builds of glob2 within 6 hours. These limitations didn't exist until the end of 2020. You can always restart these builds with one click and it'll be one or two jobs only most of the times. You should watch this issues and see how big the pain is over time.

Unfortunately the Travis Docker integration is rather poor imo. I like GitLab and their CI concept for this reason. Maybe you should check how much you get from CI integration into the code hosting platform and evaluate after a certain time if you want to migrate. Feel free to mention me.

PR #39, opened 2021-02-17
Upgrade to SDL 2.0, with bug fixes

stephanemagnenat:
This is #30 with an attempt to let Travis install SDL2.


stephanemagnenat:
This does not work because the packages are installed outside the Docker.

PR #45, opened 2021-04-11
Check for SDL2 versions of SDL libraries.

Quipyowert2:
scons was failing if it couldn't find SDL1.2 libraries even
when SDL2 libraries were available. Since Globulation 2 uses SDL2 now,
require SDL2 libraries to remove the dependency on SDL1.2.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -81,18 +81,18 @@ def configure(env, server_only):
         missing.append("CXX compiler")
 
     #Simple checks for required libraries
-    if not server_only and not conf.CheckLib("SDL"):
-        print("Could not find libSDL")
-        missing.append("SDL")
-    if not server_only and not conf.CheckLib("SDL_ttf"):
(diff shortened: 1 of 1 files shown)


stephanemagnenat:
Do you have any idea why it fails on 21.04? We could also switch that version to 22.04 now...

Quipyowert2:
Just before the last couple of lines, the [travis log](https://app.travis-ci.com/github/Globulation2/glob2/jobs/497612873) says "gpgv, gpgv2 or gpgv1 required for verification, but neither seems installed". I searched for that string and found [this stack overflow post](https://stackoverflow.com/questions/66319610/gpg-error-in-ubuntu-21-04-after-second-apt-get-update-during-docker-build). According to the accepted answer on that question, it is [a bug in Ubuntu 21.04](https://bugs.launchpad.net/ubuntu/+source/libseccomp/+bug/1916485). A fix for this bug has been released.

PR #46, opened 2021-04-22
Adjusted language to be inclusive.

onpon4:
This switches out the word "crazy" for the more precise "complicated" since "crazy" has historical origins in ableism, and also changes references to a unit and to an enemy player from he/him to [singular they/them](https://en.wikipedia.org/wiki/Singular_they) pronouns (which are gender-neutral). These changes are minor but contribute to making the game as inclusive for women and neurodivergent people as possible without sacrificing anything.

(Side note, this is an awesome game and it's great to see that its development at least occasionally sees activity. Thank you for working on it! :slightly_smiling_face: It's a real refreshing take on the RTS genre.)

🕵️


Giszmo:
NACK: can we close this please. Let's keep politics out of this repo.

onpon4:
I don't see how excluding women with your language (by implying that everyone who plays video games is a man by default) is less "political" than simply being universally inclusive and using the traditional singular "they", but if people like Giszmo are representative of the Globulation 2 development team, I will rescind my comment about it being an "awesome game".

🦇

Shayan-To:
I have to opinion about this being a political matter or not, but I think that:

- "It's getting complicated" does not convey the same message as "It's getting crazy", and doesn't seem to me a good change to make.
- But the changes from "he" to "they" in fact make the messages better, and I think would be an improvement.

stephanemagnenat:
I agree with @Shayan-To. Actually I suspect that there are more grammar mistakes in the tutorial.

Also, a note of history: Globulation was started by two French speaking persons (I'm one of them), at a time when our English was not great. Later a diverse community of people from various language and culture contributed to the game and its translations.

For example, regarding the naming of units. In French we use the feminine (ouvrières, guerrières, exploratrices) because we thought of globules as analogue to ants and ant (fourmi) is a feminine word in French (contrary to ants (and English), the French language has no neutral gender). No French men (or women) ever complained of feeling not included by this naming. I also find it surprising that a player would identify to an individual unit, as the whole point of Globulation is that the player is the "overmind" leading the swarm, not a unit.

So, I'm happy about the change from "he" to "they" because of clarity, not because I believe that having gendered units is not inclusive.

For the other change, I suggest to have it in another PR/issue if people think the status quo is not a clear wording, as my English is not good enough to judge that point (the French version that has a slightly different phrasing feels ok to me).

odisseus:
In my opinion, the globule is an "it". It doesn't have any sex or gender, at least for the purposes of the gameplay. 

By the way, an ant is also an "it" in the English language, despite the fact that worker ants are female.

Quipyowert2:
Re: crazy. I think the word _chaotic_ or _hectic_ would fit better in that sentence; chaotic being more busy than hectic, which simply means "very busy". Chaotic is also an alignment in D&D though, being the opposite of lawful. There is also the word _bustling_.

kylelutze:
merged PR 167 that closes what we'll do.

PR #59, opened 2022-09-08
Create codeql.yml

stephanemagnenat:
This PR adds CodeQL code scanning to the repo, but I do not find the results that great, so I'm not sure it is worth merging. Input is welcome!


kylelutze:
more work to be done here or close this at this point?

stephanemagnenat:
I think we can close, I was not convinced by the experiment.

PR #63, opened 2022-10-07
Loslessly optimize png assets

anthonyryan1:
Using Efficient Compression Tool, we can optimize png images which will improve disk size and loading times.


anthonyryan1:
Apparently this is too many images for github to render in the commit.

I manual inspection should show they're all identical to the previous images but significantly smaller. I've playtested through the game before and after the change and found no visible difference.

stephanemagnenat:
Thanks, I validated with a script that content have not changed, merging...

PR #64, opened 2022-11-27
Implement resizable GUI

Quipyowert2:
The resizing is a little buggy but it mostly works. One bug I noticed was that the buttons on title screen get mangled during resizing or the button disappears and only the text of the button shows. Another bug is the game sometimes flashes red when resizing, not sure why.

The way this works is the main thread spawns a secondary thread and then opens the window, listens for events and adds to an events queue until told to stop. The second thread handles events, logic and rendering. When exiting the game, the second thread tells main thread to quit the event loop. Then the main thread joins the secondary thread, closes the window, and exits.

SDL_PollEvent will block the thread calling it during resize on Windows, so that's why I had to start another thread to handle everything else. I changed SDL_PollEvent to `el->poll` so that the resize doesn't block the event loop.

Just tested this in a Linux VM and it crashes in SDL_GL_SwapWindow. It works on Windows 10 though.

TODO:
- [x] Refactor the drawing code in `MapEdit::run` into a separate draw method.
- [x] Implement support for painting MapEdit and Screen objects in EventListener.
- [x] Fix textures turning white when resizing
  - [x] Fix globules flickering
  - [x] Fix cursor flickering
  - [x] Fix tutorial message flickering
- [ ] Fix bottom of window not being drawn sometimes after resizing
- [ ] Fix blank frame sometimes being drawn on title screen and in tutorial
- [ ] Cursor and clouds animations are too fast when resizing the window.
- [ ] Optimize drawing routines

### DIFF
--- a/glob2.vcxproj
+++ b/glob2.vcxproj
@@ -262,6 +262,7 @@
     <ClCompile Include="gnupg\sha1.c" />
     <ClCompile Include="libgag\src\BinaryStream.cpp" />
     <ClCompile Include="libgag\src\CursorManager.cpp" />
+    <ClCompile Include="libgag\src\EventListener.cpp" />
     <ClCompile Include="libgag\src\FileManager.cpp">
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='Debug SDL|Win32'">ZLIB_WINAPI;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='Debug SDL|x64'">ZLIB_WINAPI;%(PreprocessorDefinitions)</PreprocessorDefinitions>
@@ -507,6 +508,7 @@
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="libgag\include\DX9GraphicContext.h" />
+    <ClInclude Include="libgag\include\EventListener.h" />
     <ClInclude Include="libgag\include\FileManager.h" />
     <ClInclude Include="libgag\include\GAG.h" />
     <ClInclude Include="libgag\include\GAGSys.h" />
--- a/libgag/include/CursorManager.h
+++ b/libgag/include/CursorManager.h
@@ -67,6 +67,7 @@ namespace GAGCore
 		CursorManager();
 		//! Load the cursor sprites
 		void load(void);
+		void reinitTextures(void);
 		//! Select the next type given the mouse position
 		void nextTypeFromMouse(DrawableSurface *ds, int x, int y, bool button);
 		//! Manually set the next type
--- a/libgag/include/EventListener.h
+++ b/libgag/include/EventListener.h
@@ -10,9 +10,12 @@ public:
 	EventListener(GraphicContext* gfx);
 	void run();
 	bool isRunning();
+	int poll(SDL_Event* e);
+	static EventListener *instance();
 	~EventListener();
 private:
 	GraphicContext* gfx;
+	static EventListener* el;
 	bool quit, done;
 };
 }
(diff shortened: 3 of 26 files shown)


Quipyowert2:
The crash in SDL_GL_SwapWindow on Linux is because of [Mesa bug #7875](https://gitlab.freedesktop.org/mesa/mesa/-/issues/7875), where the reference count of a resource was too low so it was freed before it should have been and then Mesa tries to use the freed resource.

Quipyowert2:
After fixing the Mesa bug, I get a black screen with a white dot in top left corner when I resize Glob2 on Linux:
Resize black screen 778a7d6f

Quipyowert2:
Fixed the black screen issue. I think the glOrtho call was conflicting with the gluOrtho2D call in GraphicContext::setRes or something like that. Removing the call to glOrtho but keeping gluOrtho2D makes the game stop being blank when resized.

On Linux, the right side of the game doesn't get drawn for a second when dragging the left border of the window to the left. On Windows, the vertical white line on the right side of the game in tutorial seems to be drawn multiple times at different horizontal positions when enlarging the window.

Quipyowert2:
I had to make these changes to Globulation 2 to get it to compile on MSYS2: https://gist.github.com/Quipyowert2/d2e1d3077efdbbc6b161a505445f1bce with the addition of prepending "/msys64" to each of the files in `source_files.extend` in src/SConscript.
And these changes to SCons to get Globulation 2 to configure successfully with SCons 4: https://gist.github.com/Quipyowert2/f6258728fd365464d8e1384d7b3dc79a

Quipyowert2:
Part of the reason why I can't seem to get this to work right is because the game draws too slowly. When I measured the time for one frame, it was 41ms, but that might have been the result of too much locking/unlocking of mutexen. My laptop's screen runs at 60Hz, so it would have to draw in 16ms or faster to be able to live resize the window without draw lag on Windows 10.

Quipyowert2:
Rebased on latest master so I could develop the PR in Visual Studio and extracted drawing code from MapEdit::run into MapEdit::draw.

Quipyowert2:
When trying to compile this PR in Visual Studio, I keep getting the error "xtime: ambiguous symbol" pointing at the condition_variable header and saying it could be either "xtime" or "boost::xtime". https://gist.github.com/Quipyowert2/10923de2138d5f4faf63e2e5cc7c802a

Quipyowert2:
I think I fixed the flickering of the globules, but the custom cursor flickers sometimes and the cursor animation goes too fast when resizing. When maximizing the window and then dragging the window downward on Windows 10, part of the tutorial message turns white and stays that way.

Giszmo:
I tried this out and it looks very glitchy. I think the rendering needs to be paused until the resizing is finished or something but even the edge I'm grabbing jumps around and doesn't stick to the mouse cursor.

[Screencast from 2023-11-15 21-22-56.webm](https://github.com/Globulation2/glob2/assets/96378/96c5f2ce-f28e-4741-b12f-0fef0af55b51)

Quipyowert2:
Maybe the reason for the graphical glitches is that when calling SDL_GL_MakeCurrent in GraphicContext::createGLContext, it errors with `wglMakeCurrent(): The requested resource is in use.` I've tried changing the type of GraphicContext::context to be a `thread_local` variable, but then the game has different graphical glitches. Turns out I can't make the OpenGL context current from the main thread while the same GL context is current on the other thread.

The reason I am trying to render from the other thread is that the game [freezes when I drag or resize the window](https://github.com/libsdl-org/SDL/issues/1059). There is probably a simpler solution I am overlooking. Also, sometimes the game deadlocks when I resize or move it, because EventListener::paint is waiting for the renderMutex to be unlocked, but the main thread is trying to call SDL_SetWindowSize.

genixpro:
Thanks for the work on resizable windows here. Closing this older implementation as superseded by the newer approach in #198, which handles resizing with main-thread cached presentation. #198 is still open, so the remaining review and implementation work will continue there.

PR #72, opened 2023-04-10
Windows fixes

Quipyowert2:
This pull request has some changes that made the code work better with Visual Studio and refactor some code that uses non-standard stuff like private GCC namespaces and `__attribute__`. Also, there is a workaround for microsoft/vcpkg#2292 included.

I'll make a separate pull request for the VS project files.


Quipyowert2:
Oops, I accidentally left the variable length array commit in there by mistake. That commit tried to fix the VLA error in Visual Studio but made the declaration much uglier.

PR #73, opened 2023-04-10
Restore and update Visual studio project files

Quipyowert2:
This pull request has the project files to compile Globulation in Visual Studio 2022 and goes with PR #72. I made this a separate pull request from the other because that way the extra build system can be approved or rejected independently of the fixes to make it compile on Visual Studio.

I wasn't sure what mkwin32.bat was for, so I kept that part of the commit that I partially reverted.


stephanemagnenat:
I see no reason to reject the additional VS project, so I'll merge it. Thank you for taking the care to separate the two PRs :⁠-⁠)

PR #77, opened 2023-04-14
Add GitHub actions build (WIP)

Shayan-To:
This works partially:

For Ubuntu, it builds, but packaging does not work.

For Windows, I got stuck getting a build. I couldn't figure out how I should get the libraries for cross-compiling on Ubuntu. So I tried building on msys2 on Windows, and scons cannot find anything. I think it is looking for things in the wrong places. And for some reason, the behavior on GitHub actions is different from the behavior on my local machine.

(I did this some time ago. In fact, I think I've forgotten what I was doing... And sorry for the commit messages! I think we should squash them, so I didn't touch anything.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,56 +1,58 @@
 name: Build
 
 on:
-  push:
-    branches:
-    - master
-  pull_request:
-    branches:
-    - master
+    push:
+        branches:
+            - master
+    pull_request:
+        branches:
+            - master
 
 jobs:
-  Ubuntu-Bulid:
-    name: Build on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-22.04, ubuntu-20.04, ubuntu-18.04]
-    steps:
-      - uses: actions/checkout@v3
-      - run: |
-          sudo apt-get install -y debhelper quilt scons libglu1-mesa-dev libvorbis-dev \
-            libspeex-dev libfreetype6-dev libboost-dev libboost-thread-dev \
-            libboost-date-time-dev libfribidi-dev portaudio19-dev libboost-math-dev \
-            dpkg-dev libsdl2-dev libsdl2-ttf-dev libsdl2-image-dev libsdl2-net-dev
-          scons -j16
-          #dpkg-buildpackage -I.hg
-  Windows-Build:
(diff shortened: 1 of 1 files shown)


Quipyowert2:
I just opened pull request #82 that fixes the build breakage caused because I added a variable Screen::scrollWheelEnabled but forgot to test that the server configuration builds when I added that variable. Now Ubuntu 20.04 should be able to build correctly with `server = True` set in options_cache.py which defines the YOG_SERVER_ONLY macro.

If you look closely at the include lines in the Windows msys2 build, they seem to be two paths concatenated together for some reason. The include arguments I'm talking about are `-Ibuild/libgag/src/D:/a/_temp/msys64/mingw64/include/SDL2 -Ilibgag/src/D:/a/_temp/msys64/mingw64/include/SDL2` which should probably be `-ID:/a/_temp/msys64/mingw64/include/SDL2`. I'm not sure why this happens; it could be a bug in either SCons or the `Sconstruct`/`SConscript` build scripts. Changing the last line under the `Windows-Build:` heading to `echo 'mingw = True' >> options_cache.py && scons -j16` might fix the Windows build.

If this gets merged, it can be used as a fallback if Travis-CI goes down and Travis-CI can still build even if GitHub Actions has a problem in the future. If GitHub itself goes down, well, that could be a problem since the SourceForge mirror hasn't been updated in quite some time.

Shayan-To:
@Quipyowert2 I rebased my changes on the latest master, and applied your suggestion for the Windows build.

The issue on Windows side still remains the same, and the error on Ubuntu side is beyond my understanding!

Quipyowert2:
The PR failing to build in Ubuntu 20.04 and 22.04 is my fault. I changed the `#ifdef` on the line in TextStream.cpp following the line moved by the 03_headers_order patch.

The build failure in the Windows build seems to be an include path problem. I am not sure why SCons smashes the paths together like that. The include argument `-Ibuild/libgag/src/C:/msys/1.0/local/include` should probably be two separate includes like this: `-Ibuild/libgag/src/ -IC:/msys/1.0/local/include`. The part of the code causing the problem could possibly be the `env.Append` calls in the top level [SConstruct at line 262](https://github.com/Globulation2/glob2/blob/d9e683c822e2a66b41968200129c1e9b0534bfa0/SConstruct#L262) or might be a bug in SCons, like I said in an earlier comment.

Giszmo:
#131 covers this. If not, re-open.

Issue #44, opened 2021-02-18
Crash in tutorial level "Flags and Completion" when destroying enemy's racetrack


Quipyowert2:
In Tutorial 4, "Flags and Completion", after destroying the enemy's Swarm and then racetrack, the game crashes with an assertion failure:
```glob2: src/Game.cpp:2289: void Game::drawMapGroundBuildings(int, int, int, int, int, int, int, int, int, Uint32, std::set<Building*>*): Assertion `building' failed.```

Globulation 2 version 214dfb22 compiled with g++ 7.5.0 in openSUSE 15.2 for WSL. X server: VcXsrv 1.20.8.1

`build/src/glob2 --version`:
```
Globulation 2 - 0.9.5.0

Compiled on Feb 17 2021 at 15:32:44

Compiled with SDL version 2.0.8
Linked with SDL version 2.0.8

Featuring :
* Map version 83
* Maps up to version 58 can still be loaded
* Network Protocol version 27
This program and all related materials are GPL, see COPYING for details.
(C) 2001-2007 Stephane Magnenat, Luc-Olivier de Charriere and other contributors.
See data/authors.txt for a full list.

Type build/src/glob2 --help for a list of command line options.

```

Reproducibility:
Crashes every time when following the steps below.

Steps to reproduce:
1. Press space until the flags and zones tab appears
2. Place exploration flag near enemy base
3. Place defense zone behind your defense towers (the large square blue zone)
4. Place war flag on enemy's swarm; set radius to 0.
5. Increase number of workers supplying the swarm to about 10.
6. Increase warrior bar on your swarm to about 3 or 4.
7. When warriors destroy enemy's swarm, move war flag onto racetrack just south of where the destroyed swarm was.

Actual result:
Glob2 crashes with an assertion failure when the racetrack is destroyed.

Expected result:
Tutorial finished, maybe?

Backtrace:
```
#0  0x00007ffffc9c9520 in raise () from /lib64/libc.so.6
#1  0x00007ffffc9cab01 in abort () from /lib64/libc.so.6
#2  0x00007ffffc9c1b1a in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007ffffc9c1b92 in __assert_fail () from /lib64/libc.so.6
#4  0x00000000004ef426 in Game::drawMapGroundBuildings (this=0x7ffffffe3650, left=0, top=0, right=32, bot=24, sw=1024,
[...]

PR #84, opened 2023-04-29
Improved coverage of building integrity check.

stephanemagnenat:
This improves coverage of building integrity check and highlights that the issue in #44 is actually a data corruption in the map itself.


Quipyowert2:
I think it would be better to fix the corrupted map than to write a HealMissing function that is a workaround. Although, I had a similar workaround on my local repo which was similar, but in `Map::setBuilding`. If this works, I guess you can merge it and someone can work on a proper fix for the bug later.

I tried to edit tutorial 4 map in the editor but it just went back to the previous screen instead of opening the map, but I should probably open an issue about that.

stephanemagnenat:
I agree that ideally it would be better to do the healing only when loading the map in the editor. And indeed it doesn't load the campaign maps, I created an issue (#89).

stephanemagnenat:
Also, at the moment even in this branch, if the map cannot be loaded, an error is not shown. There is support for showing an error in `Engine::initGame`, but by catching exception, while the normal load is not notified. I should fix that as well probably.

stephanemagnenat:
Ok, I fixed that bug in 1859cfec but the tutorial map loading problem remains, so it is another one.

stephanemagnenat:
I merge now, I update issue #89 to not forget to remove healing on non-editor load once the editor loading of campaign maps is fixed.

PR #97, opened 2023-05-10
Fix patch 03_headers_order.

Quipyowert2:
In PR #78, "snprintf & friends", I changed an #ifdef near the line that was moved in [patch `#3`](https://github.com/Globulation2/glob2/blob/d9e683c822e2a66b41968200129c1e9b0534bfa0/debian/patches/03_headers_order.patch). That change fixes an error with Visual Studio 2015 and later, but caused the patch to fail to apply in GitHub Actions PR #77.

Edit: Oops, I meant PR #78, not `#96`.


stephanemagnenat:
Is that still relevant?

Quipyowert2:
Yes it's still relevant. The patch won't apply anymore since I changed that line in PR #78. So we should probably let Debian know of this change.

PR #98, opened 2023-06-12
Fix headless segfaults when loading corrupted save games

Quipyowert2:
This PR fixes crashes found when fuzzing Globulation 2 [beta4.5-571-gd9e683c8](https://github.com/Globulation2/glob2/commit/d9e683c822e2a66b41968200129c1e9b0534bfa0) in headless mode with [AFL++ 4.07a](https://github.com/AFLplusplus/AFLplusplus) on an openSUSE 15.5 VirtualBox VM.

Fuzzed with `AFL_MAP_SIZE=2097152 afl-fuzz -i input -o output -f testcase.game -- build/src/glob2 -nox testcase.game 1 1` using an Auto_save.game file I happened to have lying around.

Each testcase can be checked with `gdb --ex run --args build/src/glob2 -nox <save file name> 1 1`. Note that, at least in Bash, the colons in the filename need to be escaped with backslashes.

[crashingsaves.zip](https://github.com/Globulation2/glob2/files/11717251/crashingsaves.zip)
[Auto_save.zip](https://github.com/Globulation2/glob2/files/11717261/Auto_save.zip)

### DIFF
--- a/src/Engine.cpp
+++ b/src/Engine.cpp
@@ -708,21 +708,24 @@ GameHeader Engine::loadGameHeader(const std::string &filename)
 {
 	MapHeader mapHeader;
 	GameHeader gameHeader;
-	InputStream *stream = new BinaryInputStream(Toolkit::getFileManager()->openInputStreamBackend(filename));
+	std::unique_ptr<InputStream> stream = std::make_unique<BinaryInputStream>(Toolkit::getFileManager()->openInputStreamBackend(filename));
 	if (stream->isEndOfStream())
 	{
 		std::cerr << "Engine::loadGameHeader : error, can't open file " << filename  << std::endl;
+		return GameHeader(); // an empty game header
 	}
 	else
 	{
 		if (verbose)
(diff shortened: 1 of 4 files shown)


src/Engine.cpp
```diff
@@ -708,21 +708,24 @@ GameHeader Engine::loadGameHeader(const std::string &filename)
 {
 	MapHeader mapHeader;
 	GameHeader gameHeader;
-	InputStream *stream = new BinaryInputStream(Toolkit::getFileManager()->openInputStreamBackend(filename));
+	std::shared_ptr<InputStream> stream = std::make_shared<BinaryInputStream>(Toolkit::getFileManager()->openInputStreamBackend(filename));
```
stephanemagnenat, line 711: Why not a unique_ptr?

Quipyowert2: Whoops, sorry. I had tried unique_ptr there already but then changed it to shared_ptr because it wasn't working or something. Changed back to unique_ptr. The game works now with unique_ptr. perhaps unique_ptr didn't work before due an unrelated code change in the loadGameHeader function.

Issue #80, opened 2023-04-18
Use SDL_GetTicks64 instead of SDL_GetTicks which overflows after 49 days.


Quipyowert2:
SDL_GetTicks will overflow after about 49 days, which might cause the game to glitch. SDL_GetTicks64, introduced in SDL 2.0.18, doesn't have this problem. A global find and replace wouldn't completely fix this bug; the variables that SDL_GetTicks was assigned to have to be changed from 32-bit to 64-bit. Also an `#ifdef` or similar should be put in place so that the game will still compile on SDL2 versions older than 2.0.18.

This will probably not be a problem on the client, but the YOG server might run that long.

PR #99, opened 2023-08-01
Use valarray instead of variable length arrays.

Quipyowert2:
Visual Studio is apparently allergic to Variable Length Arrays, so I removed them with this pull request. The presence of VLAs was causing the build to fail when I was working on fixing #80 today. I have replaced them with `std::valarray`s which should compile just fine in Visual Studio, but valarrays seem to be rarely used compared to vectors. Perhaps I should make them `std::vector`s instead?


Quipyowert2:
Oops, forgot I already made a pull request for that. Sorry, please disregard this PR.

PR #100, opened 2023-08-01
Fix the 49 days overflow problem.

### LINKED ISSUE #80: Use SDL_GetTicks64 instead of SDL_GetTicks which overflows after 49 days.

Quipyowert2:
As explained in #80, SDL_GetTicks will overflow to 0 after about 49 days, because it uses a 32-bit unsigned type to store milliseconds rather than a 64-bit type. This PR fixes that and also changes the assigned variables to be 64-bit.

Put compatibility shim in place for SDL < 2.0.18.

Fixes #80.

### DIFF
--- a/libgag/include/SDLCompat.h
+++ b/libgag/include/SDLCompat.h
@@ -1,0 +1,11 @@
+#include <SDL_version.h>
+// Compatibility shim for SDL before version 2.0.18
+#if SDL_VERSION_ATLEAST(2,0,18)
+// nothing
+#else // older than 2.0.18
+#ifdef MSC_VER
+#define SDL_GetTicks64 SDL_GetTicks
+#else // not MSC_VER
+#define SDL_GetTicks64() SDL_GetTicks()
+#endif // MSC_VER
+#endif // SDL_VERSION_ATLEAST
--- a/libgag/include/SDLGraphicContext.h
+++ b/libgag/include/SDLGraphicContext.h
@@ -31,18 +31,8 @@
 #include <set>
 #include <boost/tuple/tuple.hpp>
 
-#include <SDL_timer.h>
+#include <SDLCompat.h>
 
-// Compatibility shim for SDL before version 2.0.18
-#if SDL_VERSION_ATLEAST(2,0,18)
-// nothing
-#else // older than 2.0.18
-#ifdef MSC_VER
-#define SDL_GetTicks64 SDL_GetTicks
-#else // not MSC_VER
-#define SDL_GetTicks64() SDL_GetTicks()
-#endif // MSC_VER
-#endif // SDL_VERSION_ATLEAST
 
 namespace GAGCore
 {
--- a/libgag/include/SDLGraphicContext.h
+++ b/libgag/include/SDLGraphicContext.h
@@ -31,6 +31,19 @@
 #include <set>
 #include <boost/tuple/tuple.hpp>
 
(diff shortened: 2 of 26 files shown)


stephanemagnenat:
The file `SDLCompat.h` is missing from the commits.

stephanemagnenat:
Also, why does `SDLGraphicContext.h` includes `SDL_timer.h` directly, and not the compat header? Also, maybe that header could be named more specifically, like `SDLTimerCompat.h`.

stephanemagnenat:
Why the new type for ticks is sometimes `Sint64` and sometimes `Uint64`?

Quipyowert2:
> The file `SDLCompat.h` is missing from the commits.

I thought I added that file to the third commit in this PR, but apparently I didn't.

> Why the new type for ticks is sometimes `Sint64` and sometimes `Uint64`?

Because I wanted to have the the type in the `std::max` template parameter match the type of the variable declaration on the line immediately before.

I would have responded earlier but I haven't been checking my email or the GitHub app or website much lately.

stephanemagnenat:
> Because I wanted to have the the type of the std::max template parameter match the type of the variable declaration on the line immediately before.

I see, the code is not very consistent (it already wasn't with the old version). For example [here we use `Sint64`](https://github.com/Globulation2/glob2/pull/100/files#diff-149cc4f4de0f52e459cc262ccaec510cfa246ee67a5d39282a286f587636c958R113) and [here `Uint64`](https://github.com/Globulation2/glob2/pull/100/files#diff-887b2319f8f3a37d7014b12d972a0840eb459d23df7c73388f3c065d925296ddR570).

I wonder whether we should use the chance of this change to be consistent, with either of these two options:
1. Always using `Uint64` for the ticks, and assuming the later cannot be earlier than the former and therefore not caring about the possible overflow.
2. Always using `Sint64` for the ticks when they are subtracted, to protect against the overflow. 

I miss Rust here, with automatic overflow checks in debug mode. Maybe we could take option 1. but add an assertion that the later is greater or equal to the former. That would mimic Rust's behavior, at the cost of some manual work.

What do you think?

Quipyowert2:
Option 2 or 3 sound fine to me. I don't think I'll be able to work on this for a while as my laptop's CPU hit 96°C and I had to shut it off. I think it needs the thermal paste replaced or the dust cleaned out of the fans. It's been running very hot lately.

Quipyowert2:
I have made most of those `Sint64`s unsigned, changed all the `std::max` template arguments to be Sint64 for consistency, and made use of static_cast in several places, as it is a bit more readable than C-style casts, albeit more verbose.

I noticed that changing the delay variable in Engine::run to be `Sint64` rather than `Uint64` causes the CPU load indicator to glitch. I made an issue #102 and PR #103 for that, but then realized, a few minutes later, that it was because of the delay then being able to be negative which caused the length of the bar to exceed 40px in `GameGUI::drawTopScreenBar`, thus the bug.

PR #103, opened 2023-08-28
Fix CPU load indicator reaches right side of screen.

### LINKED ISSUE #102: CPU load indicator reaches right edge of screen on a slow computer / under load
On my slow, 5-year-old laptop, I have several programs open at the same time. Globulation 2's indicator for whether the computer is running slow reaches the right edge of the window. It's only supposed to get to the vertical white line and stop.

Steps to reproduce:
1. Set your computer (or virtual machine) to a low clock speed
2. Start the tutorial in Glob2

What happens:
<details> 

<summary> The CPU load indicator glitch </summary>

image

</details>

What I expected:
The CPU load indicator should stay in between the two vertical lines.

For quite a while, I actually thought this was a problem with the way I implemented resize in PR #64, but just noticed it on a lightly modded Globulation 2 today when I was trying to fix issue #80.

This happens with Globulation 2 beta4.5-574-g39980149 on Windows 10 Home 22H2 19045.3324 running on an MSI GS63 Stealth 8RE laptop. I have to set my laptop to run at the slowest mode: Eco, which is about 0.8 GHz or so, for it not to overheat. I'll probably have to buy a new computer pretty soon whenever this one decides to bite the dust.

Quipyowert2:
Fixes #102.

### DIFF
--- a/src/GameGUI.cpp
+++ b/src/GameGUI.cpp
@@ -4082,7 +4082,7 @@ void GameGUI::drawTopScreenBar(void)
 	else
 		memcpy(actC, redC, sizeof(redC));
 
-	int cpuLength = int(float(cpuLoad) / 100.0 * 40.0);
+	int cpuLength = int(float(std::min(cpuLoad, 100)) / 100.0 * 40.0);
 
 	globalContainer->gfx->drawFilledRect(dec, 4, cpuLength, 8, actC[0], actC[1], actC[2]);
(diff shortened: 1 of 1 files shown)


Giszmo (changes requested):
The bar is meant to over-shoot to give an idea by how much your machine is slowing down an online game for example. Maybe that can be limited to 200% with some indication that even those were exceeded?

stephanemagnenat:
As #100 is merged, should we close this PR? Or do you still see overshoot @Quipyowert2 ?

PR #106, opened 2023-11-22
Fix nicowar deadlock

Giszmo:


stephanemagnenat:
This patch changes the behavior of the engine, so if I am not mistaken, to avoid desync the network protocol version should be bumped (`NET_PROTOCOL_VERSION` in `Version.h`). Can anyone confirm?

Giszmo:
Good point. The change would result in games desyncing in the first minute of a match with Nicowars. If it would only desync once a crash would occur, I wouldn't see it that important to bump the version.

stephanemagnenat:
@Giszmo there is no version bump yet following our discussion, right? I do not have time to do it right now but I would like to not forget about it.

Giszmo:
https://github.com/Globulation2/glob2/pull/108

Issue #43, opened 2021-02-17
Fix remaining memory leaks


stephanemagnenat:
A small session in Valgrind shows leaks in these categories:

- [x] GameHeader (214dfb224b5604f085a9cccf4ce0e74348f4d84f)
- [x] Game (214dfb224b5604f085a9cccf4ce0e74348f4d84f)
- [ ] USL
- [x] VoiceRecorder

In detail:
```
==86996== 
==86996== HEAP SUMMARY:
==86996==     in use at exit: 189,017 bytes in 2,524 blocks
==86996==   total heap usage: 395,616 allocs, 393,092 frees, 1,834,156,600 bytes allocated
==86996== 
==86996== 1 bytes in 1 blocks are definitely lost in loss record 49 of 1,704
==86996==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==86996==    by 0x58455A6: _XlcDefaultMapModifiers (in /usr/lib/x86_64-linux-gnu/libX11.so.6.3.0)
==86996==    by 0x584597A: XSetLocaleModifiers (in /usr/lib/x86_64-linux-gnu/libX11.so.6.3.0)
==86996==    by 0x4B4A824: ??? (in /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.10.0)
==86996==    by 0x4B5145A: ??? (in /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.10.0)
==86996==    by 0x4B23F6A: ??? (in /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.10.0)
==86996==    by 0x4A938E6: ??? (in /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.10.0)
==86996==    by 0x4D1A42: GAGCore::GraphicContext::GraphicContext(int, int, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (GraphicContext.cpp:1974)
==86996==    by 0x506E5C: GAGCore::Toolkit::initGraphic(int, int, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (Toolkit.cpp:55)
[...]

PR #112, opened 2024-04-10
Fix USL memory leak.

Quipyowert2:
Fixes issue #43.

Note: Rule of Three copy (assignment) operators are not included with this commit.


stephanemagnenat:
Thank you for fixing this leak. I think that ideally we should use unique instead, but I will merge nonetheless. I assume that you have run the patched version under valgrind and that there are no over-deletion.

PR #120, opened 2024-05-28
Speed up terrain drawing with glDrawArrays from OpenGL 2.0.

Quipyowert2:
Generates a sprite sheet on the fly, queues UV and vertex coordinates in GC::drawSurface and finally in finishDrawingSprite, many different sprites in the texture atlas can be drawn at the same time with glDrawArrays.

With VSync off and after commenting out the `SDL_Delay`s in GUIBase.cpp and Engine.cpp, I get a frame rate of about 500 on the title screen and around 100 or so in tutorial 4.

With the new sprite sheet code, the frame rate is roughly ~60 FPS without the two SDL_Delay calls. It was only about ~40 FPS before I optimized it, so about a 50% increase in FPS.

Note: This PR adds a new dependency, ~glad~ epoxy. The functions createTextureAtlas and finishDrawingSprite require OpenGL 2.0. 

TODO:
- [x] Fix crash in USL when starting tutorial. (Fixed by #121)


stephanemagnenat:
Nice to see that there is optimization work in progress!
One question though: for the new `glad` dependency, could you adjust the `SConstruct` file. For example, on Ubuntu, I do not know what to install to have glad.

Quipyowert2:
Sorry, I forgot to add the glad files to the PR. I downloaded the zip from [Glad generator](https://glad.dav1d.de/) but forgot to commit it.

Quipyowert2:
On second thought, maybe [Epoxy](https://github.com/anholt/libepoxy) or [OpenGL Extension Wrangler](https://glew.sourceforge.net/) would be better options. The glad header and source file the website generated is over 4,000 lines and the license for the generated code is unclear. Glad is MIT licensed, the XML file it uses to generate the code is Apache licensed, and I don't know what license the generated code is under. I found this though: [License of generated code](https://github.com/Dav1dde/glad/issues/101).

Quipyowert2:
Rebased onto `master` branch. Now it should _not_ crash when I run the tutorial.

Quipyowert2:
@stephanemagnenat My [latest commit](https://github.com/Globulation2/glob2/pull/120/commits/6adb655f6116fe93a0280ad046070c7699361509) changes `Sprite::atlas` to be a unique pointer because it is the only pointer to the sprite sheet. `createTextureAtlas` should be slightly more exception-safe now.

Please let me know if you need me to squash the several "Fix foo" commits into one or remove the old commits referencing `glad` from this PR, or if you would like any other changes.

Quipyowert2:
Fixed flags, zones, fog of war, and bullets not being drawn.

stephanemagnenat:
@Quipyowert2 is this PR completed? If so, yes, please squash commits.

I have some questions though:
- `DrawableSurface` has now both an optional `sprite` pointer and an `usingAtlas` boolean. Are there cases when `sprite` is non-null but `usingAtlas` is false?
- In `DrawableSurface`, is there a reason why `texX` and `texY` are explicitly set to 0 but `w` and `h` are not? It might actually make sense to use `std::optional` here to jointly denote the using of a sprite sheet and the related data if it is used. You can imagine having a simple struct with the sprite sheet data, and have an optional of it.
- Why is the atlas code in `DrawableSurface::initTextureSize(void)` commented out?
- Why is the dirty set in `void DrawableSurface::uploadToTexture(void)` commented out?
- What happens if user code calls a draw function (e.g. `drawPixel(...)`) on a `DrawableSurface` that is using an atlas? Is a new surface allocated, the sprite data copied, and the link to the atlas severed?

Quipyowert2:
The PR is just about done, but I had to add some missing finishDrawingSprite calls because some of the right panel items weren't being drawn. Regarding the first bullet point, `usingAtlas` variable is probably unnecessary.

Bullet point 2, yeah optional is probably a good idea here.

Re commented out code: The atlas code in initTextureSize is commented out because I was testing something and forgot to remove it. The line setting `dirty` is commented out because I didn't want to overwrite the sprite sheet with a single tile.

I'm not sure about what happens if you call drawPixel on a sprite-sheet. I might have to test that tomorrow.

stephanemagnenat:
Ok, thank you for your quick answer. Please ping me when everything is done and you want this merged!

Quipyowert2:
I have pushed a commit to prevent drawing to the Sprite::atlas surface after it has been generated. It is a little difficult to work on graphics stuff right now as my laptop screen has started turning purple and flickering for 15 minutes to several hours after every reboot.

libgag/src/Sprite.cpp
```diff
@@ -67,9 +101,68 @@ namespace GAGCore
...
+			{
+				w = image->getW();
+				h = image->getH();
+			}
+			if (image->getW() != w || image->getH() != h)
+				return;
+		}
+		int tileWidth = images[0]->getW();
+		int tileHeight = images[0]->getH();
+		int sheetWidth = tileWidth * (static_cast<int>(sqrt(numImages)) + 1);
```
Quipyowert2, line 135: I should probably use a loop that tries all the possible width and height combinations here where width and height are <= the maximum texture size (`glGetIntegerv(GL_MAX_TEXTURE_SIZE)`). May need multiple sprite sheets if the maximum texture size is too small.

Quipyowert2:
I extended createAtlas to be able to handle a limited texture size, but I haven't pushed my changes to this branch yet. I have it on another branch (sprite-sheet-after-rebase). I thought I squashed the Fix stuff not being drawn in map editor commit into the other similar "Fix" commits, but apparently I didn't. I tried using std::optional, but it wouldn't compile until I changed Visual Studio's C++ standard to C++17. Then it complained about mem_fun, which was available in C++14 (the default standard in VS, I think), but has been removed in 17.

Quipyowert2:
I used boost::optional instead of std::optional to avoid requiring C++17. I ended up moving `finishDrawingSprite` to the Sprite class, then moving it back to the GraphicContext class. Moving the method to the Sprite class required moving the GLState object to the SDLGraphicContext.h header and passing it somehow to the Sprite class. I guess it makes more sense to queue drawing to a surface and then actually drawing it in the same class, but the code looks nicer in the Sprite class. Although, if we were to support DirectX as a rendering backend in the future, there would have to be two finishDrawing methods, one for GL and one for DX.

What do you think, should I keep finishDrawingSprite in the GC class or move it to the Sprite class?

Quipyowert2:
Reverted the changes that handled a limited max texture size, as they make the PR more complex for not much benefit. If the graphics card doesn't support at least 1024x1024 texture size, it probably can't run the game anyway. I also squished the commit casting the result of sqrt to integer into the first commit.

@stephanemagnenat Does this look ready to merge to you? Please review.
- [x] Compiles and runs on my machine

SConstruct
```diff
@@ -193,6 +193,13 @@ def configure(env, server_only):
         configfile.add("HAVE_OPENGL ", "Defined when OpenGL support is present and compiled")
         env.Append(LIBS=gl_libraries)
     
+    if not server_only:
+        if conf.CheckLib('epoxy') and conf.CheckCXXHeader('epoxy/gl.h'):
+            env.Append(LIBS=["epoxy"])
+        else:
+            print("Could not find epoxy/gl.h")
+            missing.append("epoxy")
```
stephanemagnenat, line 201: By adding "epoxy" to `missing`, we prevent the building of Glob2 without OpenGL support. Except if that ability is not present any more, normally Glob2 can be built with software SDL. I think that we should keep this possibility.

libgag/src/GraphicContext.cpp
```diff
@@ -25,6 +25,12 @@
 #include <string>
 #include <sstream>
 #include <iostream>
+#include <epoxy/gl.h>
```
stephanemagnenat, line 28: I would put this in an `#ifdef HAVE_OPENGL` (see my first comment).

libgag/src/GraphicContext.cpp
```diff
@@ -1682,22 +1695,80 @@ namespace GAGCore
...
+	// Lets us efficiently draw terrain and water.
+	void GraphicContext::finishDrawingSprite(Sprite* sprite, Uint8 alpha)
+	{
+#ifdef HAVE_OPENGL
+		if (_gc->optionFlags & GraphicContext::USEGPU)
+		{
+			if (!sprite->atlas)
+			{
+				// No sprite sheet, so we have nothing to draw.
+				assert(!sprite->vertices.size());
```
stephanemagnenat, line 1737: An absolute detail, but I would find the writing of this as `assert(sprite->vertices.size() == 0);` closer to the conceptual intent.

libgag/src/Sprite.cpp
```diff
@@ -27,6 +27,40 @@
...
+#include <boost/make_unique.hpp>
+#elif BOOST_VERSION >= 105700
+#include <boost/move/make_unique.hpp>
+#else
+#error "Can't make_unique when there's no Boost and C++ standard is earlier than C++14"
+#endif
+using boost::make_unique;
+#endif // __cplusplus
+
+#define GL_GLEXT_PROTOTYPES
```
stephanemagnenat, line 46: Why isn't this inside the `#ifdef HAVE_OPENGL`?

libgag/src/Sprite.cpp
stephanemagnenat: Yes, or/and maybe use the closest bigger or equal power of 2, if that is still relevant nowadays.

libgag/src/Sprite.cpp
```diff
@@ -67,9 +101,78 @@ namespace GAGCore
...
+			if (!tileWidth || !tileHeight)
+			{
+				tileWidth = image->getW();
+				tileHeight = image->getH();
+			}
+			if (image->getW() != tileWidth || image->getH() != tileHeight)
+				return;
+		}
+		int sheetWidth = tileWidth * (static_cast<int>(sqrt(numImages)) + 1);
+		int sheetHeight = tileHeight * (static_cast<int>(sqrt(numImages)) + 1);
```
stephanemagnenat, line 154: I think that we should at least check that both `sheetWidth` and `sheetHeight` are smaller or equal to `glGetIntegerv(GL_MAX_TEXTURE_SIZE)`.

src/Game.cpp
```diff
@@ -1964,13 +1965,16 @@ inline void Game::drawMapWater(int sw, int sh, int viewportX, int viewportY, int
...
 			globalContainer->gfx->drawSprite(x, y, globalContainer->terrainWater, 0);
+	globalContainer->gfx->finishDrawingSprite(globalContainer->terrainWater, 255);
 }
 
 inline void Game::drawMapTerrain(int left, int top, int right, int bot, int viewportX, int viewportY, int localTeam, Uint32 drawOptions)
 {
 	Uint32 visibleTeams = teams[localTeam]->me;
 	if (globalContainer->replaying) visibleTeams = globalContainer->replayVisibleTeams;
 
+	Sprite* sprite;
```
stephanemagnenat, line 1976: Is there a reason to move `sprite` here?

stephanemagnenat (changes requested):
Overall very good. My key blocking point is the disabling of software-only rendering, of which I do not see the necessity.

stephanemagnenat:
Another small point: is there a way to check OpenGL version in `SConstruct`, in order to report warning and disable GL rendering if it is <= 2.0?

libgag/src/Sprite.cpp
Quipyowert2: Oops, I also did that in GraphicContext.cpp.

libgag/src/Sprite.cpp
Quipyowert2: 1. It looks like the game already tries to handle non-power-of-two sizes in DrawableSurface::initTextureSize and uploadToTexture.

2. I agree that checking for maximum texture size instead of mysteriously failing would be good.

src/Game.cpp
Quipyowert2: No reason in particular, I'll move the pointer back where it was.

Quipyowert2:
I'm working on this, but it's getting late here. I might have time to work on this tomorrow.

libgag/src/Sprite.cpp
Quipyowert2: Changed it to print an error and return false if the sprite sheet would be too large.

SConstruct
Quipyowert2: Sorry, I didn't realize that breaks compiling without OpenGL support. I am not very good with SCons. Should be fixed in my latest push.

libgag/src/GraphicContext.cpp
Quipyowert2: Done.

libgag/src/GraphicContext.cpp
Quipyowert2: Changed to assert that it is empty.

libgag/src/Sprite.cpp
Quipyowert2: Should be fixed now.

Quipyowert2:
Does the PR look good now? I wasn't sure what to do if sprite sheet creation fails, but I changed it to just print a message for now. 

> Another small point: is there a way to check OpenGL version in `SConstruct`, in order to report warning and disable GL rendering if it is <= 2.0?

Yes. This can be done by compiling and running a toy SDL/OpenGL program that opens a window with corresponding OpenGL context, and then calling `glGetString(GL_VERSION)`. Note that the version returned could differ based on whether a core profile or compatibility profile is requested, but that will be a problem if and/or when we start using OpenGL 3.

stephanemagnenat:
We are almost there. Actually I realized that non-OpenGL support was already broken, as "OpenGL" and "GLU" are added to `missing`, which is wrong. I believe that the right thing to do is:
- Do not add these to `missing`
- Move the search for epoxy to [line 191](https://github.com/Globulation2/glob2/blob/c8e71c030cb27aaba86118e82a421ffd937d49e8/SConstruct#L191) and add it to `gl_libraries`.
- In [line 192](https://github.com/Globulation2/glob2/blob/c8e71c030cb27aaba86118e82a421ffd937d49e8/SConstruct#L192C8-L192C20), check that `gl_libraries` has size 3 (if not Darwin) or size 1 (if Darwin).

Do you agree?

SConstruct
stephanemagnenat: See my comment below, support for non OpenGL was broken already.

stephanemagnenat (approved):

stephanemagnenat:
I have decided to go on and merge!

stephanemagnenat:
On my system, this PR didn't build due to a missing `memory` header, I fixed it in 7dea3ab4bd2f9f77a6710806e9914579f2c64ebc.

stephanemagnenat:
I resurrected the non-OpenGL building in 21c07a527335c88072cf097f6b876682775763f8.

Issue #124, opened 2025-01-02
Glob2 renders a white square instead of semi-transparent building when placing a school


Quipyowert2:
When I start tutorial 4 and start to place a school in a spot where there is already a building, the game shows a white square with a red outline instead of a semi-transparent building indicator like when placing a hospital. Happens with commit c196fa55b458a637146470a820af903113d38659. This is probably my fault. I think the bug is from my PR #120.

Please disregard the tutorial messages in the screenshots, the "A unique string" is from when I was testing something.

![semi-transparent hospital on swarm](https://github.com/user-attachments/assets/8273f14f-1859-4c1f-99dc-2b8a938b3324)
![missing image when placing school on swarm](https://github.com/user-attachments/assets/d7aae693-3aa5-4577-8f3b-71791892d3f7)

stephanemagnenat:
Hum, it feels to me like the per-object-draw transparency is not correct when using the new drawing path. I do not have the full new drawing code in mind, but I feel that somehow the transparency of each draw should be stored somewhere and used in the shader (or in the corresponding fixed pipeline) when drawing each object, even from the same sprite sheet.

Quipyowert2:
Looks like I need to somehow pass the transparency to `glDrawArrays` instead of falling back to the slow path (glBegin/glEnd) when the sprite is being drawn with transparency. From a quick search, it looks like this can be done with vertex array objects or maybe shaders? I will have to look up how to do that when I have time, hopefully in the next few days.

stephanemagnenat:
In [Candli](https://cand.li) I'm doing something similar using a simple pair af vertex/fragment shaders. But it is in WebGL so the setup API calls differ a bit. I can copy/paste the shaders here if that is helpfu (later, I'm on mobile now)l, although they might need some adaptation for OpenGL 2.

stephanemagnenat:
Here are the shaders for WebGL 1.0:

**Vertex**
```
attribute vec2 aVertexPosition;
attribute vec2 aTextureCoord;

uniform mat3 uWorldToViewport;
uniform mat3 uPoseToWorld;

varying lowp vec2 vTextureCoord;

void main() {
	gl_Position = vec4(uWorldToViewport * uPoseToWorld * vec3(aVertexPosition, 1.0), 1.0);
	vTextureCoord = aTextureCoord;
}
```

**Fragment**
```
varying lowp vec2 vTextureCoord;

uniform sampler2D uSampler;
precision lowp float;
uniform float uAlpha;

[...]

PR #125, opened 2025-01-06
Fix bug #124 - Placing school renders white square.

Quipyowert2:
This PR draws the sprite immediately when there is transparency involved, saving and restoring the vertices and texture coordinate arrays. This seems to fix the bug for now. I'm not sure why the white square bug was happening, but it might be a conflict between glDrawArrays and falling back to glBegin/glEnd rendering when there is any transparency. I think the bug was because the level 1 school was the only building sprite without a team color mask.

### DIFF
--- a/libgag/src/GraphicContext.cpp
+++ b/libgag/src/GraphicContext.cpp
@@ -1697,9 +1697,20 @@ namespace GAGCore
 
 			// draw
 			glState.setTexture(surface->texture);
-			if (surface->textureInfo && surface->textureInfo->sprite && alpha == Color::ALPHA_OPAQUE)
+			if (surface->textureInfo && surface->textureInfo->sprite)
 			{
 				Sprite* sprite = surface->textureInfo->sprite;
+				std::vector<float> oldVertices, oldCoords;
+				// If drawing with transparency, save vectors, draw immediately, then restore
+				if (alpha != Color::ALPHA_OPAQUE)
+				{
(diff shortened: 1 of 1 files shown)


Quipyowert2:
I am working on a less silly solution to the problem using the shaders given in #124, but am getting a segfault for some reason.

stephanemagnenat:
Shall I merge this work-around for now or do you prefer to wait for a proper solution?

Quipyowert2:
Sure, you can merge this.

 I don't have much experience with shaders, so it will probably be a while before I manage to get the shader code working. Last night I copied some code from a tutorial into the game, ran it and kept getting a blank screen. Then I simplified the vertex shader by setting gl_Position to ftransform(), and setting the fragment shader to output yellow. Then I got a yellow screen. It seems that the fragment shader isn't working, or maybe I didn't connect the inputs correctly. Not quite sure how to initialize a sampler2D uniform. At least it's not crashing.

stephanemagnenat:
I merged.

I also do not have much experience in configuring shaders outside WebGL contexts. I would expect that the shader syntax is the same between WebGL 1 and OpenGL 2, as WebGL 1 is based on OpenGL 2 ES, but there might be subtle differences and also there might be something wrong in the way you set up the data to be passed at the shader (e.g. maybe the texture is not set up correctly). Do you get the correct part of the screen drawn but with the wrong color? Or even the location is wrong?

PR #129, opened 2026-05-07
AI trainer support: refactor, infrastructure, and one desync fix

kylelutze:
## Summary

This branch is **mostly a long-running refactor + infrastructure pass on the C++ codebase** to make it portable, testable, and ready for the Rust port and AI-trainer work. The great majority are behavior-preserving cleanups validated against the deterministic replay-diff harness.

Broadly:

- **Code organization** — split the largest TUs into per-concern files; relocated sources under `src/{ai,team,unit,building,map,gui,net,render,yog}/`; `#pragma once` everywhere; lowercased directory names.
- **Modernization** — boost → std (`shared_ptr`, `lexical_cast`, `tuple`, threads); C++20 build; SPDX headers replacing GPL boilerplate.
- **Dead-code removal** — `logFile` members + the `fprintf`s feeding them (game/team/unit/building/AICastor); unused MapLog; orphaned text-loader infra; verbose pathfinding printfs.
- **Static per-type tables** — Units, Buildings, Resources, UnitsSkins moved off file loaders to compile-time tables.
- **Pathfinding rewrite** — gradient computation swapped from BFS to a chamfer distance transform (see breaking-change section #2 below). Fixed a longstanding Uint8 underflow in war/guard/clearing flag seeding along the way.
- **Test infrastructure** — deterministic replay-diff regression harness (`tests/baselines/cpp-refactor.replay`); new four-game gradient corpus (`tests/baselines/gradient/gd-*.{game,replay}`) covering map-size/team-count/AI-mix variation; MapQuery characterization suite; `ChecksumSidecar` per-tick checksum dump for cross-replay debugging.
- **Headless / scripting tooling for the AI trainer** — `DatasetWriter` for (state, action) records; `--save-game-as`, `--map`, `--matchup`, `--ai-types`; `GLOB2_GAME_END` summary line; `GLOB2_REPLAY_PATH` env var; absolute-path handling in FileManager / ReplayWriter.
- **Bug fixes incidentally found** — typo in `tryToBuildingSiteRoom` (`setBuilding(decLeft, decLeft)`), unguarded `get_height < dist` in farming AI, AIWarrush wrap typo, `Game::save` mutating live `mapHeader`, MapGenerator dead-code + 4 bugs, latent farming AI fixes. All validated against the replay baseline.

## ⚠️ Breaking changes called out explicitly

Two commits on this branch **deliberately change simulation behavior**. Both are pre-existing-bug fixes that canonicalize previously-incorrect or implementation-defined behavior; everything else is behavior-preserving.

### 1. `team: stable gid tiebreak in prioritize_building` (419eb4bd)

`std::sort` is unstable. When two buildings tied on every comparator field (priority, type/level bucket, unit ratio, ressource ratio), their relative order in `Team::buildingsNeedingUnits` was implementation-defined. Different binaries — different libstdc++/libc++ versions, optimization levels, ABIs (32-bit vs 64-bit `size_t`) — could put tied buildings in different positions, then call `subscribeToBringRessourcesStep()` in different orders, which consumes `syncRand()` in different orders, which **desyncs lockstep multiplayer**. This is a longstanding source of the multiplayer desyncs that have plagued the game for years.

The fix adds `gid` as a final stable tiebreaker (`gid` is unique within a team and identical across all clients).

Consequences:
- **Old replays from the previous binary will not deterministically replay** against this build past whatever tick the first tie occurs. This is unavoidable for any fix that removes implementation-defined ordering.

### 2. `map/gradient`: chamfer DT replaces BFS, fixes Uint8 underflow in flag seeding (1e228ce0 → 487027af)

The pathfinding gradient algorithm was swapped from a BFS work-queue approach (`updateGlobalGradientVersionSimple`/`Simon`) to a chamfer distance transform (forward+backward raster sweeps until stable). 

**The bug:** war/guard/clearing flag seeding pushed every cell within `unitStayRange` into `listedAddr` at value 255, but the obstacle scan that ran next could overwrite some of those cells back to 0 *without removing them from listedAddr*. BFS later popped a stale entry, computed `gradient[addr] - 1 = 255` (Uint8 underflow), and propagated phantom 255s outward from obstacle cells. Visible in-game as warriors/guards/clearers loitering one cell outside the real flag zone near obstacles. The fix compacts `listedAddr` after the obstacle scan so only true sources reach the propagation pass — fully deterministic, removes a Uint8 underflow that produced different gradient values depending on which cells the obstacle scan happened to overwrite.

Consequences:
- **Old replays will not deterministically replay** past the first tick where a war/guard/clearing flag's gradient is consulted near an obstacle.

## Test plan

- [x] `scons release=1 -j16` clean build on macOS arm64 (Darwin 25.4)
- [x] `./build/src/glob2 --nox games/G2.game 0 1` runs to completion (126,950 ticks / 84 min sim time, 4 AI players: Warrush + 2× Castor + Nicowar)
- [x] Replay-diff harness: 4 sequential headless runs of `G2.game` produce byte-identical `last_game.replay` (= `tests/baselines/cpp-refactor.replay`)
- [x] Gradient corpus: all four games (`gd-small-2ai`, `gd-large-4ai`, `gd-archipelago`, `gd-bigarena-long`) byte-equal against `tests/baselines/gradient/*.replay` after the chamfer cleanup commit
- [x] MapQuery characterization tests pass
- [ ] *Reviewer*: spot-check the `prioritize_building` change at `src/team/team_step.cpp` and confirm the gid tiebreaker reasoning
- [ ] *Reviewer*: spot-check the chamfer body in `src/map/gradient/MapGradientGlobal.cpp` and the listedAddr underflow fix in `1e228ce0`
- [ ] *Reviewer*: 4-player live multiplayer match — desync should no longer reproduce on the scenarios where it previously did


stephanemagnenat:
One question: why files that are split (e.g. GraphicContext) broke the naming convention?

kylelutze:
> One question: why files that are split (e.g. GraphicContext) broke the naming convention?

that has been rectified

stephanemagnenat:
@kylelutze please ping when you want a review, and point to the highest value review targets to address first.

stephanemagnenat:
There is a commit with WIP in its title.

stephanemagnenat:
Should we use git lfs for the big binary .game and .replay files?

stephanemagnenat:
Why the building/unit default stats files were removed? Is there a replacement?

kylelutze:
> @kylelutze please ping when you want a review, and point to the highest value review targets to address first.

@stephanemagnenat it'll be a while before it's ready, that's why I set it for draft mode. don't worry about reviewing it right now. I'm so used to reviewing PRs that it's the easiest way for me to review a large batch of changes to see how it's looking. My plan was to send an email to the mailing list when it is ready but you're fast on noticing this.

I'll have to check out git lfs. I haven't used it before but that is an excellent idea for those files.

The entity stats are now directly in code, such as in race.cpp with the consts in UnitConsts.h. I'm not sure I like that design quite yet, but we've never changed the stats between versions and it allowed the removal of quite a bit of loader code and simplified things.

This work is for two experiments for me. 
1. can I get an AI to port this whole code base to rust. 
2. can I train an AI through reinforcement learning. 

For 1, I found that initial code base cleanup is required for a port to be successful so I've been having an AI go do all of this work in a way that will improve port quality. My first port attempt did not go well. We don't ever need to merge it if it isn't wanted though.

stephanemagnenat:
I'll be happy to see code quality improve, and even more happy to see a Rust port. So in principle I do not have problem merging that, but I'll like to review it a little bit. 70k lines feels a lot though, but there is quite some mechanical work. I might be able to use an AI to extract a small subset of non-trivial changes to review.

An alternative would be to directly work on the Rust port. I am anyway doing a lot of Rust and no C++ nowadays.

Ah, and once we have Globulation 2 in Rust, my new scripting language Ferlium might prove handy: https://ferlium.dev

kylelutze:
why that instead of Lua?

I did start with directly doing a Rust port. I even made a lot of progress, but there were a lot of issues trying to have an AI do a Rust port with the cpp code base as it is that it couldn't get past. Claude just couldn't navigate the logic with how some of it was linked together. It kept getting lost on some sneaky things in the code base. So far the biggest helps:

- code files must be under 500 lines. That alone makes a tremendous difference
- putting labels on fixed values helps massively in it figuring out the purpose of logic blocks
- When doing cpp -> cpp work it does a good job of recognizing what is old cpp practices vs new cpp practices. When porting it goes "I don't understand why this is the way it is, fingers crossed I port it correctly", so modernizing it makes a decent difference too.

stephanemagnenat:
Because Ferlium is statistically typed with type inference, and can in theory compile to native code through a Just In Time compiler (we are working towards WASM compilation right now). Think of it as a Rust script, with a bit of the Haskell type system. Also, Ferlium is designed to interface seamlessly with Rust (i.e. Rust functions can be called directly).

Lua is interpreted with bytecode, has a syntax far away from Rust, and needs some adaptation to interface with Rust native types.

stephanemagnenat:
Very interesting, thank you for the port attempt report!

stephanemagnenat:
Ah, btw, once it is in Rust, should we have a Web-based Globulation?

kylelutze:
I was going to work on mobile first

stephanemagnenat:
Going through the web might address both desktop and mobile. That is what we do with [Candli](https://cand.li) and, while not perfect, works quite well.

Giszmo:
How is this effort going? Did you get to do the rust rewrite? I'm tempted to fix the bugs in glob2 that you identified but would love to see glob2 revived with a web/mobile version, too.

Giszmo:
Took the branch for a spin on Debian 13 (trixie), GCC 14.2. Good news first: with one fix it builds and runs well — full headless AI matches complete (gd-large-4ai: 42k ticks, clean GLOB2_GAME_END), GUI and replay playback work, and I got identical sim results from GCC 14 and GCC 15 builds, so determinism holds across toolchains. Findings:

1. Build breaks on Linux/libstdc++: `src/net/irc/IRCTextMessageHandler.cpp` needs `#include <algorithm>` for `std::find` (works on macOS libc++ by transitive include).

2. Startup crash (SIGILL/SIGSEGV depending on compiler) in non-OpenGL builds: `Sprite::createTextureAtlas()` has no return statement when `HAVE_OPENGL` is undefined — falling off the end of a bool function is UB, and -O3 compiles it into a trap. Configure disables GL silently if libepoxy/GLU dev packages are missing, so this is easy to hit without noticing. Suggested fix: `#else return false;` before the `#endif`. Note this also affects master (came in with #120), so it probably deserves a standalone issue + cherry-pickable fix.

3. Exclusive fullscreen at non-native resolution is broken under Wayland (screenshot attached): UI laid out for the requested mode, but XWayland's emulated mode-switch provides a desktop-sized surface. Consider SDL_WINDOW_FULLSCREEN_DESKTOP + logical scaling instead of exclusive fullscreen.

4. The test plan references `games/G2.game`, which isn't in the repo — the `games/gd-*.game` baselines work; maybe update the PR text or add G2.game.

5. Minor: with $HOME unset (CI/containers), instant abort via unchecked `getenv("HOME")` in FileManager — worth hardening given the headless-trainer use case.

Happy to test follow-up pushes on Linux.

kylelutze:
Thank you for checking it out! I certainly have more cleanup to do on it. If you want to add any fixes to the branch I welcome it. I got caught up with real life again and haven't had time to keep working on this. There are still about 300 known bugs to fix too and new models are out so I'll probably need to re-run the code review prompt with fable at some point.

I just pushed everything which also pushed the AI I was working on. I didn't feel like waiting until porting it to rust as my first attempt at that did not go well at all.

genixpro:
I have also been reloading this project. I've been playing around with approaches to make nicowar better lol two decades later i still want to make that AI play it better. I stumbled upon this paper and I'm a tad jealous they claim to beat "all built in AI's" and it was from around the time i was contributing. https://www.um.edu.mt/library/oar/bitstream/123456789/22897/1/cig08_1.pdf i was too young to even know there was academic research involving my code.

I am curious about this rewrite branch and what i can do to contribute.

kylelutze:
So this PR went a little bit in a different direction than what I was intending and has gotten huge. I've been playing it on my mac and I don't think I've added any regressions to the core game, but I don't know about platform specific. I have fixed a few bugs that have broken past game replays. Considering how long the game has been dormant I don't think that should be considered a problem. There were some determinism bugs and uint8 uses, like with fog, that caused overflow issues.

I also was working on AI Cortex. my goal was to see if I could get ML to train a better AI, but even using basic algorithms I haven't been able to reliably beat nicowar which was my first benchmark.

At some point we should decide on what to do with this branch considering how huge it is, etc. It was not my intent, but that's where it's at now.

genixpro:
> So this PR went a little bit in a different direction than what I was intending and has gotten huge. I've been playing it on my mac and I don't think I've added any regressions to the core game, but I don't know about platform specific. I have fixed a few bugs that have broken past game replays. Considering how long the game has been dormant I don't think that should be considered a problem. There were some determinism bugs and uint8 uses, like with fog, that caused overflow issues.
> 
> I also was working on AI Cortex. my goal was to see if I could get ML to train a better AI, but even using basic algorithms I haven't been able to reliably beat nicowar which was my first benchmark.
> 
> At some point we should decide on what to do with this branch considering how huge it is, etc. It was not my intent, but that's where it's at now.

If I were you, I would task codex on breaking this up into a bunch of discrete PR's that are easier to review on their own.

Giszmo:
I see review requests trickle in and merged a bit into master but at this point I don't mind going in bigger chunks. I plaid on my branch which is a ton of commits on top of yours with many bug fixes relevant to master, too ... I don't want to go back to master and even if we broke something I'm afraid of losing momentum.

I'm throwing Fable 5.1 at these reviews and bugfixes and trust it's doing a great job and especially with your LLM checking my work and the reverse, we gave the game a great leap forward.

kylelutze:
Once the build tests pass here, I'm down to get this merged and switch to doing smaller PRs going forward to continue the improvements. Or we can just keep working off this branch fixing everything.

Giszmo:
Guys, I'm a bit impatient. @kylelutze you started getting stuff merged to master in smaller pieces and I just created a ton of PRs in the same vain, too but it feels so pointless. Can't we just merge this and #132 and go on from there? What could possibly go wrong? I know on my end master is broken as is.

PR #131, opened 2026-08-31
Modernize the build: SCons/Python 3, zlib fixes, boost to std

kylelutze:
First slice of #129. That PR is too large to review, so it is being split into a stack of smaller PRs merged bottom-up. This is the base of that stack — it is the first four commits of the branch, unmodified, so no rebasing or cherry-picking was involved.

## What's here

| Commit | Change |
|---|---|
| `d6a40f1d` | `update scons usage` — Python 3 / modern SCons compatibility; drops the hand-rolled Darwin `TOOL_BUNDLE` bundle builder in favour of SCons' built-in support |
| `728adc1f` | `fix zlib usage bugs` — error handling in `FileManager`, `StreamBackend`, `StreamFilter`, `MapThumbnail` |
| `eeb66980` | `switch from boost threads to std threads` |
| `61cf0970` | `Replace boost::shared_ptr, lexical_cast, and tuple with C++11 equivalents` |

The boost work is mechanical substitution: 22 `boost/shared_ptr.hpp`, 4 `boost/thread/thread.hpp`, 3 `boost/tuple/tuple.hpp`, 3 `boost/thread/recursive_mutex.hpp`, 2 `boost/lexical_cast.hpp`, 1 each of `weak_ptr.hpp` and `tuple_comparison.hpp` swapped for `<memory>`, `<thread>`, `<mutex>`, `<tuple>`, `<string>`. 186 files, +1290/−1226, and the line counts are dominated by include churn.

## Notes for the reviewer

- **This does not remove the boost dependency.** It removes the smart-pointer, thread, `lexical_cast` and `tuple` surface. `SConstruct` still links `-lboost_date_time`, and `libgag/src/Sprite.cpp`, `src/Utilities.h`, `src/AIEcho.h` and several YOG headers still include boost. Dropping boost entirely is not in scope here.
- **`CLAUDE.md` is added by `d6a40f1d`** (107 lines, agent/contributor build notes). It rode along with the SCons commit. Happy to strip it into its own commit or drop it if you'd rather it not land here — say the word.
- **`test/SConstruct`** gets a one-line fix: `sys.platform == "linux2"` never matches under Python 3 (which reports `linux`), so the test build fell through to the MSVC `else` branch on every platform. This fixes Linux. macOS still falls through to `/EHsc /MD /GR` and does not build — pre-existing, fixed later in the stack.

## Test plan

- [x] `scons release=1 -j16` clean build on macOS arm64 (Darwin 25.6) — exit 0, no errors
- [x] Runtime smoke: `./build/src/glob2 -test-games-nox` ran two complete AI games to natural end (26,594 and 23,650 ticks) across Castor / Nicowar / Numbi / Warrush, with no asserts, aborts, or uncaught exceptions — exercises the swapped threading and smart-pointer paths under load
- [ ] *Not run:* the deterministic replay-diff harness does not exist at this point in history (it lands later in the stack), so there is no same-seed A/B available for this slice. `-test-games-nox` seeds from `time(NULL)` and cannot be pinned here.
- [ ] *Reviewer:* Linux build — the SCons and Darwin-bundle changes are the parts most likely to regress a platform I can't test
- [ ] *Reviewer:* confirm you want `CLAUDE.md` in this repo


libgag/src/FileManager.cpp
kylelutze, line 1: One of the few changes part of my first set of commits not related to   modernizing. It's doing some hardening against previously unchecked failures.

CLAUDE.md
kylelutze, line 1: Since this doesn't break the code, I would like to commit this as-is and have the future commit changes drop in on top of this instead of having extra commits that deviate this PR from the bigger PR

SConstruct
```diff
@@ -121,51 +121,31 @@ def configure(env, server_only):
...
     elif conf.CheckLib("boost_date_time-mt") and conf.CheckCXXHeader("boost/date_time/posix_time/posix_time.hpp"):
-        boost_thread="boost_thread-mt"
+        boost_date_time="boost_date_time-mt"
     else:
-        print("Could not find libboost_date_time or libboost_date_time-mt or boost/thread/thread.hpp")
+        print("Could not find libboost_date_time or libboost_date_time-mt or boost/date_time/posix_time/posix_time.hpp")
         missing.append("libboost_date_time")
     env.Append(LIBS=[boost_date_time])
-    env.Append(LIBS=["boost_system", "pthread"])
+    if conf.CheckLib("boost_system"):
```
Giszmo, line 143: Minor/non-blocking: this now only appends `boost_system` to `LIBS` if `conf.CheckLib("boost_system")` succeeds; every other dependency check in this function adds to `missing[]` and prints a "could not find" message on failure, but this one silently continues if the lib isn't found. Probably harmless on modern Boost (`system` is header-mostly there), but worth a one-line comment explaining why this one dependency is optional, or make it consistent with the others.

libgag/src/GraphicContext.cpp
```diff
@@ -2138,16 +2145,29 @@ namespace GAGCore
...
-			fprintf(stderr, "Toolkit : can't set screen to %dx%d at 32 bpp\n", w, h);
+			fprintf(stderr, "Toolkit : can't create window %dx%d\n", w, h);
 			fprintf(stderr, "Toolkit : %s\n", SDL_GetError());
 			return false;
 		}
+		// SDL_GetWindowSurface is incompatible with SDL_WINDOW_OPENGL;
+		// in GPU mode, create a small dummy surface so format-dependent code works.
+		if (optionFlags & USEGPU)
+		{
+			sdlsurface = SDL_CreateRGBSurface(0, w, h, 32,
```
Giszmo, line 2170: Leak: in `USEGPU` mode, `sdlsurface` is reassigned to a freshly-allocated `SDL_CreateRGBSurface(...)` every time `setRes()` runs, with no `SDL_FreeSurface()` on the previous value first. The non-GPU branch (`SDL_GetWindowSurface`) is fine since that surface is owned by SDL and freed with the window, but this dummy surface is heap-allocated here and orphaned on every resolution change / fullscreen toggle. Suggest freeing the old `sdlsurface` before this assignment when it was GPU-mode-created (or track ownership explicitly so the free doesn't double-free an SDL-owned surface).

libgag/src/FileManager.cpp
```diff
@@ -426,13 +426,23 @@ namespace GAGCore
...
-		gzwrite(gzStream, &buffer[0], fileLength);
-		
+		if (gzStream == NULL)
+		{
+			std::cerr << "FileManager::gzip: gzdopen failed for " << dest << std::endl;
+			fclose(destStream);
+			delete srcStream;
+			return false;
+		}
+		int written = gzwrite(gzStream, &buffer[0], fileLength);
```
Giszmo, line 436: `gzwrite(...)` returns 0 both on error *and* when `fileLength == 0` (a legitimate empty-file write). Combined with `return (written > 0)` below, `gzip()` on an empty input file will always report failure even though nothing actually went wrong. Suggest comparing `written == fileLength` instead of `written > 0`, and only logging when `written == 0 && fileLength != 0`.

libgag/src/FileManager.cpp
```diff
@@ -426,13 +426,23 @@ namespace GAGCore
...
+		if (written == 0)
+			std::cerr << "FileManager::gzip: gzwrite failed for " << dest << std::endl;
+
 		// Close
-		gzclose(gzStream);
+		if (gzclose(gzStream) != Z_OK)
```
Giszmo: A failed `gzclose()` here is logged but doesn't affect the function's return value (`return (written > 0)` a few lines down, unrelated to this check). This is exactly the point where zlib flushes its buffered deflate stream, so a close failure here can mean a truncated/corrupt output file while the caller still sees `true`. Same pattern is repeated in `gunzip()` just below (`return (bytesRead >= 0)` is independent of its own `gzclose` check).

libgag/src/StreamBackend.cpp
```diff
@@ -50,8 +58,16 @@ namespace GAGCore
...
 			buffer->seekFromStart(0);
 			gzFile fp = gzopen(file.c_str(), "wb9");
-			gzwrite(fp, buffer->getBuffer(), size);
-			gzclose(fp);
+			if (fp == NULL)
+			{
+				std::cerr << "ZLibStreamBackend: failed to open " << file << " for writing" << std::endl;
+				return;
+			}
+			int written = gzwrite(fp, buffer->getBuffer(), size);
```
Giszmo, line 66: Same empty-buffer-looks-like-error issue as `FileManager::gzip` (`gzwrite` returning 0 for a 0-byte `size` isn't actually an error) — cosmetic here since it only logs to stderr and doesn't change a return value, but worth noting since it's the same bug pattern in two places now.

src/IRCTextMessageHandler.cpp
```diff
@@ -21,31 +21,29 @@
 
 #include "IRCTextMessageHandler.h"
 #include "IRCThreadMessage.h"
+#include <functional>
```
Giszmo, line 25: **Blocker — doesn't compile on Linux/libstdc++.** `<algorithm>` isn't included here (only `<functional>` was added), but `removeTextMessageListener()` at line 112 uses `std::find(listeners.begin(), listeners.end(), listener)`. It apparently built on macOS/libc++ where `<algorithm>` gets pulled in transitively, but fails on GCC/libstdc++:
```
src/IRCTextMessageHandler.cpp:112:34: error: no matching function for call to 'find(...)'
```
Verified locally with `g++ -std=c++20 -fsyntax-only`. One-line fix: add `#include <algorithm>` alongside `<functional>` here.

src/YOGServerPasswordRegistry.cpp
```diff
@@ -130,6 +130,6 @@ std::string YOGServerPasswordRegistry::transform(const std::string& username, co
 	SHA1Final(digest, &context);
 	std::string final = "";
 	for(int i=0; i<20; ++i)
-		final += boost::lexical_cast<std::string>(digest[i]) + "-";
+		final += std::to_string(static_cast<int>(digest[i])) + "-";
```
Giszmo: **Blocker — silently breaks every existing YOG account's password.** `digest` is `unsigned char digest[20]` (line 129 above). `boost::lexical_cast<std::string>(unsigned char)` streams the byte through `operator<<`, which for `unsigned char` inserts the *raw byte as a character* (not its decimal value). `std::to_string(static_cast<int>(digest[i]))` instead emits ASCII decimal digits — a different string for every digest byte outside the range that happens to already look like a 1-byte decimal number. Every password hash computed and stored with the old code will fail to verify against this new transform. This needs either exact byte-for-byte backward compatibility (e.g. `salted += static_cast<char>(digest[i]);` to match `lexical_cast`'s actual behavior) or a versioned hash format with a migration path — not just a mechanical `to_string` swap.

src/YOGServerAdministratorCommands.cpp
```diff
@@ -83,9 +84,9 @@ bool YOGMutePlayer::doesMatch(const std::vector<std::string>& tokens)
 	{
 		try
 		{
-			boost::lexical_cast<int>(tokens[2]);
+			std::stoi(tokens[2]);
```
Giszmo, line 87: Should-fix: this only catches `std::invalid_argument`, but `std::stoi("99999999999999")` throws `std::out_of_range`, which isn't caught here — it propagates out of `doesMatch()` uncaught and crashes the server (e.g. `.mute_player somebody 99999999999999`). The old `boost::bad_lexical_cast` covered both failure modes in one catch; suggest `catch (const std::exception&)` (or catching both `std::invalid_argument` and `std::out_of_range`) to restore that coverage. This is the only `stoi` conversion in this file so the blast radius is just this one command.

Giszmo (changes requested):
Independent review of this PR (cloned the branch, built the failing translation unit with g++, checked types by hand rather than trusting the diff at a glance). Filing as **Request changes**: three of the four commits here are solid — the SCons/Py3 modernization, the boost→std thread conversion (a real improvement: replaces racy busy-wait polling in destructors with proper `join()`), and most of the zlib error-handling hardening. But the last commit (`61cf0970`, boost→std types) has two blockers:

1. **Doesn't compile on Linux/libstdc++** — missing `#include <algorithm>` in `IRCTextMessageHandler.cpp` (verified with a real g++ build, see inline comment).
2. **Breaks every existing YOG server password** — the `lexical_cast`→`to_string` swap in `YOGServerPasswordRegistry.cpp` changes the hash's byte-to-string encoding, invalidating all previously stored password hashes (see inline comment for the exact mechanism).

Both are small, mechanical fixes, not a rethink — the split-into-a-stack approach for reviewing #129 is exactly right and this slice is close. Detailed comments inline below; also flagged one should-fix (an uncaught `std::out_of_range` that can crash the YOG server via `.mute_player`) and a couple of minor/non-blocking notes.

— Posted by Junior (Leo's agent) on Leo's behalf.

SConstruct
kylelutze: Added a comment: Boost.System is header-only from 1.69 on, so the check stays optional.

libgag/src/GraphicContext.cpp
kylelutze: Fixed. `freeDummySurface()` frees it only in GPU mode, and runs before `optionFlags` is overwritten.

libgag/src/FileManager.cpp
kylelutze: Fixed. Went with `written == fileLength`.

libgag/src/FileManager.cpp
kylelutze: Fixed in both gzip() and gunzip().

libgag/src/StreamBackend.cpp
kylelutze: Fixed.

src/IRCTextMessageHandler.cpp
kylelutze: Fixed. Added GitHub Actions builds to validate going forward.

src/YOGServerPasswordRegistry.cpp
kylelutze: Fixed. Went with your raw-byte suggestion, so stored hashes keep verifying.

src/YOGServerAdministratorCommands.cpp
kylelutze: Fixed. Catches `out_of_range` explicitly rather than widening to `std::exception`.

Giszmo (approved):
Both blockers from the first review are fixed and the should-fixes too:

- `IRCTextMessageHandler.cpp` now includes `<algorithm>`; the build is green on Ubuntu 22.04, 24.04 and mingw-w64.
- `YOGServerPasswordRegistry::transform` appends each digest byte as a raw char again, which is what `lexical_cast<std::string>(unsigned char)` produced, so stored hashes keep matching.
- `.mute` catches `std::out_of_range` next to `invalid_argument`, so an oversized token no longer takes the server down.
- `gzip()` returns true only when `gzwrite` wrote the whole buffer and `gzclose` succeeded; the empty-file case (0 == 0) is handled by the same comparison. `gunzip()` and `ZLibStreamBackend` report close and short-write failures.
- The GPU-mode dummy surface is released in the destructor and before each `setRes`.

Things I checked on the new commits, all fine:

- `_MSC_VER` → `_WIN32` for `epoxy/wgl.h` is right for mingw.
- SConstruct: `Tool('mingw')` before the cross settings, `pkg-config sdl2` on every platform, the libsystre fallback for regex, and the removal of the MSYS 1.0 paths. Dropping `-Dmain=SDL_main` is fine because SDL2's `SDL_main.h` does that itself once `SDL.h` is included in the translation unit that defines `main`, and the Windows CI job links.
- CI workflow covers glob2, the server target and the cppunit tests on Linux; Windows builds glob2 and the server.
- `test/SConstruct`: dropping `-ansi` is required for the C++20 sources.

One note, not blocking: `freeDummySurface()` only frees under `USEGPU`. #132 also allocates an offscreen surface for software fullscreen scaling; when that lands on top of this, the helper needs to cover both, which #132 already does on its side.

PR #132, opened 2026-09-01
Fix fullscreen rendering when window size differs from logical resolution

Giszmo:
Fixes the fullscreen rendering bug found while testing Kyle's `feat/ai-trainer-support` branch on Wayland/XWayland: the window is correctly sized to the desktop via `SDL_WINDOW_FULLSCREEN_DESKTOP` (mode-switching doesn't work under Wayland), but nothing scaled the game's drawing to that window size, so the UI — laid out for the requested logical resolution — was clipped or missing on any screen bigger than that resolution.

Root-caused and originally patched by Bob (@nostr21/buzz agent); this PR packages that fix so it's mergeable. See the [buzz channel discussion](https://buzz.nostr21.net) for the original investigation (screenshot of the broken menu, three-bugs breakdown).

## What changed

- `GraphicContext` now tracks the actual window pixel size separately from the logical resolution, renders to a logical-size offscreen surface when they differ, and blits it scaled to the window each frame (`SDL_BlitScaled` in software mode; GL viewport + `SDL_GL_GetDrawableSize` in GPU mode).
- `glScissor`'s clip rect is scaled from logical to window coordinates so GPU-mode clipping stays correct under scaling.
- Mouse coordinates from SDL events and `SDL_GetMouseState` are translated from window pixels back to logical coordinates everywhere they're read, so clicks/hover still line up with the logical-resolution UI.

## Test plan

- [x] `scons release=1` clean build (Debian, GCC 14.4)
- [x] Headless regression: two `-test-games-nox` AI games run to natural end (90000 and 47010 ticks), no asserts/crashes — exercises the changed GraphicContext/GUI event paths under load
- [x] Fullscreen scaling reproduced end-to-end under Xvfb + Openbox: forced the window to a 1280x800 desktop size while requesting `-s 960x540 -f -G`. With this fix the full menu renders scaled to fill the window with no clipped/missing widgets, and a mouse warped to a raw window-pixel coordinate correctly highlights the button under the cursor — confirms the coordinate translation, not just the visual scaling.
- [x] *Not tested here:* real Wayland/XWayland session (the original repro environment) — Leo, please confirm this fixes what you saw on your box.


Giszmo:
Pushed a follow-up commit: switched the fullscreen software blit from `SDL_BlitScaled` (nearest-neighbor) to `SDL_SoftStretchLinear` (bilinear, SDL 2.0.16+), falling back to the old nearest-neighbor blit if it fails.

**Why:** testing on non-integer scale factors (e.g. 960x540 logical → 1280x800 desktop) showed nearest-neighbor duplicating source rows/columns unevenly — visible as hard grid lines at the menu background's 32x32 grass tile seams, and inconsistent blockiness in progress-bar marker lines. Bilinear smooths that out.

**Not fixed by this:** at large scale factors (e.g. a 4K desktop with a much lower requested resolution), thin 1px chart/graph lines will still read as faint — that's inherent to rendering UI at a fixed logical resolution and magnifying it, not something a better filter alone fixes. Flagging as a known limitation, not silently claiming it's resolved.

Giszmo:
Reverted the bilinear-scaling commit above — real-machine testing (OpenGL off / software render path) showed it corrupts the screen to mostly black, not just a cosmetic improvement. Likely `SDL_SoftStretchLinear` blending using the offscreen surface's alpha channel, which isn't reliably opaque everywhere we draw onto it; `SDL_BlitScaled` (nearest-neighbor, what's live now again) ignores alpha and just copies RGB.

Current state of this branch: the core fullscreen-scaling fix (first commit) only, nearest-neighbor blit for the software path. Tile-seam/hairline-blockiness cosmetic issues are still present but the display is correct and complete, which matters more.

Giszmo:
Pushed another follow-up: scale `glLineWidth` in the GPU/OpenGL rendering path to match fullscreen scaling.

**Why:** `GraphicContext::drawLine` (and so `drawVertLine`/`drawHorzLine`) draws a native `GL_LINES` primitive at the default `glLineWidth(1)` — fixed in framebuffer pixels, doesn't grow with `glViewport`. Filled geometry scales because the fragments stretch with the viewport transform; a GL line's thickness doesn't. So hairline UI elements (e.g. the per-unit training-progress marker in `GameGUIDrawBuildingHelpers.cpp`) render as a properly-weighted blended marker in software mode (whole frame gets blitted+scaled, thickness included) but as a thin, faint hairline in GPU mode (only geometry scales, not line width) — confirmed via a side-by-side screenshot comparison from Leo.

Verified: clean rebuild, ran through menu/settings/in-game skirmish under GPU mode with scaling forced (2.4x) — no crashes or visual corruption on any line-drawn UI. Have not yet visually confirmed the specific training-bar widget this was reported against (needs an actively-training building to set up); asked for a follow-up screenshot after this lands.

Giszmo:
@kylelutze this is growing a bit out of hands here but it's fixing many many bugs I encountered on your branch and some even on master.

kylelutze:
Reviewed the diff. One real bug, plus some smaller issues.

## Point/pixel mismatch on HiDPI in GL mode

In `setRes`, `windowW/H` is first set from `SDL_GetWindowSize` (window **points**), then in the GL branch overwritten with `SDL_GL_GetDrawableSize` (**pixels**). Mouse events arrive in window points. On a macOS Retina display (drawable = 2× window size), `windowToLogical` divides mouse coordinates by the drawable size, putting every click at half its real position. The drawable size is the right thing for `glViewport`, `glScissor`, and line widths, but mouse translation needs the window size — these need to be two separate pairs of fields.

## Smaller issues

- The `GL_NEAREST` mag-filter switches and the 1/64 texel-inset UV change in `GraphicContextCompound.cpp` are applied unconditionally, not gated on scaling being active, so they alter GL rendering for everyone at 1:1 (anywhere sprites are magnified). The inset is also a magic constant tied to assumed texel density. Either gate these on `isScalingActive()` or call them out as deliberate always-on changes.
- The `DrawableSurfaceCompound.cpp` readback now reads the entire framebuffer and runs a per-pixel box-average triple loop even in the unscaled case (n=1), where a straight row copy suffices. Also inconsistent: the partial-rect `else` branch still calls `glReadPixels` with unscaled logical coordinates, so it reads the wrong region whenever scaling is active.
- Software-scaling `setRes` creates a new offscreen surface on every call and never frees the previous one, so every settings change in fullscreen leaks one.
- `windowW/H` only updates inside `setRes`; a compositor-driven resize (Wayland can do this) that doesn't route through `setRes` leaves the translation stale. Worth confirming `SDL_WINDOWEVENT_SIZE_CHANGED` handling covers it.
- The multi-pass `drawLine` double-blends where the overlapping 1px smooth lines cross, so translucent lines come out darker than intended.
- `GraphicContextDraw.cpp` gains `#include <algorithm>` twice, plus both `<cmath>` and `<math.h>`.
- `seamless_grass.py` is missing the SPDX line, and its docstring describes the old tiles rather than what the script does now.
- The `Game_io.cpp` integrity change touches state healed at load time — please run the replay-checksum A/B (Workflow 2 in `docs/replay-verification.md`) rather than relying on the headless games not crashing.

Giszmo:
Thanks, most of that was real. Fixed on the branch, one commit each:

- **HiDPI points vs pixels**: `windowW/H` now stays the window size SDL reports mouse coordinates in, and a separate `drawableW/H` feeds `glViewport`, `glScissor`, the line width and the framebuffer readback. The pixel-side users scale by the drawable unconditionally, which is the identity on a normal display at 1:1 (06fbda63).
- **Surface leak**: the GPU dummy surface and the software-scaling offscreen surface are freed before each `setRes` and in the destructor; the window surface stays SDL's (af9a0128). When this lands on top of #131's `freeDummySurface()`, the two collapse into one helper.
- **Compositor resize**: `SDL_WINDOWEVENT_SIZE_CHANGED` was only forwarded as `SCREEN_RESIZED` with the logical size. The event hook every poll site already goes through now re-reads the window and drawable sizes, resets the GL viewport, and in software mode switches to an offscreen surface when the window stops matching the logical size (eecde411).
- **Readback at 1:1**: plain row copy when the drawable equals the logical surface; the box average only runs when scaled (cd60923e). The partial-rect branch does not call `glReadPixels` on this branch; it asserts as before, so there is no unscaled read there.
- **drawLine double blend**: each pass now carries `1 - (1 - a)^(1/passes)`, which composes back to the requested alpha where the passes overlap; opaque lines unchanged (c5a9dd9b).
- Duplicate includes (dead456b), SPDX line and docstring for `seamless_grass.py` (6b6019da).

Deliberate and unchanged: `GL_NEAREST` and the texel inset are always on. The goal was to make GL render the tile art the way the software renderer does, at 1:1 as well, and the texel-centre sampling is what stops atlas sprites from seaming whenever they are magnified, which happens at 1:1 too on the end-game chart and in the editor zoom. Happy to gate them if you would rather keep the old GL look at 1:1.

Replay checksum A/B for the `Game_io.cpp` change: `docs/replay-verification.md` is not in this tree, so I ran the four `gd-*` baseline games for 3,000 ticks with `GLOB2_CHECKSUM_SIDECAR=1` on the branch base and on the branch; the sidecars and the replays are byte-identical for all four. The two load-time changes only alter saves that the old code refused (unit entering a building) or healed wrongly (building footprint straddling the map edge), and the second is the point of the fix, so those differ by design.

Rebased onto feat/ai-trainer-support at 2b7eb760.

kylelutze (approved):

PR #134, opened 2026-09-03
Decouple the custom cursor from the 40ms game loop

Giszmo:
Fixes the "custom cursor is lagging" report: `CursorManager::draw()` blitted the cursor sprite onto the screen surface, and was only called once per simulation tick (`GAME_TICKS_PER_SECOND = 25`, 40ms, in `Engine::drawAndPaceFrame()`). Movement only visually updated at that cadence, well below a compositor's refresh rate and without the near-zero latency an OS-drawn cursor gets from being composited independent of the app's render loop.

`CursorManager::update()` replaces `draw()`: instead of blitting each tick, it composites the current frame into an offscreen surface (still via `drawSprite()`, so team-color tinting through `Sprite::getRotatedSurface()` is unchanged) and installs it as a native cursor with `SDL_CreateColorCursor()` / `SDL_SetCursor()`. The OS then moves and composites it at its own rate, decoupled from the 40ms tick entirely. `SDL_SetCursor()` is still only called when the type, animation frame, or color actually changed -- animated cursors (the wait spinner) still update roughly every 40ms, same as before, just via a cursor swap instead of a redraw, per the interaction in the linked chat. The system cursor is left enabled instead of hidden, since our image now *is* the system cursor.

One correctness fix along the way: `~CursorManager()` frees the native cursor via `SDL_FreeCursor()`, but as a plain member of `GraphicContext` its destructor would normally run *after* `~GraphicContext()`'s body, which calls `SDL_Quit()` first -- freeing an SDL cursor after `SDL_Quit()` is undefined. Added `releaseNativeCursor()` and call it explicitly at the top of `~GraphicContext()`, before `SDL_Quit()`.

**Testing:** compiles and links clean (verified against both this branch's base and, since master's `SConstruct` currently links `-lboost_system` which isn't available in this sandbox's Boost 1.90 -- a pre-existing, unrelated gap -- against a locally-patched link config as a sanity check only, not committed). No display available in this environment to exercise the GUI path directly; the change mirrors the original `draw()`'s state machine exactly (same type/frame reset logic) so animation and edge-hover behavior should be identical, just cursor-swap instead of blit.

### DIFF
--- a/libgag/include/CursorManager.h
+++ b/libgag/include/CursorManager.h
@@ -21,13 +21,16 @@
 #define __CURSOR_MANAGER_H
 
 #include <vector>
+#include "SDL.h"
+
+struct SDL_Cursor;
 
 namespace GAGCore
 {
 	class Sprite;
 	class DrawableSurface;
 	struct Color;
-	
+
 	//! A support class to manage cursors
 	class CursorManager
 	{
@@ -51,7 +54,7 @@ namespace GAGCore
 			CURSOR_MARK = 11,
 			CURSOR_COUNT
 		};
-	
+
 	protected:
 		//! a vector of loaded cursors
 		std::vector<Sprite *> cursors;
@@ -61,10 +64,25 @@ namespace GAGCore
 		CursorType nextType;
 		//! the current frame of cursor sprite.
 		int currentFrame;
-		
+		//! draw color of the cursors, packed so update() can tell it changed
+		//! without needing Color's full definition here
+		Uint32 packedDrawColor;
+		//! the native cursor currently installed via SDL_SetCursor, owned by us
+		SDL_Cursor *activeCursor;
+		//! (type, frame, color) that activeCursor was built from
+		CursorType activeType;
+		int activeFrame;
+		Uint32 activePackedColor;
+		bool activeValid;
+
 	public:
 		//! Constructor, set default values
 		CursorManager();
+		//! Frees the native cursor, if any
+		~CursorManager();
+		//! Releases the native cursor, if any. Must be called before SDL_Quit();
+		//! ~CursorManager() runs too late for that (after ~GraphicContext()'s body).
+		void releaseNativeCursor(void);
 		//! Load the cursor sprites
 		void load(void);
 		//! Select the next type given the mouse position
@@ -75,8 +93,10 @@ namespace GAGCore
 		void setDrawColor(const Color& color);
 		//! Sets the default draw color
 		void setDefaultColor();
-		//! Draw the current cursor with its current frame at a given pos
-		void draw(DrawableSurface *ds, int x, int y);
+		//! Advance the cursor animation and, if the type/frame/color changed,
+		//! install a native cursor for it via SDL_SetCursor. The OS then moves
+		//! and composites it independent of our own render/tick rate.
+		void update(void);
 	};
 }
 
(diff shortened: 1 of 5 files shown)


Giszmo:
Closing in favor of the scale-aware version on `fix/fullscreen-scaling` (ce5e6038): this branch's native-cursor approach rendered at 1x regardless of window scale, near-invisible once combined with fullscreen scaling. Same fix, plus `drawableScale()`-aware compositing so it matches the rest of a scaled-up window.

libgag/src/GraphicContext.cpp
```diff
@@ -2213,14 +2216,10 @@ namespace GAGCore
...
-				// disable system cursor
-				SDL_ShowCursor(SDL_DISABLE);
-				// load custom cursors
+				// cursorManager installs its cursors as native ones (see
+				// CursorManager::update()), so the system cursor stays on
 				cursorManager.load();
-			}
-			else
-				SDL_ShowCursor(SDL_ENABLE);
+			SDL_ShowCursor(SDL_ENABLE);
```
kylelutze, line 2222: Turning the custom cursor off in Settings no longer restores the OS arrow. `updateGfxCtx()` calls `setRes()` with `CUSTOMCURSOR` cleared, `nextFrame()` stops calling `update()`, but the last `SDL_SetCursor(newCursor)` stays installed (`activeCursor` still owns it) — nothing calls `releaseNativeCursor()` or resets to the default cursor until exit. The old `else SDL_ShowCursor(SDL_ENABLE);` branch handled this because the cursor was only ever drawn into the framebuffer.

`SDL_FreeCursor` on the active cursor reverts to the default one, so:

```suggestion
			else
				cursorManager.releaseNativeCursor();
			SDL_ShowCursor(SDL_ENABLE);
```

libgag/src/CursorManager.cpp
```diff
@@ -105,26 +127,51 @@ namespace GAGCore
...
+			return;
+		}
+
 		Sprite *sprite = cursors[static_cast<int>(currentType)];
-		ds->drawSprite(x-(sprite->getW(currentFrame)>>1), y-(sprite->getH(currentFrame)>>1), sprite, currentFrame);
+		int w = sprite->getW(currentFrame);
+		int h = sprite->getH(currentFrame);
+		DrawableSurface frame(w, h);
+		frame.drawSprite(0, 0, sprite, currentFrame);
+		SDL_Cursor *newCursor = SDL_CreateColorCursor(frame.getSDLSurface(), w>>1, h>>1);
```
kylelutze, line 163: Every cursor sprite is a 10–20 frame animation and `update()` bumps `currentFrame` on every call, so the `activeFrame == currentFrame` fast path never fires in steady state. That means every 40 ms, forever, even with the mouse idle: `SDL_CreateRGBSurface` (+ in GPU mode `glGenTextures` and a power-of-two `glTexImage2D` zero-fill, then `glDeleteTextures`), a blit, `SDL_CreateColorCursor` (an OS cursor object), `SDL_SetCursor`, `SDL_FreeCursor`.

The (type, frame) space is only 190 entries per color. Build the `SDL_Cursor` for each (type, frame) once and keep them in a per-type vector, invalidated by `setDrawColor()`/`load()`; then `update()` only does `SDL_SetCursor` when the frame changes.

libgag/include/CursorManager.h
```diff
@@ -21,13 +21,16 @@
 #define __CURSOR_MANAGER_H
 
 #include <vector>
+#include "SDL.h"
+
+struct SDL_Cursor;
```
kylelutze, line 26: `SDL.h` already declares `SDL_Cursor` (via `SDL_mouse.h`), so this forward declaration is dead; nothing else in `libgag/include` forward-declares SDL types.

```suggestion
```

PR #135, opened 2026-09-03
Fix --fwhole-program typo blocking release builds

Giszmo:
`SConstruct` passed `--fwhole-program` (double dash) to g++ under `release=1`; g++ only accepts the single-dash GCC spelling `-fwhole-program` and rejects the double-dash form as unrecognized, so `scons release=1` failed at the final link step. One-character fix. Found while Leo was trying to build PR #134 off master.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -303,7 +303,7 @@ def main():
 
     if env['release']:
         env.Append(CXXFLAGS=' -O3 -s')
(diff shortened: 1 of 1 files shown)


kylelutze (approved):
Correct — g++ only accepts the single-dash spelling.

PR #136, opened 2026-09-03
Split Building.cpp into six focused translation units

kylelutze:
Pure file split of the 2896-line `src/Building.cpp` into six files under `src/building/`, grouped by concern:

| File | Contents |
|---|---|
| `Lifecycle.cpp` | constructors, destructor, freeGradients, load/save, cross refs |
| `Construction.cpp` | resource needs/wishes, launch/cancel construction, updateCallLists, updateConstructionState |
| `Update.cpp` | building-site / unit / general updates, upgrade-area forbidden zones, isHardSpaceForBuildingSite |
| `Step.cpp` | step, subscribe* steps, subscribeUnitForInside |
| `TypeSteps.cpp` | swarmStep, turretStep, clearingFlagStep |
| `Misc.cpp` | kill, unit/resource management, exits, level, flag stats, happiness, conversion, integrity, checkSum |

Every code line is a byte-exact copy of the original; only the copyright/include prelude is duplicated per file. `Building.h` is unchanged. `SConstruct` gains `#src` on CPPPATH so the moved files resolve headers in `src/`, and the SConscript and Visual Studio project lists point at the new files.

Verified with a full `scons release=1` build on macOS.


src/building/Update.cpp
```diff
@@ -0,0 +1,508 @@
...
+		{
+			ratio[i]=1;
+			totalRatio++;
+			percentUsed[i]=0;
+		}
+	}
+	return isRoom;
+}
+
+#include "GameGUI.h"
```
Giszmo, line 391: should we move this to the top now?

Giszmo (approved):

PR #139, opened 2026-09-05
Fun experiment: an explorable toroidal world

genixpro:
Ready for review, with ongoing integration checks against `master`.

A rendering experiment that makes Glob2’s wrapping map visible as a live 3D torus. **G** explicitly switches between regular 2D and the optional torus view. Arrow keys, screen edges, the minimap and middle-drag navigate within the current mode; the wheel zooms within 3D. By default, ordinary scrolling never folds the map or snaps back. Main-menu **Settings → General Settings → Automatic torus view** enables gradual folding while scrolling and a quick return after movement stops; G keeps the overview open. The preference is saved locally and is inactive with software rendering. Normal 2D retains its viewport renderer, without full-world texture captures or torus mesh updates.

Following [Giszmo’s feedback](https://github.com/Globulation2/glob2/pull/139#issuecomment-5551736580), this incorporates the feature changes from `feat/torus-pan` through `b838f8de`, with attribution to Bob. Both map axes slide around a ring whose orientation stays fixed. The star field follows horizontal navigation only: vertical movement rolls the map around the tube without tilting the sky. The camera tilt follows the playable area’s aspect ratio and favors the front-facing landscape over keeping the torus hole visible. The original screen-center map position stays anchored throughout the transition. Wide maps use a flatter view; fitting favors this foreground focus and can crop the distant rim. Wheel zoom allows a closer view. Playtesting favored explicit mode switching as the default; automatic folding remains available as an opt-in preference. Both modes use the same renderer and performance fixes.

## Rendering and interaction

- Ring and tube radii follow map aspect ratio, with isothermal texture coordinates keeping tiles locally square on the surface. Tall maps make fat rings and wide maps make thin rings; tile area still varies around the tube. Camera fitting, clouds and picking share the geometry.
- Camera and input state live in `TorusView.cpp`; OpenGL drawing and resource ownership live in `TorusViewRender.cpp`.
- Terrain, units, buildings, shadows, fog of war, and previews reuse the existing map renderer. Clouds float on a separate surface and settle as the map flattens; cloud texture coordinates account for the map capture origin and texel centers.
- The shared building renderer draws the wrapped copies intersecting the view. Previously, its 16-tile viewport margin could move buildings outside a full-world capture. The correction also applies to software rendering.
- Unit drawing uses the visible wrapped map cell from the existing traversal. This fixes explorers and ground units being misplaced near full-world texture edges and preserves distinct copies at seams.
- Picking uses the rendered triangles, with double-precision barycentric calculations for narrow triangles in shallow views. Explicit view toggles finish active painting gestures; automatic transitions hold their projection during pointer gestures. Reload clears camera, picking, and GPU state. Middle-button release is handled before modal event processing.
- Saved keyboard layouts gain missing default actions only when their keys do not conflict with custom bindings or sequence prefixes.

Software rendering remains supported, including builds with `opengl=0` and startup with GPU mode requested. Unsupported configurations retain 2D. Camera state is not saved or sent over the network.

## Large-map performance

Profiling Oazis (256 × 256) on an Apple M3 identified redundant full-world cloud generation, individual resource sprite draws, and per-sprite GPU state changes in the batching path. The capture now computes shadows only, with overview cloud sampling bounded to 128 cells per world axis. Variable-size resource frames use a padded atlas, and queued sprite draws set GPU state at batch submission. Normal 2D retains native detail and software rendering.

The overview texture cap is 4096 pixels per axis, reducing the 256 × 256 map’s color texture from 256 MiB to 64 MiB. The live map still refreshes every rendered frame. In 100-frame rendering benchmarks with GPU completion included, the torus improved from about 130 ms to 12–16 ms per frame with clouds. These measurements exclude simulation, HUD drawing and frame pacing. The optimized build has also been playtested locally.

## Validation

Resource atlas comparisons cover every frame at three scales and two opacity levels, with zero pixel difference on macOS. A repeatable loaded-game performance benchmark is included. Synthetic 64 × 128, 512 × 64, 64 × 512 and square maps cover proportions, fitting, picking and map-dimension cache changes. The map-proportioned geometry measured 16.64 ms median / 17.87 ms p95 with clouds in a 150-frame Oazis run on the M3; desktop load affects these rendering-only measurements.

Local macOS validation covers OpenGL and software-only release builds, loaded-game rendering in both modes, cloud transitions, wrapped navigation, picking, explicit mode transitions, inactive torus controls during 2D navigation, explorer positions across viewport and atlas seams, automatic-mode timing and pinning, option persistence, keyboard conflicts, and repeated resource teardown/recreation. Geometry, picking, cloud-field and building-wrap tests pass AddressSanitizer/UndefinedBehaviorSanitizer.

CI builds Linux 22.04/24.04 and MinGW-w64 game/server targets and runs the standalone tests. Linux additionally exercises the loaded game through Mesa/Xvfb and SDL software rendering, builds without OpenGL, and checks startup and rendering with GPU mode requested. See the current PR checks for results on the latest commit.

## Earlier demo

This recording shows the original manual camera experiment. It predates the fixed-axis scrolling behavior and rendering corrections above.

https://github.com/user-attachments/assets/480b73cb-157a-4ebb-b93d-2f46a7e5092d

## Scope and limitations

This PR tracks `master` through merge commits. The AI trainer/refactoring and fullscreen-scaling PRs have now landed in `master` and are included through that base. The torus is adapted to the new GUI/render modules, viewer state and pending building positions. Conflicts and CI will continue to be maintained during review. Giszmo's native Wayland setup still needs playtesting to verify the combined behavior.

Compatibility OpenGL with shaders and framebuffer objects is required for 3D. Large maps may downsample at the 4096-pixel atlas cap and can cost more than ordinary 2D. Tile size varies around the tube and perspective foreshortens tiles, especially near the inner opening. Units/buildings remain sprites on the surface; unit picking uses map cells. Native Windows/Linux visual and broader gameplay testing remain useful.

[Build instructions, architecture, tests and limitations](https://github.com/Globulation2/glob2/blob/codex/toroidal-world/doc/TorusExperiment.md).


genixpro:
https://github.com/user-attachments/assets/7ec49c59-92c6-4177-aebe-eb6bd7785559

torus-overview
unfolding
map-2d

genixpro:
https://github.com/user-attachments/assets/480b73cb-157a-4ebb-b93d-2f46a7e5092d

frame-14
frame-12
frame-0

Giszmo:
details nack

I could come around to an ack if the torus main axis remained fixed like in feat/torus-pan with the surface sliding vertically as you already fixed horizontally. So when the mouse pushes a border or you press a direction key, the zoom out begins slowly, with the snap back being faster.

Actually, having tried it out ... I like that a lot. It's still a bit buggy with buildings disappearing and the clouds being torodial but not transitioning nicely between 2d and 3d and fixing all bug will probably make the 2d code obsolete except for it being needed for software rendering, so we would have to maintain it all in both 2d and 3d :exploding_head: 

Note I rebased on my bugfix branch as your branch is completely broken here and I don't want back to before kyle's massive efforts anyway.


https://github.com/user-attachments/assets/7735a560-099a-44f0-b428-51534c4026dd

Things keep improving on my branch ...

image

genixpro:
@Giszmo I’ve brought the fixed-axis navigation and cloud changes from `feat/torus-pan` (through `b838f8de`) into this PR, with credit to Bob in the port commit.

The ring keeps its orientation while the map slides around it in both directions. Edge/key scrolling pulls back slowly and returns quickly; middle-button panning uses the ordinary map movement path. The overview now fits the playable area, and the highlight is less intense.

I found a concrete disappearing-building bug: `mapCaseToDisplayable()` moves buildings in the last 16 tiles to negative viewport coordinates, potentially outside the full-world texture. The shared building renderer now draws the wrapped copies intersecting the view, including across seams. That correction applies to software rendering too.

Cloud coordinates now include the capture origin and texel-center offset, and their height reaches zero as the map flattens. Automatic return waits for selection/painting gestures to finish. Reload clears movement, held-button and picking state; missing default shortcuts respect custom key bindings.

For maintainability, camera/input state and OpenGL drawing are in separate files. Terrain, buildings, units, shadows and fog still use the existing map renderer. Software rendering remains supported, including builds without OpenGL and GPU mode requested in those builds.

macOS GPU/software builds and loaded-game rendering pass, as do the sanitized geometry/picking/cloud/building-wrap tests. [Linux 22.04/24.04 and MinGW-w64 CI passed on `31d26c9c`](https://github.com/Globulation2/glob2/actions/runs/33974416615), including Linux GPU rendering under Mesa/Xvfb and loaded-game software rendering. The last follow-up, `2ee8a381`, only updates documentation and trailing whitespace; its CI rerun is underway.

This PR remains master-based, with your torus changes adapted to that layout. Your native startup/display failure still needs a reproducer; these checks don’t establish that it’s resolved on your setup. The build command, launch arguments and failure output would help identify what remains.

genixpro:
@Giszmo Thanks for the feedback and the work on `feat/torus-pan`. I’ve incorporated the fixed-axis navigation and elevated cloud layer, with attribution to Bob, and addressed the rendering issues I could identify:

- Both map axes now slide around a ring with a fixed orientation. The star field follows horizontal movement only, so rolling the surface vertically no longer pitches the background.
- Buildings disappearing from the full-world texture were caused by the existing viewport wrapping logic. The shared building renderer now draws the wrapped copies intersecting the view, including sprite overhang; that fix also applies to software rendering.
- Explorers exposed the same viewport-position assumption in unit drawing. Units now use the visible wrapped map cell already supplied by the traversal, fixing their position near atlas edges and across seams. A regression check fails on the previous code and passes with this correction.
- Clouds share the world cloud field, with corrected capture coordinates and a separate elevated surface that settles as the view returns to 2D.

After further playtesting, I’ve kept 2D and 3D as separate modes, switched explicitly with **G**. The movement-triggered pullback and snapback interrupted ordinary scrolling. Normal 2D now stays on its regular viewport renderer without full-world captures or torus mesh updates; the torus remains optional. The separated modes felt much better in playtesting.

I’ve also cleaned up the implementation: camera/input state is separate from OpenGL drawing and resource ownership, the automatic folding state is removed, and both views reuse the game’s map renderer. Software rendering remains supported, including builds without OpenGL.

Local OpenGL and software rendering checks pass, along with the software-only build/startup checks. CI is running on the latest changes. The PR description and rendering notes now reflect this version.

This still uses the master baseline rather than merging the separate refactoring/fullscreen branches. I haven’t reproduced your native startup/display failure; could you share the error or steps that trigger it, and whether the latest version still fails on your setup? I’d appreciate another look at this version when you have time.

Giszmo:
@genixpro The startup/display failure is not torus-specific, it is master on a 3840x2160 Wayland desktop. `SDL_WINDOW_FULLSCREEN_DESKTOP` keeps the window at desktop size (Wayland cannot mode-switch) while the game keeps drawing at the requested logical resolution, so the main menu, laid out for the smaller size, ends up in a corner of the screen and the mouse does not line up with it. Master has never worked here since the SDL2 port for that reason, which is why all my testing happens on `feat/torus-pan`.

The fixes master lacks, in dependency order:

1. **#129 `feat/ai-trainer-support`** (kylelutze, 423 commits): the current file layout (`src/render/`, `src/gui/`, ...), the portability fixes for non-macOS and headless builds (da8e1fe3), missing includes (47ed78bd). Everything below sits on it.
2. **#132 `fix/fullscreen-scaling`** (33 commits on #129): this is the one that makes the menu render.
   - cd5c3b96 the scaling itself: window size tracked apart from the logical resolution, logical-size offscreen surface blitted with `SDL_BlitScaled` in software mode, GL viewport from `SDL_GL_GetDrawableSize` in GPU mode, `glScissor` scaled, mouse events translated back to logical coordinates.
   - 06fbda63 drawable size kept apart from window size (HiDPI), eecde411 sizes refreshed on resize, af9a0128 surface ownership.
   - f21f283a, 004b4e5c, 16acb6e3, c5a9dd9b GL line widths under scaling; ec1bca96 1 px lines as quads.
   - 8cd64746, 46244dde, 2740ba13 nearest sampling and texel-centre coordinates, so atlas sprites stop seaming when scaled.
   - a8bed8f4, cd60923e GL framebuffer readback of the whole scaled drawable (dialog backgrounds, screenshots).
   - ce5e6038, 2efde2e6, 358cd2b2 native, scale-aware cursor; e10432dc restart on renderer switch.
3. **`feat/torus-pan`** (19 commits on #132): the behaviour you already ported, plus two you do not have yet: 345631a5 units at the traversed cell and buildings at every wrapped copy (the same bug you found, fixed against the `src/render/` layout), and 6f67ab7c the sun on the far plane with sea glint and dull land, black kept black.

So: rebase `codex/toroidal-world` onto `fix/fullscreen-scaling` (or target it as the base) and the menu will be sane on this machine; on master it will keep failing whatever the torus does. As for the mode switch, I prefer the fold that starts when the view moves and snaps back when it stops, as on `feat/torus-pan`; explicit G remains available there too.

(written by Bob, my agent)

genixpro:
@Giszmo can we split the difference and make the torus zoom-out and snap-back mode configurable? I've found it very awkward in my testing

genixpro:
@Giszmo The two navigation styles are now supported through **Settings → General Settings → Automatic torus view** from the main menu. It is off by default, preserving explicit G switching. Enabling it gives gradual folding while scrolling and a quick return after movement stops; G keeps the overview open. The preference is saved locally, and automatic transitions wait for selection/painting gestures to finish. Software rendering stays in 2D.

Further profiling on a 256 × 256 map also exposed redundant cloud generation and sprite submission overhead. The rendering benchmark improved from about 130 ms to 12–16 ms per frame on the M3, with the map still refreshing every frame. GPU and software checks pass, including pixel comparisons for the resource batching change.

Thanks for explaining the Wayland failure. I understand that it belongs to master’s fullscreen scaling rather than this feature. I’d like to keep this PR independent and focused while #129 and #132 are still pending, rather than coupling it to that larger dependency chain. This leaves the Wayland issue outstanding here until the separate scaling work lands.

Giszmo:
You could also help make those other improvements land ;)

Giszmo (commented):
NACK while we fix bugs. Don't make the revival harder than it is with fun experiments. I fell for the temptation to play with that torus view and would advocate to switch it on by default but merging 3k LOC in a major refactoring effort is counter productive.

Giszmo:
Try your torus world with extreme map geometries. Out of the box a 64x128 map looks like this here:

image

For comparison, a 512x64 looks like this:

Screenshot From 2026-09-06 01-36-01

genixpro:
@Giszmo I've moved this PR to draft for now and merged current master, including the refactoring and fullscreen-scaling work now that those PRs have landed. The torus integration follows the new GUI/render modules and separate viewer and pending-building state. Explicit G switching remains the default, with automatic folding off by default in the main Settings menu.

The integration also caught a nearest-neighbour sampling mismatch between standalone and atlas resource sprites; that is fixed, with identical pixels at native, reduced and enlarged scales. Local OpenGL and software-only build/rendering checks pass, as do the broader test executables and the selection/terrain regressions. The 256 × 256 rendering benchmark remains about 12 ms per frame with clouds on the M3 (rendering only).

I'll keep the draft current with master and resolve integration conflicts as they arise. Native Wayland playtesting of the combined changes is still needed.

genixpro:
@Giszmo Thanks for the rectangular-map examples. Addressed in 1016f0b8: the torus ring/tube proportions now follow the map dimensions, with isothermal texture coordinates so tiles stay locally square on the surface. Wide maps produce thinner rings and tall maps produce fatter ones. Tile area still changes around the tube, and perspective still foreshortens it.

The geometry is shared by rendering, clouds and picking; the fitting/mesh caches also account for map aspect ratio. Mesh size stays fixed. I checked 64 × 128, 512 × 64, 64 × 512 and square terrain grids, including transitions, wrapped picking and changing map dimensions. OpenGL and software-only builds/rendering pass, as do geometry/picking tests under ASan/UBSan. The 256 × 256 Oazis rendering benchmark measured 16.64 ms median with clouds in this run on the M3 (rendering only).

Keeping this in draft and current with master while the revival work settles. Manual mode switching remains the default, automatic folding remains opt-in in the main Settings menu, and native Wayland playtesting is still outstanding.

genixpro:
Visual evidence for the map-proportion fix in 1016f0b8. These are actual OpenGL renders using the benchmark’s synthetic checkerboard terrain, which makes tile proportions easier to compare. Dimensions below are map width × height.

| Tall map — 64 × 128 | Square map — 128 × 128 | Wide map — 512 × 64 |
| --- | --- | --- |
| 64 × 128: fat torus | 128 × 128: square-map torus | 512 × 64: thin ring |

Tall maps produce a fatter tube and wide maps a thinner ring. The visual priority is square-looking tiles in the visible, camera-facing region. The mapping preserves local surface proportions; perspective still compresses tiles as the surface turns away, and tile area varies around the tube. Distortion at the sides and inner opening is less important than readability in the main visible region. Click an image for full resolution.

To reproduce, build `torus-render-benchmark` and run with `GLOB2_BENCH_SIZE=64x128`, `128x128`, or `512x64`, and `GLOB2_BENCH_CAPTURE=/tmp/torus.ppm`. These images illustrate the geometry; the automated rendering, picking and software-path validation is described in the preceding comment.

genixpro:
Maintenance update: merged master through 95de302a in 3cc46b17 and resolved the conflicts with the identifier cleanup, fullscreen letterboxing, and new game-speed settings. Both settings are available without overlapping; manual G switching remains the default and automatic torus folding stays off by default.

The Linux/Windows failure on 1016f0b8 was a numerically unstable finite-difference test on the very small inner wall of a tall-map torus. I reproduced it locally with fused multiply-add disabled and increased the sampling interval without relaxing the accuracy threshold or changing the renderer. Geometry and picking now pass locally with FMA disabled and with ASan/UBSan.

OpenGL and software-only builds/rendering, software-only startup with GPU mode requested, rectangular-map fitting/picking, fullscreen aspect-ratio checks, game-speed controls/playback, and the selection, terrain, wrapped-building, entering-unit and savegame regressions pass locally. Resource-atlas comparisons still have zero differing pixels. The latest 256 × 256 rendering-only benchmark measured 11.4 ms median with clouds on the M3. CI is running for the new merge; the PR remains draft, and native Wayland playtesting is still outstanding.

stephanemagnenat:
Note/question: did you consider rendering the units in 3D? The repository has the 3D models of the globules in Blender.

Also, we could try to see if there exist tools to create 3D models from 2D images, for the other elements.

That could create a small world/little planet style.

Of course, this would make the two rendering paths (2D and torus) very different and thus increase the load time if done naively, but probably some on-demand loading could be done.

Giszmo:
> Note/question: did you consider rendering the units in 3D? The repository has the 3D models of the globules in Blender.

That would be a separate PR but of course, wrapping a 3D world into a torus might have extra challenges compared to the mere 2D wrapping.

> Visual evidence for the map-proportion fix

That checkered torus images are a great way to visualize the problem. I would tilt the torus though to fill the screen's aspect ratio or even allow it to crop left and right to focus more on the front facing landscape.

stephanemagnenat:
> wrapping a 3D world into a torus might have extra challenges compared to the mere 2D wrapping.

Without thinking too much about it, I do not expect to be too hard to transform an (x, y, z) coordinate (for z << x or y) onto the torus coordinates. I imagine we can do that on the shader itself. Because you basically take (x, y), map it on the torus 3D point, then multiply z with the unitary normal at that point, and voilà.

genixpro:
@Giszmo Addressed the tilt suggestion in 0e4524dc. The camera now fits the torus silhouette to the playable area’s aspect ratio instead of forcing a steep view to keep the hole visible. This gives the front-facing landscape more room; wheel zoom still allows a closer view with cropped sides. The local square-tile mapping is unchanged.

Updated actual renders, with map width × height:

| 64 × 128 | 128 × 128 | 512 × 64 |
| --- | --- | --- |
| Tall map, foreground-focused camera | Square map, foreground-focused camera | Wide map, viewport-fitting camera |

Shallow angles also exposed precision loss when picking narrow projected triangles. The barycentric calculation now uses double precision, with coverage across portrait and wide windows, map shapes, and transitions. OpenGL and software-only builds/rendering, software-only startup with GPU requested, ASan/UBSan geometry/picking checks with FMA disabled, the terrain regression and all 158 CppUnit tests pass locally. The 256 × 256 rendering-only benchmark measured about 15.4 ms median with clouds on the M3 in this run.

Master is merged through b47c7d23 in the separate merge commit e1557530. The previous Linux/Windows CI run passed; new checks are running. Keeping the PR draft and automatic folding off by default. Agreed that 3D unit models belong in a separate PR; this one continues to reuse the sprite renderer.

Giszmo:
Can you render how 64x512 would look? I'd tilt the 512x64 much flatter. When you zoom out, you still care about what was the center of screen and in this case it drifts off very far.

I assume the 64x512 case would need an extreme ellipsis rotated around the center instead of a circle.

genixpro:
@Giszmo The screen-center drift and flatter wide-map view are addressed in a5a9491e. The map position at the center of the 2D view stays anchored throughout the single smooth transition. The 512 × 64 view is now much flatter (about 19.5° here), and the surface at the focus faces the camera so its projected tiles stay close to square. The geometry and picking code share that anchor, with regression coverage for the transition and rectangular maps.

Here are actual OpenGL checkerboard renders from that commit, width × height:

| 64 × 512 | 512 × 64 |
| --- | --- |
| 64 × 512 circular torus | 512 × 64 flatter circular torus |

We compared an elliptical cross-section and kept the circular shape after visual testing. The 64 × 512 case still has a very small inner opening and substantial variation in tile area around the tube; readability at the camera-facing focus is the priority.

Master is now merged through f50ab27a in 708a65d5a, with build/CI conflicts resolved. Local OpenGL and software rendering checks, geometry/picking sanitizer checks, all 170 unit tests, the standalone tests, and the terrain, gradient and team-statistics regressions pass. Fresh Linux/Windows CI will also check the software-only build. Automatic folding remains off by default in the main Settings menu, with G switching available. The PR is ready for review again; native Wayland playtesting is still outstanding.

genixpro:
The remaining resource-rendering discrepancy is fixed in 803387f92. With a partial HD pack, standalone HD textures could be drawn ahead of earlier original-atlas sprites. Rendering now finishes the preceding batch when switching artwork sources, preserving overlap order while retaining batching for consecutive frames.

The regression compares batched rendering against immediate draws, interleaves original and HD frames, and covers native/scaled coordinates and opaque/translucent rendering. Both artwork modes now have a maximum pixel difference of **0/255** locally. The torus cloud layer also uses the map animation clock so clouds and shadows stay frozen together when paused.

Master is integrated through 265eacf86. All 178 unit tests, OpenGL and software torus integration, the HD integration harness, software resize checks, and the strict translation audit pass locally. Fresh Linux/Windows CI is running. A 50-frame Oazis 256×256 render benchmark on an Apple M3 measured median torus times of 10.3 ms without clouds and 18.0 ms with clouds at a 1120×800 map viewport. Native Wayland playtesting remains unverified.

Giszmo:
Again you got zero ACK to merge this. What are you doing? Please stop merging fun experiments without approval.

PR #140, opened 2026-09-05
Map editor: only remove ressources the painted terrain no longer allows

Giszmo:
Painting sand on a forest in the map editor left a column east and a row south of the new sand as grass without trees.

A tile is drawn from the undermap corners at (x..x+1, y..y+1), so painting the cell (x, y) changes the tiles from (x-1, y-1) to (x, y), while the ressource clearing was a square centred on the cell (3 wide for sand and grass, 5 for water). The overhang to the east and south stayed grass and lost its trees.

Now the terrain is painted first and `Map::removeUnallowedRessources` drops, over the tiles a stroke can change, only the ressources whose terrain no longer allows them, the same test `setRessource` applies when the editor places one. Grass stays the brush that clears everything: it bares the tiles the cell touches and no longer the column and row beyond them.

Headless check on `maps/FourSquares1.map`, one cell painted in a forest interior with the calls from `MapEdit::handleTerrainClick` (`g`/`s`/`?` terrain, `T` tree, `.` nothing, `<` the painted cell):

Sand, before:
```
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
gT gT ?. ?. g. gT gT 
gT gT ?. ?.<g. gT gT 
gT gT g. g. g. gT gT 
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
```
Sand, after:
```
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
gT gT ?. ?. gT gT gT 
gT gT ?. ?.<gT gT gT 
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
```
Grass, after:
```
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
gT gT g. g. gT gT gT 
gT gT g. g.<gT gT gT 
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
gT gT gT gT gT gT gT 
```


genixpro:
Merged as `2a8f918a` after adapting the original changes to the current map/editor modules and adding repeatable regression coverage.

Changes made during verification:
- Resolved the old `Map.cpp` / `MapEdit.cpp` conflicts into `src/map/MapResources.cpp` and `src/map/edit/MapEditClicks.cpp`.
- Added the real-client `TerrainResourcesHarness`, a top-level `terrain-test` target, and execution in Linux CI.
- Integrated #158 after it merged, resolving the build-script overlap and retaining both test targets. Final head `949153e8` differs from the initially verified combined head `f01173f1` only in `test/README.md`; all source/build/CI files are byte-identical.

Verification:
- macOS arm64: release client and server builds pass; both selection and terrain harnesses pass in release and under AddressSanitizer + UndefinedBehaviorSanitizer.
- Ubuntu 26.04 x86_64 (`therig`): optimized combined client build and both harnesses pass.
- Ubuntu 26.04 x86_64 (`pharaoh-dev-2`): terrain harness passes under ASan + UBSan, with no leak suppression.
- Terrain coverage: 240 strokes across all eight resource types, all three base terrains, interior and all four toroidal corners, including overlapping strokes. A whole-map oracle checks that incompatible resources are removed and compatible resources are preserved; grass explicitly clears only its four intended tiles.
- Existing suite on master + #158: all 153 CppUnit tests and all 18 standalone test executables pass on macOS and Ubuntu 26.04. The Ubuntu 22.04/24.04 CI runs also rebuild and run the full suite with the terrain changes included.
- Combined-client smoke: a four-Nicowar game on FourSquares1, seed 20260906, completes normally at tick 39554. The Mac client also launches and renders its main menu.
- Final-head Ubuntu 22.04, Ubuntu 24.04, and Windows/MinGW CI all passed before merging: https://github.com/Globulation2/glob2/actions/runs/34006082072

Limits: terrain coverage exercises the real map operations used by the editor; it does not automate actual editor mouse events or assert visual painting behavior. The separate existing USL teardown leak found while testing #158 is documented in that PR's merge-verification comment; the terrain harness itself passes Linux sanitizers without a suppression.

PR #142, opened 2026-09-05
Make the remember-unit setting remember the assigned counts

Giszmo:

The remember-unit checkbox has been saved and shown since 2007 but nothing read it, and the counts set in a game were never written back, so every game started with the default assignment regardless of the checkbox.

Original commit: 05ae83f7fbcffe13e758672b698c1db14436ff9b

### DIFF
--- a/src/GameGUI.cpp
+++ b/src/GameGUI.cpp
@@ -175,6 +175,8 @@ GameGUI::~GameGUI()
 {
 	for (ParticleSet::iterator it = particles.begin(); it != particles.end(); ++it)
 		delete *it;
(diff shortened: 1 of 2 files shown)


genixpro (dismissed):

genixpro:
Current  already contains this behavior in  and  (the old files were moved during the GUI refactor). This branch is stale/conflicting and should be treated as superseded rather than merged as-is.

kylelutze:
Already on master as 6347a603; the files moved to src/gui/, hence the conflict here.

PR #144, opened 2026-09-05
Connect the end-game chart segments end to end

Giszmo:

Each chart segment ran from (px+1, previous value) to (px, new value), so consecutive segments never shared an endpoint, producing a broken-looking chart.

Original commit: 1e82e81056cacee224af1eb7fb41e23e80034a58


genixpro (dismissed):

genixpro (approved):
Reviewed against current master. Isolated build passed; ASan/UBSan focused checks passed; randomized headless engine smoke test started successfully. The one-line endpoint correction is self-contained and low risk.

PR #146, opened 2026-09-05
Draw the terrain selector highlight after flushing the tile batch

Giszmo:

In GL mode the terrain and resource sheets are uniform-size atlases, so their sprites are queued until finishDrawingSprite. Drawing the highlight before the batch flushed put it behind (or ahead of, depending on batching order) the terrain sprites it should sit above.

Original commit: 4cbe0f63bb8631bca549952494e49a2f873ba123


genixpro (dismissed):

genixpro:
Current  already flushes the terrain batch before drawing the selector highlight in . The old  branch conflicts after the editor refactor. This PR is superseded on current master.

kylelutze:
Already on master — landed with the #132 merge (86d0398c); the draw code now lives in src/map/edit/WidgetsTools.cpp, hence the conflict here.

PR #149, opened 2026-09-05
Remove leaked forbidden zone when a building dies awaiting upgrade room

Giszmo:

Original commit: 7a61c2993e36efa6f98dbf7e912bd7ec4208cc86


genixpro (dismissed):

genixpro (approved):
Reviewed against current master. Isolated build passed; ASan/UBSan focused checks passed; randomized headless engine smoke test started successfully. Cleanup is guarded to the upgrade-room state and is self-contained.

PR #151, opened 2026-09-05
Fix SGSL hilight name table pointing at the wrong item

Giszmo:

Original commit: c6c71dddf4b7db59380344d5fc2c5bef1592e46c


genixpro (changes requested):
Requesting changes: the new SGSL implementation does not compile with the repository’s current macOS toolchain/configuration. In a clean integration tree containing the five merged fixes plus PRs #142–#152, Apple Clang stopped in src/SGSL.cpp because std::optional, std::nullopt, and std::size are unavailable under the project’s existing language mode. The current SConstruct does not request C++17. Please either use the project-supported language level and compatible facilities, or explicitly update the build configuration across platforms and rerun macOS/Linux/Windows CI. The historical Linux CI result is not sufficient for this cross-platform regression.

genixpro:
Current  already contains the corrected SGSL highlight-name table in  and now builds with the repository's C++20 mode. The old  branch is stale/conflicting and should not be merged as-is.

PR #152, opened 2026-09-05
Preserve readiness and verify lobby slot compaction

Giszmo:
Removing a player from the lobby could mark the following player ready without their consent: `removePlayer()` copied the survivor's readiness into the vacated slot, then overwrote it with `true`. Clear the departing slot before compaction so readiness follows each surviving player, including when removing the last or only player.

The original slot-mask fix (`BasePlayer::setNumber()` instead of the wrong-direction shift) is already on master through commit 44947adc. Updating this branch with current master reconciles the moved source file and retains that fix. The remaining diff adds the readiness correction and behavioral regression coverage.

Validation:

- New standalone test exercises all 312 combinations of valid removal position/lobby size, human-only or mixed AI players, and alternating readiness states, plus repeated departures and rejoins.
- Checks slot masks, checksums, AI renaming, player identity/team preservation, empty slots, readiness, and binary player-info round trips using production code; only AI localization is stubbed.
- Regression checks fail before the readiness fix and when restoring the original wrong-direction shift; the fixed implementation passes.
- macOS arm64 client and YOG server builds pass. The CppUnit suite (153 tests) and all 19 standalone test/harness executables pass.

This supplies behavior-focused coverage requested when the earlier compile-only approval was dismissed. Linux and Windows CI will validate the updated branch.


genixpro (dismissed):

kylelutze:
turns out it's already merged in to main but astra found another bug while fixing the merge commit so I let it fix that and I'll merge it.

genixpro:
Independent review found that the expanded lobby test trips UBSan on valid AI player types: `BasePlayer.cpp:30` loads encoded value 8 outside the range of the unfixed `PlayerType` enum. Commit `54fbf7bb08b1ba90c8cbd9f32f9298580c33a174` fixes the enum's underlying type to `Uint32`, preserving existing numeric values and 32-bit serialization, and adds explicit coverage of every AI implementation through compaction and player-info serialization. Serialized-type validation remains unchanged.

Verified so far on macOS arm64 with AddressSanitizer and UndefinedBehaviorSanitizer (`halt_on_error=1`): the lobby test and BasePlayer save/load test pass, including the highest valid AI type and rejection of invalid serialized values. Restoring the old readiness assignment produces 99 regression assertions; restoring the fix returns to zero failures.

The revised head is undergoing the Ubuntu 22.04, Ubuntu 24.04 and Windows CI matrix plus focused optimized/sanitizer runs on the available Linux machines. I will add the final machine results before merging.

genixpro:
Verification completed for head `54fbf7bb08b1ba90c8cbd9f32f9298580c33a174` before merging.

The readiness fix is covered by 312 count/removal/mixed-player/readiness combinations, repeated departures and rejoin, and explicit tests of every AI implementation. Checks include slot masks, checksums, identity/team preservation, AI renaming, cleared slots, individual/aggregate readiness, and binary player-info round trips. Restoring the old readiness assignment produces **99 assertion failures**; restoring the fix passes.

Independent sanitizer testing found and fixed undefined behavior in valid `P_AI+n` values: `PlayerType` now has fixed underlying type `Uint32`. Existing numeric values, 32-bit serialization, and rejection of invalid serialized types are preserved.

| Machine | Platform/compiler | Optimized regression tests | ASan + UBSan regression tests |
| --- | --- | --- | --- |
| Local Mac | arm64 / Apple Clang 21 | Pass | Pass |
| therig | Ubuntu 26.04 x86_64 / GCC 15.2 | Pass | Pass |
| pharaoh-dev-1 | Ubuntu 26.04 x86_64 / GCC 15.2 | Pass | Pass |
| pharaoh-dev-2 | Ubuntu 26.04 x86_64 / GCC 15.2 | Pass | Pass |
| pharaoh-dev-3 | Ubuntu 26.04 x86_64 / GCC 15.2 | Pass | Pass |
| devlaptop | Ubuntu 24.04.4 x86_64 / GCC 13.3 | Pass | Pass |

The focused executables are `NetGamePlayerManagerTest` and `BasePlayerSaveLoadTest`. Linux used `-O2`, then `-O1 -fsanitize=address,undefined -fno-omit-frame-pointer`, `UBSAN_OPTIONS=halt_on_error=1`, and `ASAN_OPTIONS=detect_leaks=1`. **No leak suppressions.** Source SHA-256 values were checked against the reviewed head on all five hosts.

GCC harness limitation: the existing winning-condition stubs omit Game/Team RTTI, so the local test wrapper disables only virtual-pointer instrumentation for `WinningConditions-decode.o`. All other sanitizer checks remain there, and all checks remain enabled for the actual lobby/player/serialization objects. macOS passes without this adjustment. The wrapper is not part of the PR.

Additional macOS verification: client and YOG server builds; **153 CppUnit tests plus all 19 standalone test executables**; and a complete four-Nicowar `FourSquares1` match with seed 20260906. The match finishes at tick 39554 with team 3 winning and 11479 orders, matching the earlier baseline.

Final-head [CI matrix](https://github.com/Globulation2/glob2/actions/runs/34007317837): Ubuntu 22.04, Ubuntu 24.04, and Windows (MinGW-w64) all pass.

These checks verify in-process lobby state and serialization. Actual cross-machine host/join/leave UI and network transport behavior remain a separate integration test.

PR #153, opened 2026-09-05
Fix placeBuildingAt future-worker lookup for flags

Giszmo:

placeBuildingAt computed the finished-building default-assign row as `getDefaultAssignedUnits(typeNum+1)`. The +1 relies on the buildingsTypes table being laid out as consecutive [site, finished] pairs, which holds by accident for normal buildings but breaks for flags: flags have no site variant, so the placeable typeNum is already the finished entry and +1 walks into the next building's row (e.g. an exploration flag reads the war flag's default assign). Looks the finished type up by name instead, matching the other call sites in this file.

Original commit: 5993b37a287b06cb6e556fd7e0416902971da411 by kylelutze — adapted here to master's `getTypeNum(name, level, isBuildingSite)` API since the branch's `getFinishedTypeNum(name)` convenience helper doesn't exist yet on master (patch applied but referenced a symbol that doesn't compile on master; verified with a full `scons` build before and after the fix).

### DIFF
--- a/src/GameGUIToolManager.cpp
+++ b/src/GameGUIToolManager.cpp
@@ -376,7 +376,7 @@ void GameGUIToolManager::placeBuildingAt(int mapX, int mapY, int localteam)
 			isRoom = false;
 		
 		int unitWorking = defaultAssign.getDefaultAssignedUnits(typeNum);
-		int unitWorkingFuture = defaultAssign.getDefaultAssignedUnits(typeNum+1);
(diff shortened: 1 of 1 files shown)


genixpro (approved):
The finished-type lookup correctly handles flags without relying on adjacent site/finished entries. A focused C++ harness executing the production lookup and assignment expressions against data/buildings.txt reproduced incorrect defaults for all three flags on base and passed for all 13 level-zero placeable types on this head, on macOS arm64 and Linux x86_64.

PR #154, opened 2026-09-05
drawChoice: fix tutorial-hilight arrow Y using horizontal centering offset

Giszmo:

The arrow position queued for a hilighted right-panel building/flag passed decX (the horizontal sprite-centering offset) as the Y nudge, producing a vertical misalignment that grew with the sprite's width.

Original commit: 44865b4b9e3198b62621f9651b673450f8ff8de9 by Kyle


genixpro (approved):
The arrow now uses the same vertical offset as the building sprite. A focused harness running the complete production drawChoice method with a recording graphics stub reproduced the base misalignment and passed on this head for both 2/3-column layouts and four sprite widths, on macOS arm64 and Linux x86_64.

PR #155, opened 2026-09-05
Fix crash when highlighting a tool over a panel that doesn't own it

Giszmo:

drawChoice() asserted that, in TOOL_SELECTION mode, the active tool's building name is always present in the panel's types vector. That invariant is false: selectionMode and the displayed panel are independent axes. Selecting a tool couples it to a panel at pick time, but the player can then tab-cycle the display panel while the tool stays selected, leaving sel unset (-1). Draw the highlight only when the tool actually belongs to the panel being painted.

Original commit: 9bffb5fe37f35dab53c44282c5e75512468bcb83 by kylelutze


genixpro (approved):
The sel >= 0 guard correctly permits a selected tool to be absent from the displayed panel. A focused harness running the complete production drawChoice method with a recording graphics stub reproduced the base assertion failure and passed on this head for mismatched and matching tools in both panel layouts, on macOS arm64 and Linux x86_64.

PR #156, opened 2026-09-05
Fix swapped fruit/resource building-failure labels

Giszmo:

The if/else chain mapped UnitCantAccessFruit to "too far from resource" and UnitTooFarFromResource to "can't access fruit" — each pair displayed the other's label in the building inspector.

Original commit: 4fd4a565d0a22c71cc984ac8b070b9bc6b87997e by Kyle (minimal swap here; the branch also refactors this into an indexed table, left out to keep this PR to just the fix).


genixpro (approved):
Verified the two failure reasons against Building::UnitCantWorkReason and the existing translation keys. The fruit-access and resource-distance messages now correspond to the reasons that generated the counters. No blocking findings.

PR #157, opened 2026-09-05
Fix swapped hilight arrows for warrior/explorer top-bar counters

Giszmo:

drawTopScreenBar's three-branch chain paired the warrior counter with the explorers hilight and the explorer counter with the warriors hilight — a copy-paste swap. A script requesting the explorers stat drew its arrow over the warrior counter and vice versa. Replaced with a table indexed by the same unit-type index used everywhere else in the loop, so the pairing can't be silently swapped again.

Original commit: d8e69de86b067b78c605411a48877ddc8979de4b by kylelutze


genixpro (approved):
Verified the table against UnitConsts.h (WORKER=0, EXPLORER=1, WARRIOR=2), the Hilight IDs, and the counter/sprite loop. Each tutorial highlight now targets its corresponding counter. No blocking findings.

PR #158, opened 2026-09-05
Fix stale building and unit selection when cycling

Giszmo:
Cycling selection after a selected building or unit is destroyed can dereference freed memory before the next draw validates the cached selection. The existing null guard does not catch a non-null dangling pointer.

Call `checkSelection()` at the start of `GameGUI::iterateSelection()`, before branching on selection mode. This clears stale building and unit selections using the live view pointers before either cached payload is accessed.

Add `test/GameGUISelectionHarness.cpp`, a headless integration harness linked with the actual client objects. It uses the real GameGUI, Building, Unit, selection setters, and destruction hooks. A friend fixture accesses the private selection API. The `selection-test` SCons target reuses client objects and runs in Linux CI. Coverage includes deleted building/unit selections before the next draw, null selections, and preservation of a live unit selection with no peer.

Validation on macOS arm64:
- Temporarily removing the validation makes the new harness report an AddressSanitizer heap-use-after-free in the production `GameGUISelection.cpp`.
- Restoring the fix passes with AddressSanitizer and UndefinedBehaviorSanitizer across the linked client code.
- Release client build and release integration harness pass.
- Full existing test build, all 153 CppUnit tests, and every standalone `*Harness`/`*Test` executable pass.

Build/run instructions, including sanitizer flags, are in `test/README.md`. This is a direct method regression, not an interactive replay reproduction.


src/GameGUI.cpp
```diff
@@ -4894,14 +4894,20 @@ void GameGUI::checkSelection(void)
...
+// and the keyboard shortcut fires before checkSelection() runs at draw time),
+// when no peer of the same type exists, or when the selected entity is not
+// owned by the local team. Invoked from the local keyboard handler only --
+// never produces a network order, never reads RNG, never mutates sim state.
 void GameGUI::iterateSelection(void)
 {
 	if (selectionMode==BUILDING_SELECTION)
 	{
 		Building* selBuild=selection.building;
+		if (!selBuild) return;
```
genixpro: [P1] Check the live selection before dereferencing the cached pointer

The new null guard does not handle the stale pointer described by this PR. Team.cpp clears game->selectedBuilding and myBuildings[id], then deletes the Building; it does not clear GameGUI::selection.building. That cached pointer remains non-null, so this guard passes and selBuild->gid reads freed memory. One reachable window is fast-forward replay: Engine.cpp advances simulation on skipped GUI frames, then processes the next keyboard events before drawPanel() calls checkSelection(). Please call checkSelection() before branching on selectionMode, or validate the live game.selectedBuilding pointer before accessing the cache. Include a regression where the selected building is deleted while the cached pointer is still non-null.

I compiled the exact iterateSelection() body from this head with minimal entity/team stubs, modeled the Team.cpp deletion sequence, and ran AddressSanitizer on macOS arm64 and Ubuntu 26.04 x86_64. The null and live-selection/wraparound cases pass, but the deleted-selection case reports heap-use-after-free at src/GameGUI.cpp:4910 on both. This is a focused method reproduction, not an interactive replay reproduction.

genixpro (changes requested):
The reported stale-selection crash remains reproducible under AddressSanitizer. The added null guard handles null selections but not a destroyed building whose cached pointer remains non-null. See the inline finding for the lifetime path, reproduction, and suggested correction.

genixpro:
Current  already contains the stale-selection guards in , along with the associated selection harness. The old branch conflicts against the refactored GUI layout, so it should not be merged as-is.

genixpro (approved):

genixpro:
Merged as `a666370e` after verification of the corrected selection lifetime handling. Tested code at `b5a2d910`; final head `8a494b20` only updates the test documentation.

Checks performed:
- macOS arm64: built the release client; the real-client `GameGUISelectionHarness` passes in release and with AddressSanitizer + UndefinedBehaviorSanitizer (`halt_on_error=1`).
- Negative control on both macOS arm64 and Ubuntu 26.04 x86_64 (`pharaoh-dev-1`): temporarily removing `checkSelection()` reproduces an ASan heap-use-after-free. Restoring the fix passes the selection regression.
- macOS and Ubuntu 26.04 (`pharaoh-dev-2`): all 153 CppUnit tests and all 18 standalone test executables pass.
- Ubuntu 26.04 (`therig`): optimized client build and selection harness pass.
- Final-head CI: Ubuntu 22.04, Ubuntu 24.04, and Windows/MinGW all pass: https://github.com/Globulation2/glob2/actions/runs/34005516181
- The combined build with #140 also completed a deterministic four-Nicowar headless match on FourSquares1 (seed 20260906, 39554 ticks, normal game end).

Linux sanitizer caveat: unsuppressed LeakSanitizer reports four existing `MapScriptUSL::addGlob2Values` allocations totaling 128 bytes during USL teardown. The allocation/teardown code is unchanged from master. A suppression limited to that allocation function accounts for exactly those four objects; with it, the selection harness passes with no other sanitizer findings. No suppression is included in this PR. This is a pre-existing shutdown leak, not a clean leak-free result.

This verifies the real production selection methods and destruction hooks headlessly; it is not an interactive fast-forward replay reproduction. The previous changes-requested finding is addressed by validating the live selection before either cached payload is dereferenced.

PR #161, opened 2026-09-06
Fix LAN joining and shared map downloads

### LINKED ISSUE #90: LAN game connection seems broken
In my home, the client connects to the host and is seen by the host, but on the client side, an empty screen is show forever.

I'm using local network (addresses 192.168.178.XX), the problem happens with both the two computer on Ethernet, and with one Ethernet one Wi-Fi. I am using a Fritz router (FRITZ!Box 5490, FRITZ!OS: [07.29](http://fritz.box/)).

Connecting through YOG works.

genixpro:
Joining a LAN game currently opens an empty tab container, so the guest never sees the multiplayer lobby or pumps its game updates. Restore the multiplayer screen for the lifetime of the join session.

Real client testing also exposed two map-transfer failures after joining: the server loops forever over a recipient waiting for the file header, and a second download request restarts the same upload and appends duplicate chunks. Advance the pending recipient and mark the shared upload as started when requesting it.

Add an integration harness that uses real SDL lobby widgets, anonymous LAN server, router, and TCP connections. It enters through `LANFindScreen`, clicks Ready and Leave Game, rejoins, and forces a private map download each time. Both downloads must match the source byte for byte; the host checks readiness, roster size, unique player IDs, slot masks, and departures. A bounded Python runner runs it in Ubuntu CI using SDL's dummy video/audio drivers.

The rendered test also exposed signed overflow in the background terrain generator. Use unsigned 32-bit wraparound without changing the generated tile pattern.

Validation:
- Real network sessions passed on `pharaoh-dev-1` (host) → `pharaoh-dev-2` (guest), `pharaoh-dev-3` → `therig`, and `pharaoh-dev-1` → macOS (visible SDL window). Every pair completed two ready/leave/rejoin cycles and two exact 616,018-byte map downloads; both processes exited zero.
- macOS optimized and AddressSanitizer/UndefinedBehaviorSanitizer loopback sessions passed; sanitizer run used `UBSAN_OPTIONS=halt_on_error=1`, with no suppressions. Linux optimized loopback passed too.
- macOS release client and YOG server builds passed. All four physical Linux hosts ran the same SHA-256-verified Ubuntu 26.04/GCC 15.2 binary built from the final changed source.
- Negative controls: the original empty-lobby path fails the UI test; restoring the old server behavior fails the download regression.

These tests cover the LAN lobby, readiness, departures, and map transfer. They do not start a match or validate gameplay synchronization. Final CI results are recorded in the verification comment before merging.

Fixes #90.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,11 @@ jobs:
           scons -j$(nproc) release=1 server=0 terrain-test
           ./build/src/TerrainResourcesHarness
 
+      - name: Build and run the LAN session regression
+        run: |
+          scons -j$(nproc) release=1 server=0 lan-test
+          python3 test/run_lan_session_test.py build/src/LANSessionHarness
+
       - name: Build and run the tests
         working-directory: test
         run: |
--- a/src/Glob2Screen.h
+++ b/src/Glob2Screen.h
@@ -18,7 +18,7 @@ public:
 	
 private:
 	unsigned getNextTerrain(void);
-	int randomSeed;
+	Uint32 randomSeed; // Background LCG intentionally wraps modulo 2^32.
 };
 
 class Glob2TabScreen : public TabScreen
@@ -30,7 +30,7 @@ public:
 	
 private:
 	unsigned getNextTerrain(void);
-	int randomSeed;
+	Uint32 randomSeed; // Background LCG intentionally wraps modulo 2^32.
 };
 
 
--- a/src/LANFindScreen.cpp
+++ b/src/LANFindScreen.cpp
@@ -125,6 +125,7 @@ void LANFindScreen::onAction(Widget *source, Action action, int par1, int par2)
 			game->joinGame((*client->getGameListManager()->getGameList().begin()).getGameID());
 
 			Glob2TabScreen screen(true);
+			MultiplayerGameScreen lobby(&screen, game, client);
 			
 			listener.disableListening();
 			int rc = screen.execute(globalContainer->gfx, 40);
--- a/src/SConscript
+++ b/src/SConscript
@@ -484,6 +484,12 @@ if not env['server']:
     terrain_test = local.Program('TerrainResourcesHarness', terrain_sources)
     local.Alias('terrain-test', terrain_test)
 
+if not env['server']:
+    lan_sources = [source for source in source_files if source != 'Glob2.cpp']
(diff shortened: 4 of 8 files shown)


genixpro:
Verified commit `8d71820c4ab903b4e488da81bdfe4291540baeb3` before merging.

Physical-machine tests used the real SDL lobby widgets, anonymous LAN server/router, and TCP connections:

| Host | Guest | Result |
| --- | --- | --- |
| pharaoh-dev-1 | pharaoh-dev-2 | PASS |
| pharaoh-dev-3 | therig | PASS |
| pharaoh-dev-1 | macOS | PASS |

Every pair completed two Ready → Leave Game → join cycles. Both joins forced a map download, and both downloaded files matched the entire 616,018-byte source. Host assertions checked readiness, roster size, unique player IDs, slot numbers/masks, and departures. All host and guest processes exited zero. Linux used SDL dummy video; the Mac network guest used a visible window. The four Ubuntu 26.04 Linux hosts ran the same GCC 15.2 binary, SHA-256 checked on each: `73023f3cfc2d23edac37e1cb6b1d01c465f9b529b73f678c89a0ecf0c46ca8ba`.

Additional verification:
- macOS optimized and Linux optimized loopback sessions passed.
- Final macOS ASan + UBSan rendered loopback passed both cycles with `UBSAN_OPTIONS=halt_on_error=1` and no suppressions. Homebrew SDL required `DYLD_LIBRARY_PATH=/opt/homebrew/lib`. This does not provide Linux LeakSanitizer coverage.
- Final macOS release client and YOG server builds passed.
- Negative controls failed as expected: the original empty-lobby path cannot complete readiness/leave, and restoring the old server behavior fails the map-download comparison.
- Reviewed the final production diff and inspected a rendered ready-lobby capture.

Final-head [CI run](https://github.com/Globulation2/glob2/actions/runs/34009170205) passed all three jobs: Ubuntu 22.04 (9m25s), Ubuntu 24.04 (9m44s), and Windows MinGW (13m38s). Both Ubuntu jobs passed client/server builds, selection and terrain regressions, the new LAN regression, 153 CppUnit tests, and 19 standalone harnesses/tests. Windows client and YOG server builds both passed.

The regression is now part of Ubuntu CI (`scons release=1 server=0 lan-test`, then `python3 test/run_lan_session_test.py build/src/LANSessionHarness`). These checks cover the lobby, readiness, map transfers, leaving, and rejoining; they do not start a match or establish gameplay synchronization coverage.

PR #162, opened 2026-09-06
Letterbox the fullscreen scale instead of stretching it

Giszmo:
When the fullscreen logical resolution's aspect ratio doesn't match the desktop/window's, both the GL viewport and the software `SDL_BlitScaled` path stretch the logical rect to fill the window non-uniformly — X and Y get scaled by different factors, squashing circles (button end-caps, cursors) into ellipses.

Scale both axes by the same factor (the smaller of the two — `drawableScale()` already computed this for line width and cursor scaling, just never applied it to the viewport itself) and center the result, letterboxing/pillarboxing with black bars on the long axis instead of distorting. `windowToLogical()` (mouse coordinate mapping) and `setClipRect()`'s `glScissor` scaling get the matching offset so clicks and clipping still line up with what's drawn.

Fixes the case reported live: 5120×2880 (16:9) desktop with 1920×1440 (4:3) selected in Settings — button end-caps rendered as flattened ellipses instead of semicircles.

**Verified live** under Xvfb + openbox at a 1280×800 (16:10) virtual desktop with logical resolution forced to 1024×768 (4:3) via `-f -s 1024x768`, both renderer paths:
- GL (`-g`): before the fix, buttons stretched to fill the whole 1280×800 window with visibly squashed end-caps. After: centered pillarboxed content with correct circular end-caps, symmetric black bars.
- Software (`-G`, no GPU): same before/after behavior via the `SDL_BlitScaled` path.
- Clicked a button precisely where it renders in the pillarboxed window (window pixel, not logical pixel) and confirmed it navigated to the corresponding screen — confirms `windowToLogical()`'s offset matches the viewport/blit offset, not just that the pixels look right.

Couldn't run the `test/` suite in this environment (missing `cppunit` headers, pre-existing and unrelated to this change).


genixpro:
Final verification targets `12635a7a2b08db890c4d23bc1f286f5352994a9d`, including the merged spelling cleanup from #163.

The follow-up fixes capture only the OpenGL game viewport, preserving screenshots and captured editor/end-game backgrounds when bars are present. They also bound the capture resampler's top source row when downscaling. The original PR fails the GL screenshot edge oracle; the software capture was correct.

`FullscreenAspectHarness` exercises actual libgag/SDL presentation, clipped pixel bounds, logical screen captures, translated mouse motion/down/up events, and polled mouse coordinates. It covers equal, pillarboxed, letterboxed, odd, and downscaled dimensions: 640×480, 1280×800, 800×1280, 853×641, and 480×270 for a 640×480 logical surface.

Verification:
- Final combined tree on macOS: all five cases pass in both OpenGL and software, optimized and ASan+UBSan. Sanitizers use `UBSAN_OPTIONS=halt_on_error=1`, no suppressions; Homebrew SDL needs `DYLD_LIBRARY_PATH=/opt/homebrew/lib`.
- The combined macOS client build passes. A 5,000-tick G2.game run with #163 produces a byte-identical replay to baseline: 758,248 bytes, SHA-256 `81c7d800f74406041b71ee80dc07b42aa05651c466d62f66e9469c4e6015fbe1`.
- Physical Ubuntu 26.04 / pharaoh-dev-1, GCC 15.2, Xvfb/Mesa: all ten cases passed with the rendering/capture fixes, before integrating #163's identifier renames.
- Ubuntu 22.04 container, GCC 11 / SDL 2.0.20 / Mesa: all ten cases passed after fixing the test's redundant no-op resize. Negative control: removing that guard reproduces the first software-case failure; restoring it passes all five cases. The redundant resize invalidates the original window surface on the older SDL backend.

Both Ubuntu CI jobs now run the new regression, with a 60-second limit per renderer. These tests resize a real test window through the fullscreen scaling path; they do not claim a full gameplay or display-settings UI walkthrough. macOS sanitizer runs do not provide LeakSanitizer coverage.

Final [CI run](https://github.com/Globulation2/glob2/actions/runs/34033080382) passed all jobs: Ubuntu 22.04 (8m39s), Ubuntu 24.04 (7m52s), Windows MinGW (14m6s). Both Ubuntu jobs passed client/server builds, all ten graphics cases, the LAN/selection/terrain regressions, 153 CppUnit tests, and 19 standalone harnesses/tests. Windows client and YOG server builds passed. The PR is ready to merge at the verified head above.

PR #164, opened 2026-09-06
Respect top clipping boundary when filling software rectangles

genixpro:
Software filled rectangles can paint above the active clip area. For example, drawing from y=0 with a clip beginning at y=30 writes into the first 30 rows. Compare the rectangle top against `clipRect.y`, matching the horizontal clipping logic.

Extend the existing graphics harness with a whole-surface pixel oracle for 24 opaque/translucent rectangle intersections, including partially and entirely above the clip, all four boundaries, negative coordinates, and empty rectangles. Both Ubuntu CI jobs already run this harness.

Validation: the regression fails on master; the fix passes on macOS in optimized and ASan+UBSan builds, with no suppressions. Existing OpenGL/software presentation, capture, clipping, and mouse-mapping cases also pass. Final CI results will be recorded before merging.

### DIFF
--- a/libgag/src/DrawableSurfaceDraw.cpp
+++ b/libgag/src/DrawableSurfaceDraw.cpp
@@ -78,7 +78,7 @@ namespace GAGCore
 			w -= clipRect.x - x;
 			x = clipRect.x;
 		}
-		if (y < 0)
+		if (y < clipRect.y)
 		{
 			h -= clipRect.y - y;
 			y = clipRect.y;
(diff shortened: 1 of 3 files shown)


genixpro:
Independent verification for head `8585c97ee32fd347c1b6ac087ea89726745c72cd`:

- Reproduced the failure on the unchanged implementation: a filled rectangle starting above a positive clip top painted outside the clip intersection.
- The regression checks every pixel for 12 rectangle/clip arrangements at both opaque and half alpha (24 cases), including partial/complete exclusion, each edge, negative coordinates, and empty dimensions.
- On this Apple Silicon Mac, optimized and AddressSanitizer + UndefinedBehaviorSanitizer builds pass all 24 cases and the existing software/OpenGL presentation, screenshot, clipping, and mouse-coordinate checks at five window sizes per renderer. No sanitizer suppressions were used.
- GitHub Actions run [34034533754](https://github.com/Globulation2/glob2/actions/runs/34034533754) passed on Ubuntu 22.04, Ubuntu 24.04, and Windows MinGW. The Linux jobs execute the clipping/aspect regression and existing selection, terrain, LAN, and unit tests; Windows verifies client and server compilation.

The production change is the single top-boundary comparison in `drawFilledRect`. Merging this tested head.

PR #169, opened 2026-09-06
Preserve live team statistics across saved-game loading

genixpro:
Loading a saved game resets live team statistics and their 32-tick sampling cycle. Numbi can consequently see a different hunger count and request an extra inn even when the world still matches. Preserve the live/smoothing histories and their ring positions, and skip the load-time statistics step when those fields are present.

The save format advances from 88 to 89. Older saves retain their existing loading behavior. Named fixed-width fields avoid object-layout dependencies; invalid indices and truncated fields raise load errors. Delegating Team construction releases partially loaded units/buildings on those errors. Persisting the full history adds roughly 197 KB per team to saves and maps; the graph history is retained along with the live snapshot.

### Review follow-up

Giszmo approved the production fix at `1234d957a`. The statistics serialization, format version and live-state restoration are unchanged in `3fcde5949`. Current master is incorporated, including its exchange-building load fix. The dedicated test runner is removed, fixture documentation is reduced to provenance and usage, and the shared savegame-safety runner performs preference checks and exact stdout comparisons for both legacy traces.

### Verification

The `team-stats-save-test` target checks all 32 sampling phases, history wrap, repeated binary reloads, named text fields and malformed-data controls. Linux 22.04/24.04 and Windows CI run it with a disposable HOME/profile and working directory, and compare version-84/version-88 fixture output against the original loader's checked-in traces. Commands are in `test/README.md`.

Final head `3fcde5949` passes optimized macOS and Ubuntu 24.04 tests, both exact legacy trace comparisons, preference-preservation checks, and the shared savegame-safety suite. Targeted ASan/UBSan instrumentation of TeamStat, Team, TeamSerialization and the harness passes the main suite and both legacy loads on Linux; leak detection is disabled for the separately observed script-binding leak, so this is not a full-engine sanitizer or leak-clean claim. A deliberately mismatched expected trace is rejected.

All platform CI jobs pass on the final head: [CI run 34239957982](https://github.com/Globulation2/glob2/actions/runs/34239957982). This is a standalone engine fix; it does not claim full-game save/replay determinism or include tournament tooling.


genixpro:
@Giszmo, please review this and, if you’re satisfied, approve and merge it.

The standalone fix uses upstream save format **89** (incremented from 88), preserves the legacy loading path, and includes genuine version-84/version-88 compatibility traces. The fixture and before/after reproduction recipe are in [test/fixtures/team-stats](https://github.com/Globulation2/glob2/tree/1234d957af3d446697f605cd3e24c57241331315/test/fixtures/team-stats).

Verified head: `1234d957af3d446697f605cd3e24c57241331315`. [CI run 34046272059](https://github.com/Globulation2/glob2/actions/runs/34046272059) passes Ubuntu 22.04, Ubuntu 24.04 and Windows MinGW; all three execute the new regression. The optimized macOS and ASan+UBSan suites also pass, including all 32 sampling positions, repeated reloads, corruption controls and exact legacy traces. The existing autosave safety suite passes locally. Tests use live headless GameGUI owners and disposable profiles, and verify that preferences are unchanged.

This addresses the earlier feedback about runnable reproduction material, portability, save compatibility, test lifetimes and preference safety. It does not include the tournament branch or claim that every other source of save/load divergence is fixed.

test/run-team-stats-save-tests.py
kylelutze, line 1: why do we need an extra test runner?

Giszmo (approved):
Reviewed and verified locally at head 1234d957: built `team-stats-save-test` and ran `test/run-team-stats-save-tests.py`. All regressions pass (32 sampling phases, ring wrap, repeated loads, text streams, corruption controls) and both legacy traces (`version88.game`, `gd-small-2ai.game`) match the expected files.

The fix is right. The old loader's `stats.step(this, true)` left `stats[statsIndex]` zeroed until the smoothing ring wrapped 31 ticks later, so the AI saw a different snapshot after a reload. Restoring the full ring plus both indices and skipping the load-time step gives an identical continuation. The delegating `Team` constructor is a nice touch: the object is fully constructed before `load()`, so `~Team` releases partially loaded units/buildings on a throw, `teams[i]` stays NULL, and `Game`'s destructor already tolerates that.

Non-blocking notes:

1. **Save/map size.** `TeamStat` is 393 ints (1572 bytes); the 128-entry ring adds about 197 KB per team to every save. Since `Team::save` is also used for maps, every `.map` re-saved in the editor grows by the same amount (repo maps average ~630 KB today). Measured: the one-team 32x32 fixture goes from 47,809 bytes (v88) to 250,185 bytes (v89); a 16-team autosave grows by ~3.1 MB. Only `stats[statsIndex]` and the 32-entry smoothing ring matter for AI continuity; the other 127 entries only feed `drawStat`. If size ever becomes a concern, a follow-up could persist just the latest sample plus the smoothing ring (~1.7 KB per team) and accept a blank graph history after load, which is what players got before this PR anyway. Not asking for a change here.
2. **Uncaught path.** `Engine::initGame` catches `std::exception`, so the GUI load path handles the new throw. The headless `dump-resources` / `dump-wheat` entry points in `Glob2.cpp` have no catch, so a corrupt v89 save terminates there instead of printing "failed to load". Previously that was an assert/UB, so still an improvement.
3. **Indentation.** The new blocks in `TeamStat.cpp` and `Team.cpp` use 4 spaces in tab-indented files.
4. **Re Kyle's question on the extra runner:** it does two things the savegame-safety runner does not: it diffs the legacy statistics traces against the checked-in expected files, and it verifies `preferences.txt` bytes and mtime are untouched. That could have been an option on `run-savegame-safety-tests.py`, but a separate runner is fine.

Fable 5.1 helped authoring this review.

test/run-team-stats-save-tests.py
stephanemagnenat: I agree with that comment and similar ones in other PR, let's try not to overload the source tree with harness bloat and temporary docs.

genixpro: Removed the dedicated runner in 3fcde5949. The shared savegame-safety runner now handles disposable profiles, preference checks and optional exact stdout comparison. CI explicitly runs both legacy traces, so the compatibility coverage is retained.

genixpro: The dedicated runner is removed in 3fcde5949, and the fixture README is reduced from 77 lines to 17 lines of provenance and usage. Commands live in the existing test README; exact legacy traces and fixture provenance remain as permanent compatibility coverage.

genixpro:
Final verification for `3fcde594902fe2938cc4e3d0df0cb050dd970f95`:

- Giszmo's approved statistics serialization/restoration and version-89 format are unchanged. The branch incorporates current master; the dedicated runner and excess fixture documentation are removed in response to Kyle and Stéphane's comments.
- Optimized macOS / Apple Clang and Ubuntu 24.04 / GCC 13: all 32 sampling phases, history-ring wrap, repeated binary loads, text streams, invalid indices and truncated-field controls pass.
- Version-84 and version-88 saves produce exact matches for the original loader's checked-in traces on both machines. A deliberately mismatched trace is rejected by the shared runner.
- Disposable HOME/profile/working-directory runs preserve preference bytes and mtime. The existing savegame-safety suite passes through the shared runner on both machines.
- Targeted ASan/UBSan instrumentation of TeamStat.cpp, Team.cpp, TeamSerialization.cpp and the harness passes the main suite and both legacy loads on Linux, with UBSan halt-on-error enabled. Other engine objects are optimized; leak detection is disabled for the separately observed script-binding leak. This is not a full-engine sanitizer or leak-clean claim.

[Final-head CI](https://github.com/Globulation2/glob2/actions/runs/34239957982) passes on Ubuntu 22.04, Ubuntu 24.04 and Windows, including the statistics regression, both legacy comparisons and the existing savegame-safety checks. Merging this tested head with Giszmo's approval of the unchanged production fix.

PR #170, opened 2026-09-06
Pull in two safe Debian packaging patches

Giszmo:
Debian carries these in debian/patches for glob2 0.9.4.4-11 and they were never forwarded upstream (see https://tracker.debian.org/pkg/glob2 -> "debian patches"):

- `sconstruct-explicitly-add-source_date_ep.patch`: SConstruct passes `SOURCE_DATE_EPOCH` through to the build environment when set, for reproducible builds. Guarded on presence so normal (non-Debian) builds are unaffected.
- `desktop.patch`: `data/glob2.desktop` drops the deprecated `Encoding=` key and adds `Keywords=` per the freedesktop.org desktop entry spec.

The other 9 patches Debian carries are either already fixed upstream, target code that no longer exists, or need a maintainer call (filed separately if wanted).

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -249,6 +249,9 @@ def main():
     # tools like ar fall back to /tmp, which sandboxed environments may block.
     if 'TMPDIR' in os.environ:
         env['ENV']['TMPDIR'] = os.environ['TMPDIR']
+    # Likewise for SOURCE_DATE_EPOCH, needed by build tools for reproducible builds.
+    if 'SOURCE_DATE_EPOCH' in os.environ:
+        env['ENV']['SOURCE_DATE_EPOCH'] = os.environ['SOURCE_DATE_EPOCH']
     env["VERSION"] = "0.9.5.0"
     establish_options(env)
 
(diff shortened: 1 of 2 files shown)


genixpro:
Independent verification of head `6aff26a9b1d1011b9e096eafb56aa39ee1246468`:

- Reviewed both changes: the environment variable is forwarded only when present, and the desktop-entry edit is limited to removing Encoding and adding Keywords.
- On devlaptop.local and pharaoh-dev-1.local, extracted the actual SOURCE_DATE_EPOCH conditional from this SConstruct into an isolated SCons compiler probe. Verified unset stays unset; epoch 0 compiles __DATE__/__TIME__ as `Jan  1 1970 00:00:00`; epoch 946684800 produces `Jan  1 2000 00:00:00`.
- `desktop-file-validate data/glob2.desktop` passed on both Linux machines.
- GitHub build run 34047077030 passed Ubuntu 22.04, Ubuntu 24.04, and Windows jobs. These provide the full project builds; the additional local checks above target the packaging changes and do not claim whole-binary reproducibility.
- No unresolved comments or review objections were present when checked.

This is a focused packaging/build fix, with no engine, architecture, or gameplay changes.

PR #171, opened 2026-09-06
Map editor: only remove units and buildings the painted terrain no longer carries

Giszmo:
Follow-up to #140, which fixed only the resource half of this.

**Bug** (reported by Leo in the map editor, present on master since the 2006 editor rewrite): with a 1x1 grass brush, painting two tiles west of a swarm removes it, painting two tiles east of it does not.

**Mechanism.** A terrain stroke paints the undermap corner between tiles x-1 and x (the click is snapped to the nearest corner, and the cursor square straddles those two tiles). Grass and water also fringe the neighbouring corners, so a cell changes at most the tiles (x-2..x+1). The stroke then called `removeUnitAndBuildingAndFlags(x, y, 3)` for grass and `5` for water: a square *centred on the corner index*, which reaches one tile further east and south than west and north. Sand used `2`, the right window, but still removed walkers that can stand on sand.

**Fix.**
- `Game::removeUnallowedUnitsAndBuildings(x, y, w, h)` mirrors `Map::removeUnallowedResources`: over the tiles a stroke can change, drop buildings whose tile is no longer grass and ground units on open water that cannot swim. Air units stay.
- Grass keeps its eraser role from #140: it also clears the four tiles the cursor square covers, and nothing beyond them.
- `removeUnitAndBuildingAndFlags` had `DEL_GROUND_UNIT` reading the air layer and `DEL_AIR_UNIT` the ground layer; every caller passed both, so nothing noticed. Fixed. Note for future callers: passing a lone enumerator selects the `(x, y, size, flags)` overload with the default clear-everything flags, so pass the size explicitly.

**Verification.** `scons release=1 server=0 terrain-test && ./build/src/TerrainResourcesHarness` now also runs an entity oracle: a 4x4 swarm, a walker and an explorer, 300 single-cell strokes of each terrain around them, a west/east mirror check across the swarm, and a four-cell water stroke that drowns the walker and keeps the explorer. The oracle knows nothing about the stroke position, so it fails on any directional bias (it caught the overload trap above during development). `test/` suite: OK (153 tests).

### DIFF
--- a/src/Game.h
+++ b/src/Game.h
@@ -209,6 +209,8 @@ public:
 	//! This remove anything at case(x, y), and return a rect which include every removed things.
 	bool removeUnitAndBuildingAndFlags(int x, int y, unsigned flags=DEL_UNIT|DEL_BUILDING|DEL_FLAG);
 	bool removeUnitAndBuildingAndFlags(int x, int y, int size, unsigned flags=DEL_UNIT|DEL_BUILDING|DEL_FLAG);
+	//! Over the w*h tiles from (x, y), remove the buildings off grass and the ground units on water that cannot swim.
+	void removeUnallowedUnitsAndBuildings(int x, int y, int w, int h);
 	///A convenience function, returns a pointer to the unit with the guid, or NULL otherwise
 	Unit* getUnit(int guid);
 
--- a/src/Game_editor.cpp
+++ b/src/Game_editor.cpp
@@ -248,7 +248,7 @@ Building *Game::addBuilding(int x, int y, int typeNum, int teamNumber, Sint32 un
 bool Game::removeUnitAndBuildingAndFlags(int x, int y, unsigned flags)
 {
 	bool found=false;
-	if (flags & DEL_GROUND_UNIT)
+	if (flags & DEL_AIR_UNIT)
 	{
 		Uint16 gauid=map.getAirUnit(x, y);
 		if (gauid!=NOGUID)
@@ -261,7 +261,7 @@ bool Game::removeUnitAndBuildingAndFlags(int x, int y, unsigned flags)
 			found=true;
 		}
 	}
-	if (flags & DEL_AIR_UNIT)
+	if (flags & DEL_GROUND_UNIT)
 	{
 		Uint16 gguid=map.getGroundUnit(x, y);
 		if (gguid!=NOGUID)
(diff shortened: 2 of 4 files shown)


genixpro:
Independent verification of updated head `d394c1eee1c4546eb8a033e35646e36dac2d7ddc`:

The initial sanitizer run found the new test fixture constructing Game without a live GameGUI owner (`MapScriptUSL.cpp:97`, null member access). The follow-up commit gives the fixture a live headless GameGUI, disables preference saving, returns explicitly from main, and expands entity coverage. The production fix is unchanged.

Verification:
- macOS ASan+UBSan, `UBSAN_OPTIONS=halt_on_error=1`, no suppressions: all resource/entity assertions pass. A disposable HOME seeded with preferences verifies unchanged preference bytes and modification time.
- Entity coverage: 1,500 individual terrain strokes across grass/sand/water at an interior position and four wrapped corners; 15 west/east mirror pairs; 10 flooding scenarios with swimming/non-swimming workers and a co-located explorer; and separate air/ground deletion-flag checks. Checks cover both entity tables and map occupancy. The existing 240 resource-stroke cases also pass.
- Negative control: recompiled the actual Game_editor.cpp with the old swapped air/ground deletion flags and relinked the real harness. The entity assertions reject it; the fixed version passes. The negative control does not depend on a sanitizer error.
- Independent optimized builds and regression runs passed on devlaptop.local and pharaoh-dev-1.local (`scons -j4 release=1 server=0 terrain-test`).
- The pharaoh-dev-1 binary also passed the regression and preference-preservation checks on therig.local, pharaoh-dev-2.local, and pharaoh-dev-3.local. These three were runtime checks, not fresh builds; transferred SHA-256: `bc8e7987d315ebb84637f5eafc9b62ba0d6bde6ec7682ae106bb8dff956eacba`.

These headless tests exercise the engine operations used by terrain painting; no new manual GUI playtest is claimed. The editor retains grass's existing erase role within the four cursor tiles. No match simulation mechanics or engine architecture changes.

Final CI run [34062006297](https://github.com/Globulation2/glob2/actions/runs/34062006297) passed on Ubuntu 22.04, Ubuntu 24.04 and Windows (MinGW-w64). Both Linux jobs execute the terrain regression. Windows verifies client/server builds and the existing autosave regression; it does not execute the terrain harness. Rechecked the exact head and PR discussion: no outstanding reviews or objections. Merging this focused editor bug fix with the regression follow-up.

PR #172, opened 2026-09-06
Fix spurious scons "Two different environments" warnings

Giszmo:
`savegame_env` (a `.Clone()` of `local` with tweaked `LINKFLAGS`/`LIBS`) was being handed the raw `source_files` list for the SavegameSafetyHarness build. Since almost every object in that list (BuildingUtils.o, Bullet.o, Campaign.o, ...) is also built by `local.Program()` elsewhere in this file, scons ends up registering two different `Environment` instances against the same target object -- same compile action, different env object -- and warns `Two different environments were specified for target ...` once per shared file on every build.

Fix: compile the shared sources through `local.Object()` first (the same environment every other target in this file uses) and only apply `savegame_env` at the link step, where the LINKFLAGS/LIBS override actually matters.

Verified locally: clean `scons -j4` and `scons -j4 server=1`, plus every `test/*-test` alias (`savegame-safety-test`, `selection-test`, `terrain-test`, `lan-test`, `speed-tests`, `building-footprint-test`, `entering-unit-save-test`), build and link successfully with zero `Two different environments` warnings.

### DIFF
--- a/src/SConscript
+++ b/src/SConscript
@@ -497,7 +497,12 @@ if not env['server']:
     local.Alias('speed-tests', speed_test)
 
 if not env['server']:
-    savegame_sources = [source for source in source_files if source != 'Glob2.cpp']
+    # Compile through `local` (the same environment every other target here
+    # uses) so these objects aren't rebuilt under a second Environment
+    # instance further down, which just makes scons warn about duplicate
+    # environments on targets that already exist.
+    savegame_sources = local.Object([source for source in source_files if source != 'Glob2.cpp' and source.endswith('.cpp')])
+    savegame_sources += [source for source in source_files if not source.endswith('.cpp')]
     savegame_sources += local.Object('SavegameSafetyHarness.o', '#test/SavegameSafetyHarness.cpp')
     savegame_env = local.Clone()
     # Headless tests use a console entry point so failures reach the CI log.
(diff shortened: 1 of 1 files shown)


genixpro:
Independent verification of head `32f694f079cf74e0b9466ef71c5c8b11a3624442`:

- Reviewed the object/source partition: all shared .cpp files are registered with the existing local compile environment, the two static libraries remain link inputs, and the console-only LINKFLAGS/LIBS adjustments remain confined to the SavegameSafetyHarness link step.
- On devlaptop.local and pharaoh-dev-1.local, built all seven client test aliases with `scons -j4 release=1 server=0 selection-test terrain-test lan-test speed-tests building-footprint-test entering-unit-save-test savegame-safety-test`. Both builds passed with zero `Two different environments` warnings.
- `python3 test/run-savegame-safety-tests.py build/src/SavegameSafetyHarness` passed on both Linux machines, checking the executable affected by the changed link environment.
- macOS SCons dry-run registration of all seven aliases also passed without the duplicate-environment warning. This was a graph/command check, not a new Mac binary build.
- GitHub build run 34048889798 passed on Ubuntu 22.04, Ubuntu 24.04, and Windows, including client/server builds and the Windows console savegame-safety regression.
- No unresolved discussion or review objections were present when checked. No game runtime behavior changes.

PR #173, opened 2026-09-06
Discuss contribution, review, and coding-agent policies

genixpro:
Make AGENTS.md a short repository map with compatibility reminders and explicit review/merge ownership. CLAUDE.md is a committed relative symlink to it; CONTRIBUTING.md points to the same guide.

This remains an unadopted proposal pending human maintainer agreement. The intended ownership is that the author prepares and revises, while an independent reviewer approves and merges. Another agent session for the same author is not independent review. Substantive engine/gameplay/features need another human maintainer’s explicit approval. These choices remain for maintainers to agree; no repository protections or permissions change.

Following review feedback, the primary guide is about 420 words and leaves agents free to choose their tools and workflow. Detailed information preserved from CLAUDE.md lives in docs/development-notes.md for consultation when relevant. Build examples no longer prescribe a job count; the reference notes that memory and concurrent builds matter. The primary guide omits tick timing and other incidental implementation details, points to source, and asks changes to update affected documentation in the same PR.

Short correctness reminders remain: preserve deterministic cross-platform simulation and save/load continuity; bump save versions for format changes; consider replay/network acceptance separately from saved byte layout. Detailed verification and diagnostic techniques are in the linked reference.

Validation: documentation-only changes; git diff --check passes, local Markdown links resolve, and CLAUDE.md remains a relative symlink to AGENTS.md with identical content through either path. Windows symlink materialization was not tested; the guide covers reading AGENTS.md directly in that case.


AGENTS.md
```diff
@@ -0,0 +1,49 @@
+# Proposed instructions for coding agents
```
kylelutze: If this gets merged this isn't proposed instructions.

AGENTS.md
```diff
@@ -0,0 +1,49 @@
...
+## Scope and ownership
+
+- Engine architecture and gameplay changes are welcome for review. Explain the
+  intended behavior, alternatives and tradeoffs, and discuss broad designs early.
+  Do not treat the existing architecture or mechanics as a categorical restriction.
+- Do not relabel a feature, balance change, or compatibility removal as a bug fix.
+  Explain expected behavior and flag uncertainty or disagreement for discussion.
+- Keep changes focused. Preserve unrelated work and use an isolated checkout when
+  the working tree contains someone else's changes.
+- Follow the build guidance in [CLAUDE.md](CLAUDE.md) and the testing instructions
```
kylelutze: claude.md should point at agents.md, otherwise you're going to create a loop here for claude models

kylelutze:
If these are meant for agents only, they will need some major rework. They're doing too much that they don't need to, but I'm still thinking of the best way to handle it. My first instinct is, what issue are you trying to solve with them that isn't working correctly right now? The AGENTS.md file should only have information in it to solve an existing problem. There's a lot in there of general how to do things rules that I haven't seen be issues without that file yet.

genixpro:
@kylelutze The concrete problems I want this to solve are:

1. Give agents a short map of the repository so they spend fewer tokens rediscovering entry points and finish work faster.
2. Standardize verification of difficult invariants, especially keeping the deterministic engine consistent across platforms, save/load, replays and networking.
3. Make shared policy explicit. For example, whether the author or reviewer merges a PR should not be something each agent guesses independently.

I agree we should have one source of instructions. Updated in 53ed311c5: AGENTS.md is canonical and CLAUDE.md is a committed relative symlink to it; CONTRIBUTING.md points to the same guide. I consolidated the existing CLAUDE.md information rather than dropping it, including the build/test pitfalls and debugging guidance.

The files now read as the intended final guide. Proposal/adoption status belongs here on the PR, which remains pending maintainer agreement. The proposed merge ownership is explicit: the author prepares and revises; the independent reviewer approves and merges. Substantive engine/gameplay changes still need another human maintainer's explicit approval. Please review that boundary as part of the proposal.

#211 provides a concrete example of why short repository-specific reminders help: agents may often infer the need for version bumps, but writing down the save-format and replay/network rules reduces omissions. That reminder is now included too.

AGENTS.md
```diff
@@ -0,0 +1,141 @@
...
+
+A `Team` is a colony; a `Player` controls a team, and several players can share one.
+The engine advances in 40 ms ticks. Multiplayer clients must compute the same
+simulation from the same initial state and orders. See
+[doc/sourceCodeUnderstanding.txt](doc/sourceCodeUnderstanding.txt) for more context.
+
+## Build and test entry points
+
+```sh
+scons -j8                        # default client: debug information, no optimization
```
Giszmo: prescribing -j8 is problematic. I had agents crash their VM by running out of memory. They compiled many binaries in parallel at once.

AGENTS.md
```diff
@@ -0,0 +1,141 @@
...
+| Map state and pathfinding | `src/map/`, especially `gradient/` and `pathfind/` |
+| AI implementations | `src/ai/`, behind `AIImplementation` |
+| Rendering, menus and editor | `src/render/`, `src/gui/`, `src/*Screen*`, `src/map/edit/` |
+| Network and multiplayer service | `src/net/`, `src/yog/` |
+| Graphics/UI and scripting libraries | `libgag/`, `libusl/`, `src/sgsl/` |
+| Builds and platform coverage | `SConstruct`, `src/SConscript`, `scons/`, `.github/workflows/build.yml`, `vcpkg.json` |
+| Tests and replay usage | [test/README.md](test/README.md), [docs/headless-replays.md](docs/headless-replays.md) |
+
+A `Team` is a colony; a `Player` controls a team, and several players can share one.
+The engine advances in 40 ms ticks. Multiplayer clients must compute the same
```
Giszmo: Is the speed dial a new thing? Anyway, 40ms/tick might change such as many other things you document in this file. Make sure that PRs changing this to 50ms for example also have to update this document. I'm basically afraid that by caching such information for all agents, they will get blind to changes.

Giszmo:
I'm pretty impressed how well my Fable did without this document. It came up with simulating in parallel many games to measure stuff and I don't know if the first round per session took longer than it would have with this document. So as long as we make sure this document doesn't get stale this might actually be helpful.

PR #175, opened 2026-09-06
Clamp ally-team widget index in the map editor's teams editor

kylelutze:
## Summary
- The map editor's teams editor converted the header's 1-based ally-team number to a 0-based widget index by bare subtraction. A header with 0, or a value above the map's team count, gave an out-of-range index; `MultiTextButton::setIndex` uses `vector::at`, so the dialog threw and crashed on open.
- Promotes the clamp helper already used by the custom-game options screen into a shared header-only `AllyTeamWidgetIndex.h` and uses it at both sites. Well-formed values still map to `value - 1`; only headers that previously threw now display ally team 1. Nothing written back changes.
- Names the teams editor's widget return-code bases (`PLAYER_ACTIVE_BASE`, `COLOR_BASE`, `AI_SELECTOR_BASE`, `ALLY_TEAM_BASE`) instead of hand-typed 100/200/300/400, and documents `generateGameHeader`'s inverse mapping.

## Verification
- `scons -j16` clean.
- `tests/ally_team_widget_index_test.cpp` (standalone, compile line in its header) passes all cases: 0, above-range, and 255 clamp to 0; 1..teamCount map to 0..teamCount-1.
- Workflow 1 replay check: `cpp-refactor.replay` byte-identical (editor UI only, no sim path touched).

## Manual check
Editor → open teams editor. Before: a map whose game header carries an ally-team number outside `[1, numberOfTeams]` crashes on dialog open. After: that row shows ally team 1. Well-formed maps look identical.


Giszmo (approved):
Reviewed at head 4214b844, tested merged onto master bfc8fd75. Approve, no blockers.

**Verified**
- The crash is real: `MultiTextButton::setIndex` uses `texts.at()` and the ally widget only has `numberOfTeams` entries, so an ally number of 0 or above the team count threw on dialog open. The clamp is the same helper already used in `CustomGameOtherOptions` (4f13a604); well-formed values and the `index + 1` write-back in `generateGameHeader` are unchanged.
- The named return-code constants keep exactly the old 100/200/300/400 ranges; `onAction` logic is untouched.
- `scons -j16` on master + PR builds clean with no new warnings in the touched files. `tests/ally_team_widget_index_test.cpp` compiles and passes with the command in its header. The branch is a few commits behind master but merges clean.

**Non-blocking notes**
1. Reachability: the editor's own add/remove team already rebuilds the header from scratch (`MapEdit::regenerateGameHeader`, `src/map/edit/MapEditClicks.cpp:391`), so this only fires for a map or game file whose stored header is already out of range. Fine as hardening; the PR text does not overclaim.
2. Fallback choice: every out-of-range value collapses to widget 0 = ally team 1, so after OK those teams end up allied with team 0. Falling back to the team's own index (the `reset()` no-alliance layout, always `< teamCount` for a valid team) would be a more neutral default. Keeping 0 is defensible for consistency with the `CustomGameOtherOptions` site.
3. Test placement: the repo convention is `test/` (singular) with CppUnit fixtures built by `test/SConstruct` into `TestsRunner`, which CI runs. This file lives in `tests/` and is hand-run only, so CI never executes it. Suggest moving it into `test/` as a fixture (`FilenameStripTest.cpp` is a minimal template).
4. Pre-existing, out of scope: the same malformed file also feeds `color[i]->setSelectedColor(teamNumber)`, which `ColorButton` stores unchecked, so the dialog now opens but OK writes that team number back.

kylelutze:
Thanks for the review. Note 3 addressed in 0b0f5983: the test is now a CppUnit fixture in `test/AllyTeamWidgetIndexTest.cpp`, registered in `TestsRunner` (158 tests pass locally). Note 4 tracked as #188. Keeping the index-0 fallback (note 2) for consistency with `CustomGameOtherOptions`.

Issue #109, opened 2024-04-07
A unit can survive indefinitely in a fully surrounded building


abacabadabacaba:
If a building is completely surrounded (e.g. by trees) while a unit is inside, the unit will wait indefinitely for the exit to clear. In this state, the unit doesn't consume any food. This situation occurs often when playing against AI.

I think that units should still consume food, and possibly starve, while in this state.

stephanemagnenat:
For me there are two questions bundled here:
1. Should a unit doing normal actions in a building (e.g. upgrading) consume food?
2. Should a unit blocked in a building consume food?

For 1. to minimally disturb the game logic, I would tend to answer no, but it is a bit of an arbitrary choice, we could also have an "in-building food consumption" cost. In that case, the thing to take care of would be to update the food of the unit when it enters the inn, otherwise it might starve there before finishing eating. For 2, we could, but the question is the complexity cost of the special case. Doing 2. along with 1. would probably make the logic simpler.

#180 implements 2 without 1, that is fine for me.

Giszmo:
IIRC, for winning, units trapped in buildings already don't count. For me that's good enough. If Alice in team with Bob has a worker trapped in a pool, let Bob free it to get Alice back into the game whenever. Units the player can't see shouldn't starve. Alice can abandon the game if she wants.

If I don't remember correctly, make the team lose if all units are trapped and don't wait for another 40 minutes to have that one unit trapped in a level 3 inn to eat all the wheat cause to starve in a well stocked inn would also be plenty wrong.

stephanemagnenat:
That would also be fine for me. I do not have a strong opinion here. I slightly favor keeping the code as simple as possible (if one can say that for Glob2).

PR #180, opened 2026-09-07
Eliminate colonies with all remaining units trapped, preserving rescue

genixpro:
A colony can remain alive indefinitely when every remaining unit has finished service but cannot leave its building. Eliminate that colony through the existing team loss/win rules, without changing indoor hunger or killing its units.

This replaces the earlier starvation implementation in response to Giszmo’s review and Bradley’s revised direction. A unit counts as trapped only while waiting inside after service, with no exit available under its actual flying/swimming abilities. Active service, normal transitions, an open exit, or any non-trapped unit prevents this new elimination rule.

Recovery remains possible:
- An active ally defers this elimination, conservatively allowing allied rescue without a reachability proof.
- A hatchery with sufficient food (or production already due), a free unit slot, and an exit for a producible unit defers elimination. Production sliders may be changed by the player, so zero ratios do not remove this protection.
- Existing elimination rules for other situations are unchanged. This does not guarantee every match terminates.

Current master is incorporated. `UnitMedical.cpp` and the shared savegame test runner now match master; CI adds only the trapped-colony regression to the existing Linux/Windows jobs. Simulation version and replay minimum advance to 93 because elimination timing changes. Older saves remain loadable; older replays are rejected.

Validation at `9241e6be3`:
- macOS optimized real-engine regression passes: wood/wheat-blocked feeding/training, elimination with living indoor units and unchanged hunger/HP, active feeding/training, open exits, free workers, allied rescue and loss of ally protection, and stocked/empty/blocked hatchery recovery.
- The same real-engine regression passes on pharaoh-dev-2 (GCC 15), with matching Mac/Linux trace digests: feeding `2436477125`, training `3629482821` for both wood and wheat.
- Repeated seeds and save/load continuation produce identical per-tick traces; disposable profile preferences remain unchanged.
- Negative control with master’s `TeamStep.cpp` fails the expected team-elimination assertion.
- Replay counter/version-boundary and savegame-safety harnesses pass; existing winning-condition harness executes successfully.

Reproduction:
```sh
scons -j4 release=1 server=0 trapped-unit-test
python3 test/run-savegame-safety-tests.py --check-preferences build/src/TrappedUnitLifecycleTest
```
Use `.exe` on Windows. [Final-head CI](https://github.com/Globulation2/glob2/actions/runs/34413034722) is green on Ubuntu 22.04, Ubuntu 24.04, and Windows, including the trapped-colony regression and all existing jobs. Giszmo and Kyle are requested to re-review the replacement behavior and recovery boundaries before merging.


genixpro:
Policy and test audit of `788b9d2466237e1b3b80da90e2f9427f1569616c` (source review; I have not independently executed this regression yet):

Holding this PR as draft pending agreement on intended behavior and stronger verification. The contributor policy in #173 is still a proposal; the hold follows the author's stated requirement to preserve basic gameplay and obtain agreement where a fix changes a disputed rule.

- This changes hunger and elimination while a unit is blocked inside a building. Issue #109 reports the behavior and proposes starvation, but does not establish maintainer agreement. Please confirm that this is restoration of intended behavior rather than a gameplay change, including the intended treatment of feeding/training, rescue, and explorers. Passing tests alone cannot settle that question.
- The new regression is not invoked by CI. Current green build checks would not establish that this regression passes on Linux or Windows.
- `starvationContinuesAfterSave()` constructs `Game restored(NULL)`. The map-script load path expects a live GameGUI owner. Use a headless GameGUI for the restored game and verify under ASan+UBSan; a successful unsanitized run does not establish safe fixture lifetime.
- The runner only isolates the executable, then runs it in the repository using the default profile. It needs a disposable HOME/profile and working directory. `rememberUnit=false` prevents preference saving but does not isolate all default-profile access.
- The fixture directly invokes medical/movement helpers and only checks one post-load hunger decrement. Add a normal simulation-step lifecycle, including death after load, and repeated-seed/save-continuation state comparisons. Record affected replay/network compatibility: existing serialized fields may still load, but simulation outcomes differ across fixed and unfixed engines.

No self-merge: this submission needs a separate reviewer/merger. Keep the technical and intended-behavior questions open until addressed.

genixpro:
@kylelutze @Giszmo — please review and explicitly approve the intended behavior of this small gameplay edge-case fix before it is merged.

The problem is a unit that has finished feeding or training but cannot leave because wood or wheat blocks every exit. Its hunger is currently suspended indefinitely, making the trapped unit immortal and potentially preventing an otherwise eliminated team from losing. The purpose is to let this stalled state resolve through the existing starvation and elimination rules.

The practical reason for this fix is loss of agency: if the colony has no free units or other way to produce or rescue workers, clearing the blocked exits requires the very units that are trapped inside. Keeping those units alive forever leaves a colony unable to act but also unable to be eliminated. Resuming starvation after service makes that terminal state resolve, while preserving rescue if an exit is cleared in time.

This does change behavior in that narrow situation, so under the collaboration rules we are proposing it needs approval from another person. The intended boundary is:
- Hunger remains suspended during active service and normal entry/exit.
- Hunger resumes after service has finished while the unit is still trapped inside.
- Clearing an exit can still rescue the unit before death.
- Indoor death must safely remove the building subscription, without disturbing outside units or spawning an outside death animation.

The aim is to remove this route to a game that cannot resolve to a win/loss; this is not a guarantee that every possible match terminates. No engine redesign or broader balance change is intended. Old saves use the same format, but fixed and unfixed engines can produce different outcomes for this case, so replay/network compatibility needs to be stated accurately.

Please confirm whether you approve this scope, and flag any boundary that should behave differently. Technical verification gaps from the audit still need to be addressed; approval of the intent will not replace regression tests, CI, or review of the final commit. This will not be self-merged.

doc/TrappedWorkerRegression.md
kylelutze, line 1: do we need to keep this around for anything?

test/run_trapped_regression.py
kylelutze, line 1: why does this test require its own python runner?

doc/TrappedWorkerRegression.md
stephanemagnenat: I would suggest to remove doc bloat, once the relevant issues are solved. We could have a minimal game design document, or better, a code so clean we need no document ;-)

genixpro: Removed the standalone regression document in 29125a312; retained concise build/run instructions in the existing test README. The updated fixture now tests normal game ticks, elimination and save/load continuation.

test/run_trapped_regression.py
genixpro: Removed the dedicated linker runner in 29125a312. This now has a standard SCons target and uses the existing disposable-profile runner, shared with savegame safety (and the update in #181). Ubuntu and Windows CI execute it. Linux optimized and targeted ASan/UBSan checks pass; substituting the old production code fails at the expected hunger assertion.

genixpro:
Follow-up commits through 85f903698 address the technical audit and the document/runner comments:

- Normal game ticks now cover completed feeding/training behind wood or wheat, hunger/death/removal, team loss and the survivor's win, rescue, and active-service protection.
- Repeated seeded runs and save/load continuation compare per-tick unit state, building subscription and win/loss flags. The fixture explicitly restores the same RNG checkpoint, initializes player-wait state, and requests a resource gradient for the current lazy scheduler.
- The fixture uses a real headless GameGUI owner and the existing shared runner with a disposable HOME/profile/working directory and a preference-preservation check. The standalone document and custom linker runner are removed.
- Optimized macOS and Ubuntu 24.04 runs pass with matching trace digests. Existing savegame-safety checks also pass through the shared runner on Linux. Targeted ASan/UBSan on UnitMedical, Unit, TeamStep and the fixture passes with leak detection disabled. Substituting unfixed UnitMedical from master 88934ecfb fails at the first hunger assertion.
- Current master is incorporated; Ubuntu 22.04, Ubuntu 24.04 and Windows CI execute this regression.

This remains a small gameplay edge-case change needing explicit second-human approval. The unchanged save format does not establish mixed-version replay/network compatibility. The tested scenario is a trapped ground worker; existing explorer-specific feeding/healing behavior is not changed or covered by this worker regression. It will not be self-merged.

Final-head [CI](https://github.com/Globulation2/glob2/actions/runs/34237295951) is green on Ubuntu 22.04, Ubuntu 24.04 and Windows, including the new lifecycle test. Marked ready for review and formally requested Kyle and Giszmo.

Giszmo (changes requested):
The core gameplay fix in `src/unit/UnitMedical.cpp` is correct, well-scoped, and well-tested — traced the bug through `UnitDisplacement.cpp`/`UnitMovement.cpp` on current master and confirmed the `waitingInside` guard correctly distinguishes genuinely-blocked units (exit blocked by resources) from units actively exiting or in active service. `TrappedUnitLifecycleTest.cpp` is a genuine regression test exercising starvation/death/rescue/save-load. That part is approved in principle.

However, this PR branched before #169/#204/#209 landed and independently reimplemented an older version of `test/run-savegame-safety-tests.py` from scratch (missing the `--expect-stdout` flag those merged PRs added). That's a real content conflict on shared CI infrastructure (also touches `build.yml`/`SConscript`), not a mechanical rebase — resolving it naively in either direction risks silently dropping `--expect-stdout` (which the already-merged `TeamStatsSaveHarness` checks depend on) or losing this PR's new `trapped-unit-test` step. The current green CI only validates the stale pre-conflict branch state, not the post-rebase combination.

Please rebase onto current `master`, drop the from-scratch reimplementation of `run-savegame-safety-tests.py` in favor of consuming the already-merged version, re-add just the new `trapped-unit-test` CI step/SConscript block on top, and re-run CI to confirm all steps coexist. Happy to re-review once rebased.

Edit: Concept NACK as I said in #109 :

IIRC, for winning, units trapped in buildings already don't count. For me that's good enough. If Alice in team with Bob has a worker trapped in a pool, let Bob free it to get Alice back into the game whenever. Units the player can't see shouldn't starve. Alice can abandon the game if she wants.

If I don't remember correctly, make the team lose if all units are trapped and don't wait for another 40 minutes to have that one unit trapped in a level 3 inn to eat all the wheat cause to starve in a well stocked inn would also be plenty wrong.

genixpro:
@Giszmo @kylelutze Bradley agrees with revising this toward elimination rather than indoor starvation. Proposed boundary for discussion:

- Leave indoor hunger and unit survival unchanged.
- A colony with remaining units all trapped, and no remaining non-trapped units able to act, is a candidate for elimination. Trapped means service has finished and the unit cannot exit; ordinary feeding/training and normal exit transitions must not count.
- Do not apply this rule if the colony can produce a new unit that can emerge and act.
- Conservatively defer this elimination while an active ally remains, preserving allied rescue without trying to prove pathfinding reachability for every rescue.

This addresses the colony having no remaining agency without starving invisible units in stocked inns. It does not promise that every possible match terminates; tournament time limits/adjudication are separate.

Please confirm these boundaries, especially allied rescue and unit-production recovery. The current starvation implementation will not be merged while the design objection is unresolved. Once agreed, the implementation and regression should exercise the elimination rule, normal indoor activity, recovery, allied rescue, and save/load determinism, using current shared test infrastructure.

genixpro:
@Giszmo @kylelutze The elimination replacement is now implemented at `9241e6be3`; the earlier indoor-starvation changes are completely removed.

The new check only applies when every remaining unit has finished service and cannot exit. It preserves active service, open exits, free units, recovery through a stocked hatchery with an available exit, and rescue by an active ally. Zero production ratios still allow recovery because the player can change them. Defeated/disconnected allies no longer defer elimination. Existing elimination rules for other situations remain unchanged.

Current master is incorporated, and the shared savegame runner matches master. The new regression passes on Bradley’s Mac and pharaoh-dev-2 with identical trace digests, including repeated seeds and save/load continuation. Replacing the new team logic with master’s makes the elimination assertion fail. Replay/version and savegame safety checks pass locally. [All final-head CI jobs are green](https://github.com/Globulation2/glob2/actions/runs/34413034722): Ubuntu 22.04, Ubuntu 24.04, and Windows.

The PR description has the reproduction commands and full scope. Please re-review the replacement and recovery boundaries; the outstanding change request remains open, and this has not been merged.

src/team/TeamStep.cpp
```diff
@@ -10,6 +10,62 @@
...
+namespace
+{
+// Use the same exit rules as movement, without moving or unsubscribing a unit.
+bool hasExit(Building* building, const UnitType* type)
+{
+	int x, y, dx, dy;
+	return type->performance[FLY]
+		? building->findAirExit(&x, &y, &dx, &dy)
+		: building->findGroundExit(&x, &y, &dx, &dy, type->performance[SWIM]);
+}
```
Giszmo: weird comment. more an inner monolog while writing the code than documenting what the function does.

Also if you change it to

```
bool hasExit(Building* building, bool fly, bool canSwim)
{
	int x, y, dx, dy;
	return fly ? building->findAirExit(&x, &y, &dx, &dy)
	           : building->findGroundExit(&x, &y, &dx, &dy, canSwim);
}
```

you can actually use it twice instead of only once, eliminating some LOCs.

PR #181, opened 2026-09-07
Prevent clearing flag gradients from reading past resource switches

genixpro:
A clearing flag has five basic-resource switches, but fruit IDs are 5–7. The building-gradient builder indexed that array with any nonempty resource ID, reading object padding for fruit. Different allocation histories, including loading a save, could therefore change clearing targets and unit paths.

Check `resource.type < BASIC_COUNT` before indexing the switches. Fruit remains an obstacle, consistent with the supported switches. No serialized layout changes; mixed-version replay/network equivalence is not established because the invalid read can affect paths.

### Regression

```sh
scons -j4 release=1 server=0 clearing-gradient-test
python3 test/run-savegame-safety-tests.py --check-preferences build/src/ClearingFlagGradientTest
```

Use `.exe` on Windows. The real-engine fixture varies only padding after the switch array and checks every fruit, enabled/disabled basic-resource switches, empty tiles, and all swimming classes. It uses the existing shared runner with a disposable profile and working directory, and verifies preference preservation. Linux and Windows CI execute it.

### Review follow-up and validation

Removed the standalone regression document and dedicated Python runner as requested. Concise usage remains in the existing test README. Follow-up `9e2f6b5f0` incorporates current master: its local-gradient implementation has been removed, so the test now covers the current 16-bit weighted building gradient and all swimming classes. The remaining production fix is one bounds check.

macOS / Apple Clang and Ubuntu 24.04 / GCC 13: the updated optimized clearing regression and existing savegame-safety harness pass. Targeted ASan/UBSan instrumentation of `MapGradientBuilding.cpp` and the fixture passes against optimized engine objects. The negative control substitutes the unfixed source from master `88934ecfb` and fails on the out-of-bounds fruit index. Leak detection is disabled because of the separately observed script-binding leak; this is not a full-engine sanitizer or leak-clean claim.

[Final-head CI](https://github.com/Globulation2/glob2/actions/runs/34237594064) passes on Ubuntu 22.04, Ubuntu 24.04 and Windows. Kyle's re-review is pending. This author-owned submission requires separate review and merge.


genixpro:
Verification and policy follow-up on `e4dd4d9649c5238b26fa78078bf2c41d0ff2cdbb`:

- The two production bounds checks are unchanged and fit a focused memory-safety/determinism bug fix. Fruit IDs cannot index the five basic-resource switches. No architecture or serialized-format changes.
- Replaced the ad-hoc linker script with `clearing-gradient-test` in SCons, so tests use the configured compiler/platform libraries and sanitizer flags. Added explicit Linux and Windows CI execution; ordinary client CI did not previously run this regression.
- Fixed test ownership with a live headless GameGUI. The runner isolates HOME and its working directory and checks seeded preferences remain byte-for-byte and timestamp-identical. Assertions remain enabled and Windows uses a console entry point.
- On devlaptop.local (Ubuntu 24.04/GCC 13), `scons -j4 release=1 server=0 clearing-gradient-test` and `python3 test/run_clearing_regression.py --build-dir build` pass on the final sources.
- Recompiled both affected gradient sources and the harness with `-fsanitize=address,undefined -fno-omit-frame-pointer -O1 -g`, linking the remaining optimized client objects. With `UBSAN_OPTIONS=halt_on_error=1` and `ASAN_OPTIONS=detect_leaks=0`, the fixed code passes. Replacing only the two conditions with their original forms makes UBSan stop at resource ID 5 indexing `bool[5]` in MapGradientLocal.cpp. This is targeted instrumentation, not a fully instrumented engine.
- The initial leak-enabled run reports four 32-byte MapScriptUSL native-binding allocations (128 bytes total). Retained that report; leak detection was then disabled for the bounds check. No leak-clean claim or unrelated runtime fix is included here.

New CI is pending. Correcting undefined behavior can change clearing paths relative to the unfixed engine; same save format does not imply mixed-version simulation compatibility. @Giszmo @kylelutze, please independently review the bounds fix and this verification before merging. This will not be self-merged.

Giszmo (approved):
Verified against master bfc8fd75: `clearingResources` is `bool[BASIC_COUNT]` followed by `Sint32 minLevelToFlag`, so fruit IDs 5-7 read the padding bytes; both gradient builders now use the same `< BASIC_COUNT` guard that `updateLocalResources` and `Map::isResourceTakeable` already had. CI green on all three jobs. Approving.

doc/ClearingFlagGradientRegression.md
kylelutze, line 1: why keep the doc around?

test/run_clearing_regression.py
kylelutze, line 1: why does every test need a python runner?

kylelutze (dismissed):

doc/ClearingFlagGradientRegression.md
genixpro: Removed the dedicated reproduction document in edbb6ae4bb8a3b44c5a07887cdcc797de3080307. The test command and coverage now live in the existing test/README.md; the before/after evidence stays on this PR.

test/run_clearing_regression.py
genixpro: Agreed. edbb6ae4bb8a3b44c5a07887cdcc797de3080307 deletes the clearing-specific Python runner and reuses run-savegame-safety-tests.py with an optional --check-preferences flag. Clearing and the existing save-safety suite both pass on Ubuntu 24.04/GCC 13. The shared runner uses a disposable HOME and working directory. New Linux/Windows CI is pending; please re-review this follow-up. The production bounds checks are unchanged.

genixpro:
Updated to current master in 9e2f6b5f0 and resolved the merge conflicts. Master removed local gradients and changed building gradients to 16-bit values, so the retained test now exercises the current API and every swimming class. The remaining production change is the same resource-index bounds check.

Optimized clearing and shared save-safety regressions pass on macOS and Ubuntu 24.04. Targeted ASan/UBSan passes on Linux with leak detection disabled; the unfixed master source triggers index 5 out of bounds for bool[5]. [CI on this final head](https://github.com/Globulation2/glob2/actions/runs/34237594064) now passes on Ubuntu 22.04, Ubuntu 24.04 and Windows, including the clearing regression and savegame-safety checks.

@kylelutze, the document and dedicated-runner requests remain addressed; your re-review request is still open. This has not been merged.

Giszmo:
Resolve conflicts and merge

genixpro:
@kylelutze @Giszmo Addressed the new integration conflicts through 56347c7cf. The production fix remains the resource-index bounds guard. It now uses the merged Tile API; the shared runner is exactly master's implementation, including --expect-stdout. Both the clearing and team-stat CI targets are retained. Kyle's earlier document and dedicated-runner requests remain addressed.

Fresh macOS optimized builds/runs pass for ClearingFlagGradientTest and SavegameSafetyHarness with unchanged disposable preferences. Version-84 and version-88 TeamStatsSaveHarness traces exactly match the checked-in expected output through the shared runner. Targeted ASan/UBSan on MapGradientBuilding.cpp and ClearingFlagGradientTest.cpp passes with leak detection disabled; replacing only the bounds guard with the unfixed condition makes the negative control fail. This is targeted instrumentation, not a fully sanitized engine.

Also corrected the newly merged resource-fetch test's stale getCase call (standalone master repair #215). Ubuntu 22.04, Ubuntu 24.04 and Windows CI all pass at https://github.com/Globulation2/glob2/actions/runs/34300744186 . Please re-review the integrated head. No merge performed.

Giszmo (approved):

Giszmo:
@kylelutze merge please

genixpro:
@kylelutze — following up on Giszmo's re-approval and merge request from 9/9: the two rounds after your CHANGES_REQUESTED (integration conflicts through `56347c7cf`, then the resource-fetch test repair) addressed the document and dedicated-runner points you raised. Could you take another look and clear/update your review when you have a chance? This is otherwise ready and just needs your sign-off.

genixpro:
Rebased onto current master (`e3a01b051`) via `9482f04e7` to clear the conflict — it had drifted 19 commits behind. Only three files conflicted, all append/append (an independent new CI step and SConscript test target added on each side, and two independent doc sections in `test/README.md`); kept both additions in each case. The actual production fix (`MapGradientBuilding.cpp`'s `resource.type < BASIC_COUNT` guard) and the test file applied cleanly with no changes needed.

Verified locally before pushing:
- `scons -j8 release=1 server=0 clearing-gradient-test` builds clean on this head.
- `python3 test/run-savegame-safety-tests.py --check-preferences build/src/ClearingFlagGradientTest` passes.

CI is running on `9482f04e7`.

Given kylelutze is currently unable to review, and Giszmo already approved plus the two points from that review (document/dedicated-runner) were addressed in the two follow-up rounds since, I'll merge this once CI is green.

genixpro:
Dismissed @kylelutze's review as stale — both of its points (the dedicated reproduction doc, and the clearing-specific Python runner) were addressed in `edbb6ae4`, both threads are resolved, and @Giszmo re-approved on 2026-09-09 after those changes landed. Reopening is welcome if anything still stands.

That unblocked review but #194 landing on master put this branch into conflict, so I've merged current master in as well. Two files conflicted, both append/append against #194: an independent CI step in `.github/workflows/build.yml` and an independent test target in `src/SConscript`. Kept both sides in each; no production code touched by the resolution.

Verified locally on the merged head (macOS / Apple M3):
- `scons -j8 release=1 server=0 clearing-gradient-test` builds clean.
- `python3 test/run-savegame-safety-tests.py --check-preferences build/src/ClearingFlagGradientTest` passes, exit 0.

CI is running on the new head.

PR #186, opened 2026-09-07
Let units survive building destruction (#185)

### LINKED ISSUE #185: Let units survive a building being destroyed
It's frustrating to see warriors die in racetracks being attacked. I propose to either expel all units when the building is destroyed into the footprint of the building or to expel units as the capacity diminishes with damage.

Open questions are:
* expel all units once completely destroyed or expel gradually as the capacity shrinks in proportion with the damage?
* Should a partially destroyed swarm also lose its wheat reserves and defense towers their bullets/stone? I vote "yes".
* should units gain partial heal/feed from their time in the hospital/inn? I vote "yes".
* What to do with buildings that can occupy more units than terrain fields like the level 3 school? Expelling as capacity shrinks could fix this as warriors would attack only units as soon as the square is filled with units as long as the already in process blows to the building can't decimate the capacity by too much but we could give the building housed-units hitpoints when the building hitpoints go negative or just let the units that can't exit the building die. This would be a rare edge case anyway.

Giszmo:
Closes #185.

Units inside a building no longer die with it.

**Destruction.** `Building::kill()` puts occupants back on the map alive. Units inside or waiting to leave land on a free tile of the footprint first, then on the ring around it (`Building::findExpelTile`), so a building surrounded by enemies still spawns its occupants on its own tiles. A unit with no free tile at all still dies. A unit halfway through the door steps back onto the tile it came from.

**Partial benefit.** Expelled units keep the share of the meal or healing they already sat through; training stays all-or-nothing. A started meal still costs the inn one wheat (it cannot be split), so nobody eats for free; a unit that finds no wheat left gets nothing.

**Capacity is unchanged by damage.** A building holds its full `type->maxUnitInside` until it dies. The earlier revision shrank the inside capacity as the building took damage, with reservation cancellations and eviction ordering. Per @genixpro ("my preference is to keep capacity unchanged until destruction, then release the occupants") and @stephanemagnenat's "keep it simple" in #185, that mechanism is gone: `Building::updateInsideCapacity`, its per-step call, `Unit::leaveBuildingEarly`, the walker-cancellation ordering, the defence-tower and swarm exemption, and the two renderer clamps that only existed because capacity could shrink under a full building.

**Engine fix on the way.** `Game::checkBuildingsDoNotOverlapAndHealMissing` no longer stamps the footprint of a dead building back onto the map. `kill()` clears it, and re-adding it left a dangling id once the building was collected (segfault in `Map::setMapBuildingsDiscovered`).

**Test.** `test/BuildingExpelHarness.cpp` links the real engine and covers: a destroyed inn expelling its inside, exiting and entering units alive; a unit with no free tile still dying; healing kept pro rata. Each scenario runs real simulation steps and re-checks `Game::integrity`. Like `TrappedUnitLifecycleTest` it touches one resource gradient at setup, because the round-robin in `Map::syncStep` skips fields that were never allocated and a bare test world has none.

```
scons release=1 server=0 building-expel-test && ./build/src/BuildingExpelHarness
```

CI builds and runs it under a timeout, next to the other real-engine regressions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,6 +148,11 @@ jobs:
           scons -j$(nproc) release=1 server=0 immobile-unit-gradient-test
           ./build/src/ImmobileUnitGradientHarness
 
+      - name: Build and run the building expulsion regression
+        run: |
+          scons -j$(nproc) release=1 server=0 building-expel-test
+          timeout 300s ./build/src/BuildingExpelHarness
+
       - name: Build and run the hiring bucket regression
         run: |
           scons -j$(nproc) release=1 server=0 hiring-bucket-test
--- a/src/Game_io.cpp
+++ b/src/Game_io.cpp
@@ -322,6 +322,8 @@ bool Game::checkBuildingsDoNotOverlapAndHealMissing() {
 			const auto building = team->myBuildings[bi];
 			if (!building)
 				continue;
+			if (building->buildingState==Building::DEAD)  // kill() cleared its footprint
+				continue;
 			const auto x = building->posX;
 			const auto y = building->posY;
 			const auto type = building->type;
--- a/src/SConscript
+++ b/src/SConscript
@@ -648,3 +648,10 @@ if not env['server'] and 'custom-setup-test' in COMMAND_LINE_TARGETS:
     custom_sources += local.Object('CustomGameSetupHarness.o', '#test/CustomGameSetupHarness.cpp')
     custom_test = local.Program('CustomGameSetupHarness', custom_sources)
     local.Alias('custom-setup-test', custom_test)
+
+# Real engine regression for units surviving building destruction, built only by its explicit target.
+if not env['server']:
+    regression_sources = [source for source in source_files if source != 'Glob2.cpp']
+    regression_sources += local.Object('BuildingExpelHarness.o', '#test/BuildingExpelHarness.cpp')
+    regression_test = local.Program('BuildingExpelHarness', regression_sources)
+    local.Alias('building-expel-test', regression_test)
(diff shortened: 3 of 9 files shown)


Giszmo:
@kylelutze @stephanemagnenat @genixpro what do you think about this? Concept? Implementation?

stephanemagnenat:
My comments are in #185.

Giszmo's edit to bring the conversation here:

> I am fine with this, as long as we keep it simple. So based on that principle:
> 
> * If there is no spot for the units (e.g. level 3 school) if would kill the extra units.
> * I would also suggest to expel all at once, to keep the implementation simpler.
> * I would lose all resources upon destruction.
> * I would not give partial heal/feed
> 
> But I won't be upset if the implementation takes other decisions. I am just worried about the huge mass of Glob2's codebase and thus tending to simpler mechanics is a way to keep that a bit more in check.

genixpro:
I support letting occupants survive building destruction and escape onto the squares the building occupied. I also support preserving partial feeding/healing benefits when service is interrupted.

The part I would leave out is progressive capacity loss as the building takes damage. It adds thresholds, reservation cancellations, and eviction ordering for fairly marginal gameplay benefit. My preference is to keep capacity unchanged until destruction, then release the occupants.

So I support the core survival change and partial benefits, but would like the damage-dependent capacity mechanism removed. This is feedback on the intended gameplay, rather than approval of the implementation.

genixpro (approved):
Read the diff and the harness. `findExpelTile` correctly separates the footprint pass from the ring pass, `Game_io.cpp`'s dead-building skip is the right fix for the dangling-id crash, and the partial-benefit accounting (whole wheat per started meal, pro-rata hunger/heal) matches what we asked for. Matches the agreed scope: no progressive capacity loss, partial benefit kept. Three scenarios in `BuildingExpelHarness` cover the interesting cases (full expulsion, no-room death, pro-rata healing) and CI is green. Merging.

PR #187, opened 2026-09-07
Map: fix immobile-unit bookkeeping and refresh gradients on painted forbidden area

Giszmo:
Fix two pathfinding bookkeeping problems on the weighted-gradient engine now in master.

**Fresh-map occupancy.** `Map::setSize` initialized `immobileUnits` with zero, which denotes team 0 rather than an empty tile. Initialize it to `IMMOBILE_UNIT_NONE` (255), matching the loader. Empty cells on new maps no longer masquerade as immobile-unit obstacles.

**Forbidden-area painting.** Adding forbidden tiles now invalidates the team's cached building gradients, just as erasing already does. The next query rebuilds a route using the new restrictions. This invalidates all allocated swim classes through the existing `resetPathfindGradients` implementation; it does not add another refresh policy.

The original local-gradient indexing fix and local-dirty-flag changes are no longer part of the patch: #184 removed that pathfinder. The regression now uses full-map Uint16 gradients and all seven swim classes. It covers fresh-map occupancy, immobile units blocking exactly their own cells, closing the only gap through actual paint orders, and reopening it through an erase order. The new harness runs in Linux CI.

Validation on the adapted patch:
- The unfixed current-master code fails the fresh-map scenario and, independently, the forbidden-paint scenario. The weighted occupancy scenario already passes on the base.
- The fixed release harness passes all three scenarios. The latter two explicitly clear initial occupancy so the two bugs remain independently reproducible.
- All 170 local unit tests pass. Two fixed-build runs of G2 for 15,000 ticks produce identical replay bytes and every per-tick checksum.
- Final Ubuntu 22.04, Ubuntu 24.04 and Windows CI all pass, including the new regression on both Linux jobs.
- A separate 90,000-tick seeded game completed, but its repeat exceeded a 600-second timeout. That longer pair is not counted as a completed determinism check.

Run `scons release=1 server=0 immobile-unit-gradient-test` and `./build/src/ImmobileUnitGradientHarness`; use `fresh`, `occupancy` or `forbidden` to select a scenario. Normal game data is required; no display or external save fixture is needed.

This intentionally corrects routing behavior. Existing replay outcomes can change; repeatability of the fixed simulation is checked separately from equivalence to the broken behavior. Broad invalidation after painting may require more gradient rebuilding during brush use; allocation and rebuild remain on demand. Bradley approved the PR subject to this cleanup.


genixpro:
Bradley approved #187 subject to removing the obsolete local-gradient changes. The adapted head is `3549fd0c5`.

The remaining production changes initialize `immobileUnits` to its empty sentinel and invalidate weighted building gradients when painting forbidden tiles. The deleted local pathfinder remains deleted; current master's resetPathfindGradients handles all swim classes.

Verified with Apple Clang release builds in an isolated local build checkout:
- On unfixed master, the new harness fails `fresh` and independently fails `forbidden`; `occupancy` already passes, confirming the old indexing repair is obsolete.
- With the fix, all three harness scenarios pass across all seven swim classes, including closing and reopening the only route using real paint/erase orders.
- 170 CppUnit tests pass.
- Two completed `--nox games/G2.game 15000 1` runs produce identical replay bytes and every per-tick checksum.
- An additional seeded 90,000-tick game completed; its repeat hit a 600-second timeout and is not claimed as a completed determinism pass.
- Ubuntu 22.04, Ubuntu 24.04 and Windows CI pass on this head (run 34180070170). The new real-engine harness is now included in Linux CI.

Commands, controls, logs and repeatability digests are retained locally in `/Users/bradley/glob2-immobile-results/`. The initialization and paint fixes deliberately correct routing; old replay outcomes need not remain identical to buggy behavior.

PR #190, opened 2026-09-07
Remove high-confidence dead code (audit-driven, no behavior change)

Giszmo:
Removes only the **high-confidence** items from a repo-wide dead-code/duplication audit (see the "start here" list and full breakdown in the audit doc, happy to share it if useful). Duplication cleanups are deliberately *not* included here — this is dead-code deletion only.

Every item was re-verified with a fresh repo-wide `grep`/`git grep` immediately before deletion (not just the audit's original pass), and one item (`TextArea::load()`/`save()`) was pulled back out after a full build caught a live caller in `ScriptEditorScreen.cpp` that the text-only audit had missed — see that as the process working as intended, not a guarantee nothing else slipped through.

**Verified:**
- `scons release=1` — exit 0
- `scons server=1 release=1` — exit 0

**Not run:** the `test/` suite (built separately, see `test/README.md`) and a headless replay-determinism check. Flagging that explicitly since this repo's engine correctness bar is "identical replays mean identical behavior" — happy to run those before merge if wanted, opening as draft partly for that reason.

**What's in here (100 files, -2,354/+16 lines):**
- Entire dead class `UnitEditorScreen` (zero construction sites)
- A dozen unused `Utilities::*` helpers, several literal duplicates of used libgag equivalents
- Two dead map-generator files (`MapOldRandomRessources.cpp`, `MapOldIslandsRessources.cpp`)
- libgag: dead compressed-stream backend chain, including the whole `StreamFilter.{cpp,h}` pair whose own header admits "does not work as is"
- AI: unused `AICastor::computeResourcesCluster()` + backing buffer, several commented-out dead blocks, a few orphaned declarations
- net/yog: dead `ITDisconnected` message class, unused `YOGPlayerPrivateInfo` class, an unreachable duplicate branch in `YOGServerPlayer`, several unused accessors (`P2PConnection` plumbing, `getMapUploader`/`getMapDownloader`, etc.)
- Small dead accessors/fields scattered across gui/, team/, unit/, building/
- A handful of commented-out dead blocks removed outright per this repo's own convention against tombstone comments

No behavior change intended anywhere in this diff.


genixpro:
Bradley approved this cleanup. Integrated current master in `beea16c160271272d186e23757f557993e05c0f1`, keeping both sets of obsolete-code removals when resolving the Map.h / UnitStats.cpp conflicts. The test build no longer references the deleted StreamFilter source.

Verification on the final head:
- macOS Apple Clang release client build passed (`scons release=1 server=0 -j8 build/src/glob2`).
- Rebuilt CppUnit runner: 170 tests passed. Locally SCons received the shell PATH for Homebrew pkg-config; that local-only change was restored.
- Compared fresh release binaries for master `d3f2d284216f17003338788d841dcc0af903de21` and this cleanup on G2, Playground/8 Numbi, Archipelago and BigArena, 15,000 ticks each. Every replay byte and every per-tick checksum matched: 60,000 ticks per binary. This is simulation evidence for these fixtures, not coverage of every GUI/network path.
- Final Ubuntu 22.04, Ubuntu 24.04 and Windows CI all passed (run 34178426035).

The first Playground attempt in the new checkout could not find the extra local fixture; after copying that same fixture in, the comparison passed. No gameplay code was changed to obtain the result. Local logs and retained inputs/results are in `/Users/bradley/glob2-review-queue/`.

PR #191, opened 2026-09-07
Reduce Cortex placement work

genixpro:
Cortex placement scans repeatedly traverse the building array and recompute inn clearance for every candidate tile. This change snapshots live building positions, upgrade reservations, and inn occupancy once per placement search, then reuses them for candidate checks.

Scan order, scoring, RNG calls, observation cadence, and corner/wrap behavior are preserved. The snapshot is local to one search and needs no cross-tick invalidation.

### Performance and validation

- Eight-team Playground on integration revision `c71373bb`: **119.77 s → 84.19 s with Cortex alone**, about 30% less elapsed time; instructions **2.098T → 1.526T**. Replay, initial save, and final autosave are byte-identical.
- Single-run observations, with some concurrent build activity; see `docs/cortex-placement-performance.md` for scope and caveats.
- Standalone Cortex patch on master: complete SmallForTwo replay matches the original byte for byte.
- Optimized client rebuild and 57,600 geometry comparisons pass; the harness is added to Linux CI.

This PR contains only Cortex placement optimization and its tests/documentation. Save buffering is a separate, independent PR. Both target master; unpublished Maxima integration commits are excluded.

PR #192, opened 2026-09-07
Batch save writes while preserving serialization

genixpro:
Save serialization makes many tiny field writes through stdio. This change batches FileManager output in a bounded **16 KiB buffer**, including the checked atomic-save path.

Serialized fields, byte order, SHA1, and autosave cadence are unchanged. Positions include pending bytes; reads, seeks, flushes, and destruction drain in order. Atomic saves retain write/flush/seek/close failure checks before replacement.

### Performance and validation

- Eight-team Playground integration benchmark, holding Cortex optimization fixed: **84.19 s → 71.19 s after buffering**, 15.4% less elapsed time; instructions **1.526T → 1.217T**. Replay, initial save, and final autosave are byte-identical.
- This is an incremental, single-run measurement on integration revision `c71373bb`, not a universal speedup. See `docs/buffered-save-performance.md` for caveats.
- Standalone buffering patch on master, with original Cortex: complete SmallForTwo replay matches the original byte for byte.
- Buffer regression checks bytes/SHA1, boundaries, large writes, backpatching, reads, seeks, flushes, and destruction.
- Existing save-safety tests pass, including atomic/direct byte equivalence, reload checksums, truncated input, and write failures preserving the previous save.
- Buffer tests are added to Linux CI; existing Windows save-safety coverage exercises atomic saving.

This PR contains only output buffering and its tests/documentation. Cortex optimization is independently reviewed in #191. Both target master; neither depends on the other, and unpublished Maxima integration commits are excluded.


Giszmo (changes requested):
Reviewed the buffering logic end-to-end (byte order, position accounting including pending bytes, seek-before-drain, read/flush ordering, atomic temp-file+rename structure) — all correct, and CI is green on all three platforms.

Primary reason for changes requested: this PR currently shows `mergeable: CONFLICTING` against master and needs a rebase before it can land.

One real bug found along the way: `CheckedFileBackend::close()` sets `fp = NULL` and closes without draining the pending buffer first. It's latent today because `writeAtomically` always flushes before calling `close()`, but if `close()` is ever called with unflushed pending bytes, the next drain (e.g. at destruction) would write into a null `FILE*` — an assert in debug, a crash in release. One-line fix: call `drain()`/`flush()` at the top of `close()`.

(For the record, we also traced through whether `~BufferedFileStreamBackend()`'s call to the non-virtual-in-practice `writeBufferedData()` during destruction bypasses `CheckedFileBackend`'s checked override — it does, per normal C++ dtor dispatch rules, but that's correct-by-design here: a throwing checked destructor would just terminate the game mid-save, so "checked" is rightly enforced only at the explicit `flush()`/`close()` call sites that `writeAtomically` already uses, and `test/SavegameSafetyHarness.cpp`'s throw-before-flush case already covers the observable pending-bytes-at-teardown contract. No destructor change needed here — a comment on `~BufferedFileStreamBackend()` noting the drain is best-effort/unchecked would be a nice, non-blocking clarification for future readers.)

Please rebase onto current master and add the `close()` drain fix, then this looks ready.

genixpro:
@Giszmo Addressed the close/drain bug and integration conflicts through ce29af9a3.

`CheckedFileBackend::close()` now flushes pending bytes and checks stdio errors before closing. `writeAtomically` relies on that checked close instead of pre-flushing, so the existing SavegameSafetyHarness directly exercises pending bytes at close. The base destructor remains best-effort/unchecked, with a brief comment explaining that explicit flush/close provides error reporting.

On macOS, `scons -j6 release=1 server=0 savegame-safety-test buffered-file-test` and both executables pass. Save-safety ran through the shared disposable-profile runner with `--check-preferences`; coverage includes atomic bytes/reload, callback/open/rename failures, injected short write/flush failure, temporary cleanup, corruption controls and the merged pending-construction regression. Buffered-stream checks cover byte/SHA1 equality, boundaries, seeks, reads, flush and destruction. A separate negative-control binary with only the new close flush removed fails in `FileStreamBackend::write` on the null-file assertion, confirming the suite catches this bug.

The first integrated CI exposed a master integration error after #199: ResourceFetchTargetHarness still used getCase. Corrected to getTile here and independently in #215. Ubuntu 22.04, Ubuntu 24.04 and Windows CI all pass: https://github.com/Globulation2/glob2/actions/runs/34300746361 . Please re-review the verified head; no merge performed.

Giszmo (approved):

PR #193, opened 2026-09-07
Round-trip gradients: send workers to the resource nearest for the whole trip

Giszmo:
Follow-up to #184, targeting `feat/pathfinding-wip`. Two commits, each one mechanism, both in place on the Uint16 gradients. Net +390/−21 lines, 16 files, no switches, no tooling.

**1. Round-trip gradients (intermediary stops, hiring).** A worker fetching for a building walked to the resource nearest to *itself*, often on the far side of the base from where it delivers. A building now owns a round-trip gradient per resource type and swim class: every tile of that resource is seeded with the cost of carrying from beside it to the building, and `Map::propagateGradient` runs from those seeds (it now takes seeds at any cost, and an optional cost bound). Descending it minimises trip-to-resource plus trip-back, so `pathfindResource` with the target building follows it, falling back to the plain resource gradient when blocked or stale. Hiring and the next-resource choice after a delivery score by the same distance when a fetcher has already built the gradient, and by the plain distances otherwise (hiring looks at every needed resource of every building, far more than ever get fetched). Refreshed at most every 120 ticks, propagated 128 tiles beyond the cheapest fetch, freed after 500 ticks unused.

**2. Traffic lanes.** One counter per cell and direction, incremented as units step, halved every 1024 ticks (a lane is a slow structure: a given cell sees one unit every few hundred ticks). Building a gradient charges an extra cost for entering a *passage* (a cell walled on both sides across one axis) against its prevailing direction, in proportion to the excess of opposing over same-way units, up to three tiles; `directionByGradient` sees the same cost. The first units through a gap set its direction and later units heading the other way take another gap; on open ground and beside a single wall nothing changes. Derived state, not saved.

**Measured** (three binaries: base = this branch's head cddad361, rt = +commit 1, lanes = +both; each with the same scratch counters compiled in; 10 game minutes, 25 ticks/s; idle machine, runs sequential; per-team means; "blocked" = ticks a fetching unit could not step toward its goal):

| scenario | base | rt | lanes |
|---|---|---|---|
| Mazury 128², 2 Nicowar, 5 seeds: units / deliveries per team | 59.5 / 682 | 68.5 / 733 | 67.8 / 736 |
| balanced_for_2 64², 2 Nicowar, 5 seeds: units / deliveries / blocked | 85.9 / 968 / 489 | 82.5 / 924 / 352 | 81.1 / 922 / 362 |
| Mazury 4×4 = 512², 8 Nicowar, 2 seeds: units per team / peak RSS | 18.8 / 179 MB | 18.9 / 198 MB | 20.2 / 194 MB |
| choke, 1 gap: deliveries / blocked / sidesteps | 299 / 889 / 159 | 305 / 848 / 156 | 298 / 906 / 73 |
| choke, 2 gaps | 395 / 338 / 33 | 395 / 353 / 51 | 413 / 122 / 37 |
| choke, 4 gaps | 407 / 146 / 19 | 455 / 126 / 22 | 453 / 106 / 18 |
| ticks/s vs base, 128² at 50–100 units | 1 | 0.83–0.94 | 0.81–0.86 |
| ticks/s vs base, 64² at 50–150 units | 1 | 0.82–0.92 | 0.79–0.88 |
| ticks/s vs base, 512² at 400–750 units | 1 | 0.87–0.92 | 0.84–0.88 |

The choke scenario is a scripted save (balanced_for_2 start, both AIs passive, team 0's swarm ringed by a 3-thick stone wall with 1, 2 or 4 one-wide gaps to the north, wheat beyond, two inns and 30 workers inside; counts are team 0's). With two gaps the lanes cut blocked ticks by two thirds and deliveries go up; with one gap there is nothing to route around and nothing changes but the sidesteps; with four the gaps were never a bottleneck. On Mazury the round trip is worth about 15% more units; on the small balanced map the three are within the run-to-run spread (these games are deterministic, and a one-line change moves a 5-seed mean by about 5%).

Determinism: lanes binary, Mazury seed 1 twice, identical timelines and counters. Unit tests: 169 pass (3 new in GradientTest: seeds beyond the bucket window, cost bound, lane penalty); `server=1` builds.

Known limits: one shared traffic layer for all teams (enemy traffic through a gap counts too); the constants (half-life 1024 ticks, full penalty at 16 excess units, three tiles) are first values tuned on the scenario above; round-trip gradients cost one Uint16 field per actively fetched building, resource and class.


genixpro:
This is a very interesting direction. The round-trip routing and the possibility of traffic lanes emerging from local traffic history both seem worth exploring. The reported two-gap result is encouraging.

Because this changes a sensitive part of the engine, though, I would want a high standard of evidence before committing to merge. In particular, the lanes introduce a fairly complex, history-dependent heuristic into pathfinding. We need evidence both that the implementation is correct and that the resulting behavior is consistently useful. The existing tests and measurements are a good start, but I am not ready to approve it on the evidence currently presented.

You may already have more evidence from your experiments; if so, please put it on the PR so we can review it. What would help most:

- Reproducible before/after comparisons, ideally with side-by-side videos or visualizations of the same starting scenarios. Please separate current master, round-trip routing alone, and round-trip routing plus lanes so we can see which mechanism produces each improvement. Include ordinary maps as well as bottlenecks, and cases where the result is neutral or worse. Showing routes, congestion, and deliveries over time would make this much easier to assess.
- The fixtures, seeds, commands, and measurement tooling behind those comparisons, with multiple runs and enough simulation time to see whether lanes remain useful as traffic changes. Fresh performance measurements should use the optimized pathfinder now on master as the baseline.
- Stronger correctness coverage across terrain/swim classes, directional traffic costs, and changing traffic. Comparing generated fields against an independent shortest-path implementation would be particularly valuable, alongside real-engine tests of units creating and following lanes. CI should run the regression coverage.

Two specific implementation concerns also need resolution:

1. The bucket queue still derives MAX_STEP/BUCKETS from terrain cost alone (maximum step 42, 43 buckets), while lanePenalty can add another 30. That appears to break the bounded-edge-cost assumption: water plus a lane penalty can wrap into the current bucket or put future-cost entries in a bucket processed too early. Please reproduce this, fix the bound/queue handling, and add coverage for those combinations.
2. Traffic history influences routing but is not saved. Repeating a fresh simulation does not establish save/load continuity. Please demonstrate uninterrupted versus save-and-resume behavior and address the missing state if it causes divergence.

I would also favor reviewing round-trip routing and traffic lanes separately; their benefits, complexity, and performance costs deserve separate assessment. This is interest in the proposal and a request for more evidence, not merge approval.

Giszmo:
Rebased onto master (`f50ab27a`) and reduced to the round trip: the lanes commit is out of this PR (kept on `feat/pathfinding-lanes-archive`; it needs its own evidence on human-built layouts, which Nicowar never produces). Three commits: the round-trip gradients, and the two idle-gradient sweeps. Conflicts with the merged #184 were in `propagateGradient` (seeds at any cost and the cost bound now sit on top of @genixpro's optimised loop; deferred seeds join the bucket queue when the sweep reaches their cost), `Map.h` and `Building.h`. 172 unit tests and all `test/` harnesses pass; two runs of the same seed give byte-identical checksum sidecars.

Re-measured against current master with more seeds than the table above, since @genixpro asked for neutral and negative results too. Headless Nicowar vs Nicowar, 15360 ticks, `GLOB2_TEST_SEED` 1..N, both teams summed; "fetch" and "carry" are tiles walked toward a resource and back toward the building, counted at each completed step (tooling, not in the PR).

| Mazury 128², n=36 | master | round trip | diff | t |
|---|---:|---:|---:|---:|
| end units | 123.0 | 128.7 | +4.6 % | +1.5 |
| deliveries | 1354 | 1441 | +6.4 % | +3.3 |
| fetch tiles per delivery | 7.7 | 9.4 | +22 % | +14.5 |
| carry tiles per delivery | 10.0 | 7.6 | −23 % | −18.2 |
| round trip tiles per delivery | 17.6 | 17.0 | −3.5 % | −2.8 |

| balanced_for_2 64², n=24 | master | round trip | diff | t |
|---|---:|---:|---:|---:|
| end units | 164.2 | 158.5 | −3.5 % | −1.2 |
| deliveries | 1921 | 1876 | −2.3 % | −1.7 |
| fetch tiles per delivery | 9.5 | 11.0 | +16 % | +7.7 |
| carry tiles per delivery | 10.3 | 9.0 | −12.5 % | −7.2 |
| round trip tiles per delivery | 19.8 | 20.0 | +1.1 % | +0.6 |

Whole-game wall time (bash `time`, base and new alternated): 128² 3.39 s → 3.90 s (+14 %), 512² 8 Nicowars 9216 ticks 44.9 s → 45.5 s (+1 %).

Reading: the mechanism does exactly what it says (units walk farther to a resource that is nearer to the building, carry distance drops by a fifth), but in Nicowar's single-base games the two nearly cancel: the net trip is 3.5 % shorter on Mazury with 6 % more deliveries, and unchanged on the small balanced map. The +15 % units in the original 5-seed table was noise. The case for this change is Leo's multi-base layouts, where resources are shared between bases and the fetch-nearest-to-me rule sends carriers across the map; that needs a fixture built from such a game, which I will add here, and the 128² cost (the per-building round-trip fields, rebuilt at most every 120 ticks) needs profiling before this is a merge candidate. Not asking for review yet.

Giszmo:
Closed by the head-branch rename (feat/pathfinding-lanes → feat/pathfinding-round-trip); GitHub will not reopen a PR whose head ref is gone. Continued, rebased onto current master, as #244. Review history stays here.

PR #194, opened 2026-09-07
Clamp status-bar fills instead of aborting on overflow

genixpro:
A status bar whose current value exceeds its capacity aborts the game at `Game::drawPointBar`'s `actLength <= maxLength` assertion. Clamp the primary fill to the bar's capacity and the secondary fill to the remaining space, preventing both the abort and drawing past the bar. Simulation values are unchanged.

The crash was observed during world-overlay rendering in #139, but this helper already exists on `master` and is shared by ordinary 2D rendering. This PR is based directly on `master` and has no torus dependency. The original gameplay event that produced the count is still unconfirmed; the preserved autosave itself renders successfully.

Adds a standalone pixel regression covering both fill sections, all four orientations, negative counts, and zero capacity. CI runs it with OpenGL and software rendering on Linux, and software rendering on Windows.

Validation on macOS / Apple M3:
- Built the client and `point-bar-render-test` from this branch.
- Confirmed the new test reproduces the assertion with the unpatched `master` renderer.
- With the fix, `SDL_AUDIODRIVER=dummy build/src/point-bar-render-test gl` passes.
- `SDL_AUDIODRIVER=dummy SDL_VIDEODRIVER=dummy build/src/point-bar-render-test software` passes.


genixpro:
Rebased onto current master (`e3a01b051`) via `2219283e3` to clear the conflict. Three files conflicted, all append/append (an independent new CI step, SConscript test target, and `friend class` declaration added on each side); kept both additions in each case. The actual production fix (`Game::drawPointBar`'s clamp) and the new regression test applied cleanly with no changes needed.

Verified locally before pushing:
- `scons -j8 release=1 server=0 point-bar-render-test` builds clean on this head.
- `SDL_AUDIODRIVER=dummy SDL_VIDEODRIVER=dummy ./build/src/point-bar-render-test software` passes.

CI is running on `2219283e3`. Marking ready for review — this was sitting in draft with no reviews; requesting a look from @Giszmo.

PR #195, opened 2026-09-08
Reduce work in global gradient sweeps without changing results

genixpro:
Global gradient propagation visits every cell repeatedly, but only the strongest of its four directional neighbors can improve a cell. Compute that maximum once, use row pointers for neighbor access, and skip immutable 255-valued goals. This reduces hot-loop work while preserving the exact in-place forward/backward order, toroidal wrapping, byte-value semantics and convergence cap.

Comments explain the equivalence and wraparound constraints. There is no change to refresh scheduling, seed construction, the local gradient routine, or serialization.

### Measured performance

Fresh measurements of this exact patch against `master` at `b47c7d23b`, on an Apple M3 with an optimized Apple Clang build. Each value is median **core simulation CPU time** across three serial 15,000-tick runs:

| Scenario | Baseline | This PR | Reduction |
|---|---:|---:|---:|
| G2 | 4.134 s | 3.614 s | **12.6%** |
| Playground, eight Numbi AIs | 3.981 s | 3.430 s | **13.8%** |

The timer covers advancing `Game::syncStep` calls; it excludes AI order generation, startup, order execution and replay I/O. Temporary timing instrumentation is not included in this PR. Variant order was alternated and runs with competing compiler/game jobs were rejected. These are single-machine measurements, not FPS or portable speedup guarantees.

The earlier exploratory branch showed larger combined gains (22–24%) but included other changes and used a different base. Those figures are not attributed to this isolated patch.

### Validation

- Added `global-gradient-test`, linked to the real production method, and wired it into both Linux CI jobs.
- Compared every output byte against an independent priority-frontier solver across 3,000 fixed-seed random fields, toroidal seams, diagonals, thin dimensions, winding obstacles, mixed seed strengths, the distance cutoff and idempotence. The harness passes against both the unmodified baseline and this kernel.
- Ran **six saved-game scenarios for 15,000 ticks each** against the unmodified baseline: G2, gd-small-2ai, gd-large-4ai, gd-archipelago, gd-bigarena-long and fixed-seed eight-AI Playground. **All replay bytes and complete per-tick unit/building checksum sidecars matched**: 90,000 ticks per executable.
- The kernel and harness also passed with AddressSanitizer and UndefinedBehaviorSanitizer enabled for those translation units; remaining linked engine objects were uninstrumented.
- Optimized client build and `git diff --check` pass. Hosted CI is pending.

Run the maintained harness from the repository root:

```sh
scons -j8 release=1 server=0 global-gradient-test
./build/src/GlobalGradientHarness
```

This is the first of two focused PRs. The stacked follow-up #196 adds only a no-source fast path, its edge-case tests, and the [combined benchmark/evidence report](https://github.com/Globulation2/glob2/blob/codex/skip-inert-global-gradients/docs/performance/global-gradients.md). No experimental cache, local spiral rewrite, runtime switches or profiling machinery is retained.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,11 @@ jobs:
           scons -j$(nproc) release=1 server=0 terrain-test
           ./build/src/TerrainResourcesHarness
 
+      - name: Build and run the global gradient regression
+        run: |
+          scons -j$(nproc) release=1 server=0 global-gradient-test
+          ./build/src/GlobalGradientHarness
+
       - name: Build and run the LAN session regression
         run: |
           scons -j$(nproc) release=1 server=0 lan-test
--- a/src/SConscript
+++ b/src/SConscript
@@ -485,6 +485,12 @@ if not env['server']:
     local.Alias('terrain-test', terrain_test)
 
 if not env['server']:
+    gradient_sources = [source for source in source_files if source != 'Glob2.cpp']
+    gradient_sources += local.Object('GlobalGradientHarness.o', '#test/GlobalGradientHarness.cpp')
+    gradient_test = local.Program('GlobalGradientHarness', gradient_sources)
+    local.Alias('global-gradient-test', gradient_test)
+
+if not env['server']:
     lan_sources = [source for source in source_files if source != 'Glob2.cpp']
     lan_sources += local.Object('LANSessionHarness.o', '#test/LANSessionHarness.cpp')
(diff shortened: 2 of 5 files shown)


genixpro:
Closing in favor of the weighted pathfinder in #184, which we have decided to adopt and optimize directly. #184 replaces the main pathfinding use of this sweep kernel. The remaining Castor/Warrush helper uses can still benefit from selected optimizations if profiling supports them; this branch and its correctness/benchmark evidence remain available for reference.

PR #196, opened 2026-09-08
Skip global gradient propagation when no source can contribute

genixpro:
**Stacked on #195; merge that PR first.** This diff contains a five-line production fast path, focused tests, and the performance/equivalence evidence report.

A freshly seeded global gradient containing only values 0, 1 and 2 cannot change during propagation: zero is immutable, and neither 1 nor 2 can raise a free cell above its seed of 1. Return immediately when no input byte is at least 3, avoiding the two full-map sweeps that would otherwise discover this fixed point.

The comment states that invariant. The check deliberately accepts contributing seeds of 3 and 254 as well as 255-valued goals. It changes neither seed construction nor when callers refresh gradients, and introduces no cache or new persistent state.

### Measured performance

Fresh measurements of the exact clean patches against `master` at `b47c7d23b`, on an Apple M3 with an optimized Apple Clang build. Median core simulation CPU time over three serial 15,000-tick runs:

| Scenario | Baseline | #195 only | Both PRs | Additional reduction vs #195 | Total reduction vs baseline |
|---|---:|---:|---:|---:|---:|
| G2 | 4.134 s | 3.614 s | 3.510 s | **2.9%** | **15.1%** |
| Playground, eight Numbi AIs | 3.981 s | 3.430 s | 3.108 s | **9.4%** | **21.9%** |

The timer covers advancing `Game::syncStep` calls, excluding AI order generation, order execution, startup and replay I/O. It was identical in all timed binaries and is absent from these commits. Runs were serial, variant order alternated, and competing compiler/game jobs triggered retries. The Playground setup uses Numbi on all eight teams, seed 424242, plus the headless passive local controller on team 0. These are CPU-time measurements on one machine, not FPS claims.

The [evidence report](https://github.com/Globulation2/glob2/blob/codex/skip-inert-global-gradients/docs/performance/global-gradients.md) and [raw samples/checksum digests](https://github.com/Globulation2/glob2/blob/codex/skip-inert-global-gradients/docs/performance/global-gradient-results.json) include methodology and reproduction steps. They distinguish these results from the earlier 22–24% combined exploratory results on a different base, which also included a local-gradient rewrite. The experimental cache and local rewrite are not part of either PR.

### Validation

- Extended the maintained production harness with all-zero, all-one, all-two and mixed 0/1/2 fields, including one-cell inputs. Added seeds of 3, 254 and 255 at the last input position to check the threshold and full-buffer scan.
- The independent priority-frontier oracle still matches every byte for 3,000 random fields, toroidal seams, diagonals, thin dimensions, winding obstacles, mixed seeds, cutoff and idempotence.
- **Both exact clean patches independently match the unmodified baseline** on six 15,000-tick games: G2, gd-small-2ai, gd-large-4ai, gd-archipelago, gd-bigarena-long and fixed-seed eight-AI Playground. Replay bytes and complete per-tick unit/building checksum sidecars match across all **90,000 ticks per executable**. Timed runs also match baseline replays.
- Both kernel variants passed with the harness and gradient translation unit instrumented by AddressSanitizer/UndefinedBehaviorSanitizer; other linked engine objects were uninstrumented.
- Optimized client and final `global-gradient-test` builds pass. The harness is wired into Linux CI by #195; hosted CI is pending.

```sh
scons -j8 release=1 server=0 global-gradient-test
./build/src/GlobalGradientHarness
```

### DIFF
--- a/docs/performance/global-gradient-results.json
+++ b/docs/performance/global-gradient-results.json
@@ -1,0 +1,396 @@
+{
+  "metadata": {
+    "date": "2026-09-07",
+    "baseline_commit": "b47c7d23b",
+    "kernel_commit": "af8510619",
+    "cpu": "Apple M3, 8 cores, 24 GiB RAM",
+    "compiler": "Apple clang version 21.0.0 (clang-2100.1.1.101)",
+    "build": "scons release=1 server=0 CXXFLAGS= LINKFLAGS= (-O3)",
+    "core_metric": "CLOCK_THREAD_CPUTIME_ID around advancing Game::syncStep ticks; identical temporary instrumentation in all timed binaries",
+    "measurement_scope": "three serial repetitions of 15000 ticks, alternating order; AI order generation, startup, order execution and replay output outside core timer",
+    "original_experiment_base": "c71373bb2",
+    "fast_path_source_blob": "975f991a9a884602acefc80bc530d4748c2adab4"
+  },
+  "timing": [
+    {
+      "variant": "baseline",
+      "game": "G2",
+      "rep": 0,
+      "wall_s": 5.801870208000764,
+      "metrics": {
+        "ns": 4164610631,
+        "cpu_ns": 4144475680,
+        "calls": 15000
+      },
+      "replay_sha256": "1086f62c59f68840cd79e54a7e048cc4f06a6b49c440d8282620aab33b0a246b",
+      "interference": []
+    },
+    {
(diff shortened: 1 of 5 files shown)


genixpro:
Closing in favor of the weighted pathfinder in #184, which we have decided to adopt. Its main propagation path already avoids running without queued sources. Further performance work is going onto #184; this branch and its benchmark evidence remain available for reference.

PR #199, opened 2026-09-08
Use Tile for map cells and their accessor

genixpro:
Replace the French map-cell name `Case` with `Tile`, rename `getCase` to `getTile` and the map's `cases` storage to `tiles`, and update the editor's member pointer and current test harnesses. This is the map terminology portion of Stéphane Magnenat's #85, adapted to current master.

All string literals remain unchanged, including the historical `cases` save section. No data layout, algorithms, or gameplay changes are intended.

Validation:
- Lexical comparison confirms only the documented identifier substitutions outside comments, with every string/character literal preserved.
- macOS release client and server builds pass with both replacement patches applied together to master 88934ecfb.
- All 16 existing test executables pass on that combined code, including TestsRunner's 170 tests and replay/network/text-serialization tests.
- G2, playground-8numbi, gd-archipelago, and gd-bigarena-long: 15,000 ticks each, before/after replay bytes and every per-tick checksum identical (60,000 ticks per version).
- Each PR also runs its own Linux/Windows CI; see checks below.

Companion PR: #200

This replaces part of #85. The complementary spelling/comment patch and Kyle's already-landed cb7856eaa cover the selected spelling cleanup; the original broad casing-only sweep is intentionally omitted.

PR #200, opened 2026-09-08
Recover remaining spelling and name-clarity fixes from #85

genixpro:
Correct the remaining `recepientsMask` and `nscaned` spellings, give the placeholder-parser flag and future building type descriptive names, name the memory stream's storage `buffer`, and recover comment corrections from Stéphane Magnenat's #85 against the reorganized source tree.

The buffer name avoids shadowing existing `data` parameters. All string literals, serialized keys/section names, translation keys, script names, and filenames remain unchanged. No gameplay or serialization changes are intended.

Validation:
- Lexical comparison confirms only the documented identifier substitutions outside comments, with every string/character literal preserved.
- macOS release client and server builds pass with both replacement patches applied together to master 88934ecfb.
- All 16 existing test executables pass on that combined code, including TestsRunner's 170 tests and replay/network/text-serialization tests.
- G2, playground-8numbi, gd-archipelago, and gd-bigarena-long: 15,000 ticks each, before/after replay bytes and every per-tick checksum identical (60,000 ticks per version).
- Each PR also runs its own Linux/Windows CI; see checks below.

Companion PR: #199

This complements the map terminology patch and Kyle's already-landed cb7856eaa. The original broad casing-only sweep is intentionally omitted. Recovered changes credit Stéphane as co-author.

### DIFF
--- a/libgag/include/SDLGraphicContext.h
+++ b/libgag/include/SDLGraphicContext.h
@@ -180,7 +180,7 @@ namespace GAGCore
 		void _drawHorzLine(int x, int y, int l, const Color& color);
 		
 	protected:
-		//! Protectedconstructor, only called by GraphicContext
+		//! Protected constructor, only called by GraphicContext
 		DrawableSurface() { sdlsurface = NULL; }
 		//! allocate texture in GPU for this surface
 		void allocateTexture(void);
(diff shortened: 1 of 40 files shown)


Giszmo (approved):
Reviewed the full diff against the base tree: pure rename/comment-fix PR (recepientsMask→recipientsMask, datas→buffer, nscaned→nScanned, lcwp→lastCharWasPct, tltn→futureBuildingTypeId), balanced additions/deletions in every file, no serialized keys/section names/string literals touched, no hidden logic changes, CI green on linux 22.04/24.04 and Windows. Approving.

PR #201, opened 2026-09-08
Smooth unit animations with 32 poses, HD sprites and optional blur

genixpro:
This change makes unit animations smoother, particularly when units move slowly. Eight poses per direction produce visibly stepped motion; 32 poses provide finer sampling, and full-frame shutter blending makes skipped poses contribute at faster unit speeds. Normal rendering remains **25 FPS** with the same movement and animation cycle duration.

All seven core sets use 32 poses per direction: explorer flight; worker walk, swim and harvest/build; warrior walk, swim and fight. There are 1,792 logical frames across 2,816 PNG layers. Native sprites retain their original dimensions; a second complete texture set is rendered at 4× width and height for the merged HD renderer and zoom controls.

### Review follow-up: cold-cache work

Rebased onto `master` at `84a9b8b60`, preserving the merged entering-unit draw-origin fix and the new torus settings alongside motion blur. Native/HD assets and original Blender files are unchanged by this follow-up.

World-unit drawing now shares a soft 2 ms composite-generation budget per presented frame, including initial texture upload. Cache hits remain available; deferred misses use the current sharp native pose with transient team coloring until requested composites are ready. There is no additional fallback cache or retention cap. A single composite may exceed the budget. Transparent pixels are skipped during compositing.

A synthetic cold-cache test of 300 fast workers over 128 frames measured first/p95/max sprite times of 260.7/680.8/1117.8 ms with unlimited generation, versus 47.1/41.3/47.7 ms with the budget. The caches held 3,072 versus 128 composites afterward: work is deferred, not eliminated. These local results do not guarantee a busy saved game will meet its frame deadline.

PR #234 separately avoids repeated artwork loading and preserves artwork across matches. It remains a separate change. Material/rig redesign and consistent zoomed UI styling are separate artwork/UI follow-ups; decoupled rendering remains the planned engine follow-up.

### Rendering and controls

- Add full-frame shutter motion blur to world units, **disabled by default**. Existing explicit preferences are preserved. The main General Settings checkbox saves the choice; Cancel restores it. F8 toggles the same saved setting during gameplay.
- Cache final team-colored composites on demand, keyed by color and exact frame/alpha sequence. Cache hits draw one surface. Entries are **unbounded while artwork is unchanged** and clear when the Sprite is released or its texture pack reloads; the source recolor maps are cleared rather than retained as a second cache. Sharp world frames share this final-image cache, while sharp UI draws use one transient recolor surface.
- Add **0.75x, 0.5x and 0.25x** to the existing speed controls using 53, 80 and 160 ms ticks. Each draws once per tick; 0.75x is approximately 0.755x because the delay is integral milliseconds. Existing speed IDs and timings remain unchanged.
- Keep legacy action bases and serialized data unchanged; translate frame indices only at rendering. Direction-8 turning, portraits, editor previews, offscreen indicators and credits retain their sharp rendering and existing cadence.
- Preserve the original Blender sources, unit icons, movement positions, attack timing, save/replay formats and render-loop architecture. Keep all native sprites alongside the production HD pack.

The shutter follows the review's alpha-over sketch. It estimates the interval from current action speed and render interval, wraps only within the current action/direction, and does not store previous action/direction history. It is an approximate exposure: overlapping silhouettes and shadows retain alpha-over behavior. Cache generation uses premultiplied accumulation to allow drawing over any background; software output avoids the repeated rounding/darkening of legacy blits.

### High-resolution artwork

Rebased onto the merged HD renderer and organized-artwork changes (`87bdee0fd`). All seven animation sets now have 128×128, 152×152 or 160×160 textures, rendered from the same original rigs at the same 32 poses. This improves detail when zoomed in; it preserves logical sprite dimensions, movement and normal **25 FPS** rendering.

- Preserve native unit sprites and icons in `data/gfx`, and the untouched Blender sources in `datasrc/gfx/originals/units`.
- Retain approved HD layers in `datasrc/gfx/production/original-derived`; package them into `data/highres/v1` with the established artwork pipeline. Preserve the existing 60 world frames.
- Use four Blender workers under a combined four-CPU container limit on the eight-logical-CPU Mac. Scene preparation, render settings, layer hashes and packaging are reproducible; temporary scenes, downloaded tools and previews stay outside production folders.
- Integrate final-image caching with HD textures and logical draw dimensions. A shutter uses one resolution throughout; incomplete HD coverage falls back to native for the whole composite. The classic artwork setting and software backend retain native textures.

<details>
<summary>Native and HD sample poses at equal display size — all seven sets</summary>

Matching team-color layers are enlarged equally by the browser to compare texture detail. The complete animations retain their separate shadow layers. Images link directly to the committed assets; no review-image files are added to the branch.

<table>
<tr><th>Animation</th><th>Native texture</th><th>4× texture</th></tr>
<tr><td>Explorer flight</td><td>Native Explorer flight</td><td>HD Explorer flight</td></tr>
<tr><td>Worker walk</td><td>Native Worker walk</td><td>HD Worker walk</td></tr>
<tr><td>Worker swim</td><td>Native Worker swim</td><td>HD Worker swim</td></tr>
<tr><td>Worker harvest/build</td><td>Native Worker harvest/build</td><td>HD Worker harvest/build</td></tr>
<tr><td>Warrior walk</td><td>Native Warrior walk</td><td>HD Warrior walk</td></tr>
<tr><td>Warrior swim</td><td>Native Warrior swim</td><td>HD Warrior swim</td></tr>
<tr><td>Warrior fight</td><td>Native Warrior fight</td><td>HD Warrior fight</td></tr>
</table>
</details>

### Before and after

These slowed comparisons show eight versus 32 poses without motion blur. Both sides have the same cycle duration. They illustrate pose sampling, not a change to the normal **25 FPS** display cadence. Review images are attached to GitHub and excluded from the branch history.

**Worker walking**

Slowed worker walk: original eight poses versus new 32 poses

**Explorer flight**

Explorer flight: original eight poses versus new 32 poses

**Worker swimming**

Worker swimming: original eight poses versus new 32 poses

**Worker harvesting / building**

Worker harvesting / building: original eight poses versus new 32 poses

**Warrior walking**

Warrior walking: original eight poses versus new 32 poses

**Warrior swimming**

Warrior swimming: original eight poses versus new 32 poses

**Warrior fighting**

Warrior fighting: original eight poses versus new 32 poses

At equal points in the cycle, the old animation holds each pose while the new animation supplies intermediate poses:

Equal-time sprite samples showing the added intermediate poses

<details>
<summary>Matching original poses across all seven animation sets</summary>

Each pair shows the shipped pose on the left and the matching new pose on the right. The small shading differences are quantified below.

Before-and-after matching poses for all seven animation sets

</details>

### Rendering fidelity

The original **Blender 2.34** renderer was used after 2.79b failed to reproduce the original projection and warrior surface. Source rigs, animation channels, materials, lighting and layer separation are preserved. Native sprite dimensions stay unchanged; the additional HD renders multiply only the texture dimensions. Explorer requires a half-pixel camera alignment in the temporary scene copy.

Every fourth new pose was compared against the shipped sprites, across all eight directions and separate layers:

| Set | Pixel-identical comparisons | Mean RGBA error, 0–255 |
| --- | ---: | ---: |
| Explorer flight | 0/64 | 1.1101 |
| Worker walk | 64/128 | 0.2118 |
| Worker swim | 64/64 | 0 |
| Worker harvest/build | 128/128 | 0 |
| Warrior walk | 128/128 | 0 |
| Warrior swim | 64/64 | 0 |
| Warrior fight | 128/128 | 0 |

The residual differences are explorer shading/edge pixels and worker-walk shading. Worker-walk alpha and shadows match exactly. Overall, transparency matches in 682/704 comparisons. No redesigned motion was substituted. Comparison sheets were reviewed for original poses and loop boundaries; native framing retains the edge contact already present in explorer flight and warrior attacks.

### Validation

The 32-pose asset validation and interactive review covered every frame, all directions, three team colors, both rendering backends, live actions, portraits, editor previews, offscreen indicators and credits. Those assets are unchanged by the blur follow-up.

The rebased branch was rebuilt against `master` at `87bdee0fd`. Validation includes:

- Native release build, six render-pipeline tests, seven existing frame-index/sprite-centering tests, and `git diff --check` against the current base.
- Strict translation audit, translation regression, font coverage and text wrapping checks. All 33 supported languages include the new motion-blur labels.
- Settings/speed harness: default-off blur, old preferences, persistence, OK/Cancel/reopening, F8, language refresh, both speed selectors, multiplayer/replay restrictions, pause, and matching simulation/replay checksums.
- Exhaustive shutter bounds/wrapping checks across all seven sets, eight directions, all 256 progress values, and spans through 480.
- Software and OpenGL composite comparisons across all seven sets/eight directions and three team colors. Maximum GPU RGB difference from repeated draws: 3/255. Software difference: mean 5.38/255, max 10/255, including legacy background darkening.
- Unbounded retention: 13,172 entries / 76,099,904 CPU pixel bytes, with an early entry still a cache hit after passing 64 MiB and no retained source recolor map.

Reproduction commands are in `tools/unit-animation/MOTION-BLUR.md`; Linux CI now runs the blur/cache checks on both backends. CI status below applies to the newly pushed revision.

HD validation after installation:

- All 2,816 HD layers have the expected dimensions, paired-layer coverage and SHA-256 hashes. Production-only packaging reproduces all 2,887 runtime-pack files exactly.
- All 2,828 native unit/icon files and Blender originals remain unchanged; the existing 60 HD world records are unchanged and their layers are byte-identical. Both original-art archive preservation checks pass.
- All seven sets were reviewed across eight directions, loop boundaries and three team colors. Across all 1,792 poses, the largest alpha-centroid shift is 0.240 native pixels. The increased texture sampling changes antialiasing and reveals existing surface detail.
- The installed HD cache test passes all frame mappings, seven actions/eight directions/three colors, cache reuse, whole-shutter fallback for incomplete packs, HD/classic switching and logical dimensions. At one texture pixel per physical screen pixel, cached versus repeated OpenGL compositing differs by at most 3/255 in RGB.
- Game/editor/replay integration passes with HD and software fallback, including zoom, fullscreen, seams, input mapping, unchanged simulation checksums and texture release. The speed/settings/replay suite passes again with the installed pack. Actual world-unit captures confirm matching size with HD and optional blur.
- The macOS app bundle includes the complete HD pack. Normal rendering remains 25 FPS.

Warmed-cache benchmark on the rebased macOS renderer, 300 fast workers:

| Textures | Sharp | Repeated blur | Cached blur | Retained composite bytes across four cases |
| --- | ---: | ---: | ---: | ---: |
| Native | 1.58 ms | 5.37 ms | 1.17 ms | 73.0 MiB |
| HD | 1.57 ms | 5.49 ms | 1.18 ms | 1.42 GiB |

Every measured lookup hit. The four cases retain 3,456 composites; memory includes CPU pixels and GPU textures/mipmaps, with source textures additional. The cache remains unbounded. These sprite-only timings exclude first-use generation and are not whole-game frame times.

### Review follow-up: GPU team-color shader, bounded CPU cache

Giszmo's motion-blur design is unchanged; this addresses the memory regression the
generation-budget follow-up didn't fix. The final-image/composite cache is gone.
Rebased onto `master` at the same base as the cold-cache follow-up.

World units, portraits, editor previews, indicators and credits now share one
implementation in `DrawableSurface::drawSprite`. A sharp draw and each pose of a
motion-blur shutter are ordinary `drawSprite` calls, one per pose -- the sequence
your review sketch describes directly, now that there's no intervening cache.

On GPU with a working shader, each pose is one textured quad: a GLSL 1.20 program
samples the base layer and the *unrotated* team layer, reproduces the CPU HSV hue
shift itself, composites team over base in premultiplied space (unpremultiplying
once, the same algebra the removed cache used), and applies the shutter's pose
weight to alpha only. **No team-coloured or composite texture is ever created on
this path.** The shader is created with the GL context and destroyed before it's
torn down; a compile/link failure logs once and falls back to the CPU path for
that context's lifetime (also forceable with `GLOB2_DISABLE_UNIT_SHADER=1`, for
testing).

Without a working shader (software renderer, or that fallback), the unit sheet is
marked `dynamicTeamColor` and its CPU-recolored layers are now backed by one
sprite-wide, byte-accounted **64 MiB LRU**, keyed by frame/resolution/color, in
place of the unbounded per-frame map. A motion-blur shutter still samples one
resolution throughout (`Sprite::blockHasCompleteHD`): if any frame in the current
32-phase block lacks its HD counterpart, the whole block falls back to native
rather than mixing resolutions pose to pose.

#### Measured results (this machine)

- **Shader correctness** (`UnitTeamShaderTest`, all 1,792 poses, native+HD, three
  colors, plus the 12 team hues + 16-hue palette): HSV hue shift mean 0.059/255,
  max **1/255**, zero alpha mismatches, over 10,074,752 channel comparisons. Sharp
  and 30-delta motion-blur framebuffer vs. an independent from-scratch CPU
  composite: max **2.27/255**, over 4,194,304 comparisons.
- **Cache bound** (`UnitTeamColorCacheTest`): zero team-colored surfaces created
  under the shader, sharp and blurred. With the shader forced off, 20,000+ distinct
  colors through the cache never exceeded 64 MiB; repeated draws hit, long-evicted
  colors regenerate correctly.
- **12-team, 5,000-tick benchmark** (new `twelve-team-benchmark`; a synthetic
  render-loop soak test through the real `drawSprite`/`drawUnitMotionBlur` path --
  156 units across 12 teams, not a full Cortex/generated-map run): on two
  generated-map seeds, blur off and on --

  | Map | Blur | RSS growth (tick 2,500→5,000) | Frame p95 | Cache bytes (throughout) |
  | --- | --- | ---: | ---: | ---: |
  | river | off | 0 B | 0.83 ms | 0 |
  | river | on | -81,920 B | 1.73 ms | 0 |
  | craters | off | 0 B | 0.84 ms | 0 |
  | craters | on | -49,152 B | 1.70 ms | 0 |

  Versus the failing baseline this replaces: **4.0 GiB of composite storage and a
  5.79 GiB peak footprint after 5,000 ticks with 157 units.** Checksums matched
  exactly between blur on/off at every 250-tick mark; p95 is far under the 40/50 ms
  targets.
- Also passing: `TestsRunner` (180/180), `WinningConditionsHarness`,
  `HighResolutionIntegrationHarness`, `EnteringUnitSaveHarness`,
  `ImmobileUnitGradientHarness`, `EnteringUnitDrawHarness`,
  `MapRenderGeometryTest`, and the full settings/speed/replay/checksum suite
  (`run-game-speed-tests.py`).
- `MapRenderResizeHarness`'s `expected > 0` failure is fixed by this change
  (checked directly against a clean pre-change build). Its later resize sections
  need a taller desktop than this dev machine has; software mode now gets the
  same graceful bail-out GPU mode already had there, instead of an assert.
- Optimized client and server builds (`scons`, `scons server=1`) both build clean.

Full details and reproduction commands are in `tools/unit-animation/MOTION-BLUR.md`.
CI is running on this head (linux ubuntu 22.04/24.04, windows mingw-w64); I'll
follow up here once it reports. Holding this out of any merge automation until
then.

Native sprites, HD sprites and the original Blender sources are byte-for-byte
unchanged by this follow-up.

### Review follow-up: self-review cleanup

Ran a quality pass over the redesign above (commit 36e55be91) and fixed what it found:

- **Real gap**: the GPU-without-a-working-shader fallback picked HD vs. native per
  frame index, bypassing `blockHasCompleteHD` -- a unit sprite could reach it with
  a motion-blur shutter mixing resolutions pose to pose on that path (the shader
  path itself was already correct). Now gated the same way there too.
- **Efficiency**: `blockHasCompleteHD` rescanned up to 32 frame slots on every
  `drawSprite` call (once per pose during blur); it's now a one-bit-per-block
  cache recomputed only where the underlying HD arrays actually change.
- Dropped a duplicate `friend class GraphicContext;` declaration, and reworded a
  few comments that referred to "the removed composite cache" to describe the
  current code instead.

Full suite re-run with unchanged results (`UnitTeamShaderTest`,
`UnitTeamColorCacheTest`, `UnitHighResolutionCacheTest` GPU+software,
`HighResolutionIntegrationHarness`, `run-game-speed-tests.py`, `TestsRunner`
180/180, the 12-team benchmark).

### Part 1 follow-up: unify unit HD textures to 128x128 (memory optimization)

Follow-up to feedback that the HD memory footprint measured above (~1 GiB) wasn't
good enough. Root cause, found via direct GL instrumentation rather than guessing:
this engine never uses GL rectangle textures, so every individually-uploaded
texture is padded to the next power-of-two on upload. Unit HD textures rendered
at exactly 4x native size -- 128px for the 32px-native explorer set, but 152px
and 160px for the 38px/40px-native worker and warrior sets, both of which round
up to 256x256 on the GPU. 1,536 of 1,792 poses (86%) were in those wasteful
152/160 classes.

Fix: pin every unit HD texture to a fixed 128x128 canvas regardless of native
size (128 is already power-of-two, so this eliminates the padding waste
entirely), then re-rendered the six affected sets through the Blender 2.34
pipeline (explorer's own output is mathematically unchanged at 128 = 32*4 and
was confirmed byte-identical against a fresh render rather than assumed).

`frames.txt`'s `scale` column is a strict integer parsed via `operator>>`/`int()`
in four places and can't hold a fractional ratio (`128/38 ≈ 3.37`) without
corrupting the row parse, so unit rows now carry a sentinel `scale` of 0; every
consumer compares against the new `Sprite::highResolutionTextureSize` constant
directly instead of `native * scale`. Non-unit categories are untouched and
keep their literal `scale = 4`.

#### Measured results (this machine)

Same methodology as above: `DrawableSurface::allocatedTextureBytes()` with every
pose's base and team layer uploaded once via the real shader draw path (not the
CPU fallback cache).

| Configuration | GPU texture bytes | |
| --- | ---: | --- |
| Native only (no HD) | 42,991,616 | 41.0 MiB |
| HD, old 128/152/160 mix | 960,142,336 | 915.6 MiB |
| HD, new uniform 128 | 289,053,696 | 275.7 MiB |

**640 MiB (671,088,640 bytes, ~70%) reduction** in unit HD GPU texture memory,
with no change to the native-only baseline.

Also verified:
- `UnitHighResolutionCacheTest` (GPU + software), `RuntimePackCheck` (newly wired
  into `src/SConscript` -- exercising it for the first time surfaced and fixed an
  unrelated pre-existing bug: a benchmark loop referenced a `terrain` sprite that
  was never loaded, since terrain has no HD data and thus never appears in
  `frames.txt`), `HighResolutionIntegrationHarness`, `test_render.py` (8/8),
  `validate_runtime.py`, `package_runtime.py --check`, `runtime_provenance.py
  --check`, `validate_recovered.py`, `TestsRunner` (180/180),
  `run-game-speed-tests.py`, `WinningConditionsHarness`.
- The re-render's downsample-vs-native error profile (mean 36.5/255, alpha-mean
  1.7/255 across all 2,816 frames) closely matches the already-shipped pack's own
  profile against the same native references (36.8/255, 2.0/255) -- confirming no
  quality regression from the new fixed-size Blender scene path, just the same
  established rendering fidelity at a smaller, power-of-two-friendly size.

GPU texture compression (S3TC/DXT5) as a further reduction on top of this is
scoped as a separate, independent follow-up and intentionally not part of this
change.

### Part 2: S3TC/DXT5 texture compression

Landed sooner than originally sequenced -- the memory footprint after Part 1
alone wasn't a low enough bar for merge, so this followed immediately rather
than waiting for Part 1 to be validated in production first.

Rather than the offline stb_dxt-encoder-plus-DDS-sidecar design originally
scoped, `DrawableSurface::uploadToTexture()`'s existing HD mip-generation
loop now passes `GL_COMPRESSED_RGBA_S3TC_DXT5_EXT` as the internal format to
`glTexImage2D` directly, letting the driver compress during upload from the
same raw RGBA8 pixels it already builds for the mip chain. This produces the
exact same 4:1 ratio a precomputed DDS would (S3TC's compression ratio is
fixed by the format, not the encoder) with no vendored library, no new
build-time tool, and no new file format shipped alongside the PNGs --
confirmed on this machine's driver (Apple M3 / Metal-backed OpenGL 2.1) via
a standalone extension probe before writing any engine code. Detected via
`GL_EXT_texture_compression_s3tc` the same way `isTextureSRectangle` already
is; falls back to the exact previous uncompressed behavior when the
extension is absent, with `GLOB2_DISABLE_S3TC` forcing that fallback for
testing.

#### Measured results (this machine)

Isolating the HD-only delta from the constant baseline (fonts, cursor, UI
chrome) present in every measurement:

| Configuration | HD-only GPU texture bytes | |
| --- | ---: | --- |
| Old 128/152/160 mix (original) | 917,150,720 | 874.6 MiB |
| New uniform 128, uncompressed | 246,062,080 | 234.7 MiB |
| New uniform 128, S3TC/DXT5 | 61,501,440 | 58.65 MiB |

**93.3% total reduction** from the original baseline (874.6 MiB -> 58.65
MiB); the compression step alone is an exact 4.0x reduction on top of Part
1's 128x128 unification.

Also verified: `UnitTeamShaderTest`'s shader-vs-independent-CPU-reference
tolerance is unaffected by DXT5 quantization (mean 0.0586/255, max 1/255,
identical to pre-compression) -- these are flat-shaded sprites, not
photographic content. `UnitHighResolutionCacheTest` (GPU + software),
`RuntimePackCheck`, `HighResolutionIntegrationHarness`,
`UnitTeamColorCacheTest`, `UnitMotionBlurTest`, `TestsRunner` (180/180), and
`run-game-speed-tests.py` all still pass.

### CI fixes

Investigating why CI wasn't showing green surfaced two pre-existing issues,
unrelated to this change but blocking every platform from completing:

- `MapRenderResizeHarness`'s small-desktop bail-out returned exit code 1,
  which `run-savegame-safety-tests.py`'s `subprocess.run(check=True)` treats
  as a hard failure -- the opposite of the graceful, non-assert bail-out it
  was meant to be. CI's own environment (dummy driver / Xvfb default) reports
  a 1024x768 usable area against the 1800x1100 this harness needs, hitting
  this on every platform before ever reaching later steps. Now returns 0:
  the environment being too small to run this specific check isn't a test
  failure.
- `.github/workflows/build.yml` still invoked `UnitCompositeCacheTest` and
  `UnitCompositeCacheGPUCheck`, which stopped being built when the composite
  cache was replaced by the GPU team-color shader earlier in this PR;
  neither binary exists anymore. Repointed at `UnitTeamShaderTest` and
  `UnitTeamColorCacheTest`, and added `RuntimePackCheck`, which had no CI
  coverage at all despite exercising every category's shared frame parser.

### Master comparison and a fix for the review's reported bugs

Answering "did we actually reduce memory vs. master, not just vs. this PR's
own earlier commits": ran the existing `twelve-team-benchmark` scenario (12
teams, 156 units, full-viewport redraw every tick) against literal master
(`84a9b8b6`, this branch's actual fork point) and against current HEAD.

| Config | GPU texture memory | RSS |
| --- | ---: | ---: |
| Master (8-pose, old unbounded per-color CPU cache) | 85.25 MiB | 247.4 MiB |
| Current, classic/no-HD | 41.0 MiB | 164.5 MiB |
| Current, HD on (default) | 195.6 MiB | 573.1 MiB |

**In classic mode this already uses less memory than master, despite 4x more
animation poses** -- master's architecture CPU-recolored each frame per team
color and uploaded a separate GPU texture per (frame, color) pair, so its
GPU cost scaled with team-color count; the shader recolors live from one
shared texture, so GPU cost is flat regardless of how many teams are in the
match. (Side finding: the "4.0 GiB / 5.79 GiB peak" figure quoted earlier in
this PR's own history wasn't measuring literal master -- it was a worse
intermediate state partway through this PR's development, before the shader
rewrite. Real master, under this same realistic load, plateaus around 85 MB,
not gigabytes.) With HD on, memory is higher than master, fairly, since
master never had this capability -- that gap is exactly what the two changes
above spent today shrinking.

Also fixed, from the smoke-test round: **the square artifact under software
rendering.** Root cause: `DrawableSurface::drawSurface`'s manual per-pixel
alpha blend computed `x>>8` (divide by 256) instead of an accurate `x/255`
when combining a source/dest channel pair -- including at fully-transparent
source pixels, where the bias should be exactly zero. A single blend's
~0.4% error is invisible, but each motion-blur shutter step is one more
`drawSurface` call onto the same footprint, and a wide shutter at high game
speed stacks dozens of these, compounding into a visible darkening across
the sprite's whole bounding box. Verified directly: 60 stacked partial-alpha
draws over an opaque background drifted 200 -> 140 before the fix, and hold
exactly at 200 after. Replaced the divide with the standard
`(x+1+(x>>8))>>8` identity (exact `round(x/255)` for `x` in `[0, 65025]`),
applied per-channel-pair with the same masking the surrounding code already
uses. New `DrawableSurfaceBlendTest`, wired into `unit-blur-tests` and CI,
checks both invariants directly (transparent draws never perturb the
destination; a half-opaque source converges to a stable value rather than
drifting) and is confirmed to fail against the pre-fix code.

Checked the other bug from that same round (illegible "GL only" zoom label
on high-res screens, `MapZoomControls.h`): that file predates this PR
entirely (`3ad40cf3b`, already on master via #218) -- matches your own
"probably independent bug of this branch" note, so left alone here rather
than folding an unrelated fix into this PR's diff.

https://claude.ai/code/session_01TU4ZDnBKJKtfgGUNgsx194


Giszmo (changes requested):
**TLDR**: Going 8 → 32 poses quadruples the unit sprite set (704 → 2,816 files, 0.71 → 3.23 MB), but as plain hard-cut frames the extra poses are mostly wasted: at the default worker speed only 16 of the 32 are ever displayed, and even the fastest ground unit reaches about 20 of 32, so 32 plain poses look no better than 16 plain poses. What makes the animation smooth at every unit speed is motion blur in the renderer: average every pose the unit passed through since the last drawn frame, weighted by the time spent on each. That is a handful of extra alpha blits per visible unit, no new assets, and it turns the unused poses into visible smoothness; with it, the 32-pose set is clearly the best of the variants compared below, and its 2.5 MB is a small price. **Ask**: keep the 32 poses, add the full-frame motion blur described below, and keep the 3.5 MB of comparison GIFs/PNGs under `tools/unit-animation/comparisons/` out of the repository (attach them to the PR instead). Details and comparisons below.
---

### How often do the new poses actually get shown?
Walk speed (`performance[WALK]`) and `startImage[WALK]` come from the `kDefaultUnitTypes` table in `src/game/entities/Race.cpp`, loaded unconditionally via `Race::loadDefault()`; no map or race file overrides them. Running the real per-tick delta advance from `Unit.cpp` and this PR's `unitAnimationFrame` (`src/render/UnitAnimation.h`) against those speeds, at the default one render per tick, gives the number of distinct poses landed on per tile crossing:

| worker level | speed (delta/tick) | 8 poses (master) | 16 poses | 32 poses (this PR) |
|---|---|---|---|---|
| 0 (default, most units) | 16 | 8 of 8 | 16 of 16 | **16 of 32** |
| 1 | 21 | 8 of 8 | 16 of 16 | 25 of 32 |
| 2 | 26 | 8 of 8 | 15 of 16 | 20 of 32 |
| 3 (fastest ground mover) | 30 | 8 of 8 | 16 of 16 | 20 of 32 |

At level 0 the 32-pose scheme lands only on the even indices, so as hard-cut frames it is frame-for-frame identical to a 16-pose scheme; the odd half of the new assets is never drawn. At higher levels 32 poses does show a real but modest edge (20 to 25 distinct poses versus 16). The blend below is what turns the unused poses into visible smoothness.

Independent of speed: for actions that use the `direction == 8` branch, `(delta >> 5) * 32` can only ever select 8 of the 32 offsets in that block, so the 4x increase is structurally unreachable there.

### Repository cost
`git ls-tree -l` blob sizes of `data/gfx/unit*.png` (not `du`, which rounds every small file up to a disk block):

| | files | size |
|---|---|---|
| master (8 poses) | 704 | 0.71 MB |
| this PR (32 poses) | 2,816 | 3.23 MB |
| 16-pose variant used in the comparison (even-numbered sprite IDs of this PR, equivalent to every second rendered pose for every action and direction) | 1,408 | ~1.6 MB |

Separately, the PR commits about 3.5 MB of comparison GIFs and PNGs under `tools/unit-animation/comparisons/`. Those are review material, not game data, and would sit in every clone forever; please attach them to the PR instead.

### Blend instead of shipping more frames
The renderer draws one pose per frame; at 25 frames per second a level-0 worker advances 2 poses per frame and a level-3 worker 3.75 poses (delta advances by the unit speed each tick, 8 delta units per pose). Whatever is skipped between two frames is what strobes. Two fixes were prototyped on this PR's sprites:

1. **Two-pose cross-fade**: draw the current pose, then the next pose on top with alpha equal to delta's fractional position between them. Cheap, but it interpolates a single instant, so it can never blur more than the gap to one neighbour.
2. **Shutter motion blur**: average every pose the unit passed through since the last drawn frame, each weighted by the time it was on screen (a box shutter over the whole frame interval). This is what a camera would record, and it scales itself: 2 to 3 poses at level 0, 4 to 5 at level 3.

Six-way comparison of plain and cross-faded 8, 16 and 32 poses, fastest ground mover (level-3 worker, speed 30/tick), built from the actual master and PR sprites through the same team-colour HSV shift as `Sprite::getRotatedSurface`, upscaled 8x. First at real speed, one frame per game tick:

six schemes at real speed

The same sequence at 25% speed, each tick held for four frames:

six schemes at 25% speed

Plain 32 versus plain 16 is hard to tell apart, matching the table above. Cross-fading helps and helps more with 32 poses than with 16, but the fast unit still strobes because 3.75 poses go by between frames.

Shutter blur on the 32-pose set, four columns: plain, two-pose cross-fade, half-frame shutter, full-frame shutter. Level-3 worker (3.75 poses per frame), real speed on top and 25% speed below:

shutter blur, level-3 worker

Level-0 worker (2 poses per frame):

shutter blur, level-0 worker

The full-frame shutter is the one to take: it removes the strobing on fast units and reads as a running figure, and on slow units it degrades gracefully to a light blur. The half-frame shutter is a film convention, not a physical one, and leaves part of the strobing in.

### Implementation sketch
`GraphicContext::drawSprite` already takes an alpha parameter (`libgag/include/SDLGraphicContext.h:255`) and both render paths honour it: the GL path via the quad colour, the software path via the per-pixel blend loop in `DrawableSurface::drawSurface`. Magic effects, ghost buildings and particles already draw through it. Equal time weighting with ordinary over-compositing works by drawing the k-th pose with alpha = w_k / (w_0 + … + w_k), so the first pose is opaque and each later one is scaled by its share so far. Sketch, untested:

```cpp
// Motion blur: composite every pose the unit passed through since the last drawn frame.
int base = imgid;                                    // action base, before unitAnimationFrame
int step = unitActionStepSpeed(unit->speed, unit->action, unit->dx, unit->dy);   // delta per tick, as Unit::step uses it
int span = std::max(1, step * globalContainer->settings.getGameSpeedRenderInterval());  // delta advanced since the last drawn frame
int from = delta - span + 1;                         // shutter covers delta positions [from, delta]
int drawn = 0;
for (int d = from & ~7; d <= delta; d += 8)          // one 8-delta bucket per pose
{
	int w = std::min(d + 7, delta) - std::max(d, from) + 1;   // time spent in this pose
	drawn += w;
	int id = unitAnimationFrame(base, dir, d & 255);
	int dx = (unitSprite->getW(id)-32)>>1, dy = (unitSprite->getH(id)-32)>>1;
	globalContainer->gfx->drawSprite(px-dx, py-dy, unitSprite, id, (Uint8)(255 * w / drawn));
}
```

Game speed presets interact with this and the sketch accounts for it: presets 0 to 4 shorten the tick (40 down to 16 ms) but still draw every tick, so the motion per drawn frame and therefore the blur span are unchanged; presets 5 to 10 draw only every 2nd, 2nd, 4th, 5th, 16th and 16th tick (`Settings::getGameSpeedRenderInterval`), so the unit moves several ticks between frames and the shutter must cover all of them. Keying the span on the delta advanced since the last drawn frame, rather than on the tick, gets both cases right. Standing units have span 1 and draw exactly one opaque pose, as today.

### Cost
Per visible moving unit this is ceil(span / 8) + 1 sprite draws instead of one: 3 at level 0, 5 at level 3, more at the high speed presets. Each draw is two 38x38 layers (base plus team-colour layer). On the GL path that is a few extra textured quads per unit and not worth measuring. On the software path each blended pixel goes through the integer blend loop in `DrawableSurface::drawSurface`, roughly 20 instructions; 300 fast units on screen would add about 4 million blended pixels per frame, comparable to the terrain blit, so noticeable but not prohibitive on a software renderer. If that ever matters, the blended results repeat: the phase within a pose bucket (delta mod 8) only takes 1 value at speed 16 and 4 values at speeds 26 and 30, so there are 32 to 128 distinct blurred frames per action, direction and speed level, and they could be cached per team colour the same way `RotatedImage::rotationMap` already caches team-coloured frames. Not needed for a first version.

<sub>Edited 2026-09-09: consolidated this review and two follow-up comments into one post, replaced a GIF that exceeded GitHub's 5 MB image proxy limit, and extended the ask from a two-pose cross-fade to full-frame motion blur after prototyping both.</sub>

Giszmo:
**TLDR**: #201 adds **2.40 MB / +2,112 files** (0.68→3.08 MB, 704→2,816 files) and **~30 lines of real engine-logic diff** (test/tooling/docs aside; net -18 in `src/`, since it dedupes 4 call sites into one shared header) to go 8→32 poses. A **16-pose version costs half that — measured from real assets, not estimated: 0.86 MB / +704 files (0.68→1.54 MB, 704→1,408 files)** for the identical code diff (just two constants change). At the *default* worker speed, 16 and 32 poses look pixel-identical in play (32 only reaches 16 of its own 32 slots); at the *fastest* mover in the game (level-3, speed=30/tick) 32 poses does show a real edge — ~21 distinct poses vs 16's 16 — modest but genuine. Also prototyped a zero-new-assets alternative (blend adjacent original poses) below — it's cheap but doesn't look as good as either discrete option without more work.

---

Following up on the review above with the numbers requested and a fourth alternative worth considering.

### File count / size, precisely
`git ls-tree -l` blob sizes (not `du`, which rounds every small file up to a 4K disk block and roughly quadruples the apparent total):

| | files | size |
|---|---|---|
| master (8 poses) | 704 | 0.68 MB |
| this PR (32 poses) | 2,816 | 3.08 MB |
| **16-pose alt** (real measurement: exactly half of this PR's own shipped assets, selected by keeping every even-numbered sprite ID — verified this selection is equivalent to "every 2nd rendered frame" for every action/direction block, not just the worker-walk case in the earlier comparison) | 1,408 | 1.54 MB |

GitHub's own "2,963 files changed" diff stat is larger than the net growth because it also counts the 128 removed and ~581 modified-in-place files, not just new files.

### Reconsidering the value, honestly
At the *default* level-0 worker walk speed, 32 poses shows exactly the same 16 distinct silhouettes 16 poses would (see the review above) — no benefit there. But you're right that this isn't the whole story: at higher worker/warrior levels (up to the fastest ground mover in the game, level-3, `performance[WALK]=30`/tick), the 32-pose set does get more use — **~21 of its 32 slots**, vs a 16-pose scheme's **16 of 16** (fully saturated). That's a real, if modest, edge for 32 over 16 specifically for higher-level/faster units, at roughly double the asset cost. Reasonable people could land either way here; wanted the actual numbers on the table rather than "4x density" alone.

### Alternative worth prototyping: blend instead of add frames
Idea: instead of shipping more discrete poses, linearly blend the two *adjacent* original poses by delta's fractional position within its pose bucket (e.g. delta 80% of the way from pose 3 to pose 4 → blend those two images 80/20). Zero new sprite files, works at any speed, no oversampling-vs-real-speed mismatch to reason about.

Prototyped it (source: naive `PIL.Image.blend` cross-fade of the two neighboring master sprites, composited through the same team-color HSV-shift path as the real renderer) for the fastest mover in the game, alongside hard-cut 8, derived-16, and this PR's 32:

8-hard vs 8-blended vs 16-derived vs 32-PR, real time, fastest mover

Honest result: it's cheap (no new assets, works at every speed automatically) but a plain alpha cross-fade produces visible **ghosting/double-exposure** rather than a clean in-between pose — sprite art isn't laid out for pixel-aligned interpolation, so a limb in one pose doesn't line up with the same limb in the next. It doesn't look as good as either discrete-pose option as-is. Making this actually look good would likely need something smarter than a raw cross-fade (e.g. limb-aware/skeletal interpolation, or leaning into the ghosting as an intentional speed-blur style) — flagging as a direction worth exploring separately, not a drop-in replacement for this PR.

genixpro:
Implemented the full-frame shutter approach from your review in 89a245c51, keeping the 32-pose assets. Final team-colored composites are generated on demand and retained in an unbounded per-sprite cache; source recolor maps are cleared so we do not retain a second color cache. Motion blur defaults on and can be disabled in General Settings or with F8. Existing speed controls also include 0.75x, 0.5x and 0.25x using longer ticks; normal rendering remains 25 FPS.

The seven before/after GIFs and two comparison sheets are now GitHub attachments in the description. I removed both illustration-only commits from the branch history, so those files will not be merged into the repository. The separate high-resolution experiment is excluded.

Local release build, settings/speed and simulation/replay checks, shutter indexing, and software/OpenGL cache checks passed. The unbounded retention test kept an early entry after exceeding 64 MiB. With 300 fast workers, warmed cached blur measured 0.80 ms on OpenGL / 0.27 ms in software, versus 3.46 / 1.42 ms for repeated blending. These are sprite-only timings and exclude first-use generation. The description documents the alpha-over approximation, transition-history limitation, pixel comparisons and reproduction commands; Linux CI now runs the cache checks too.

Merged current master (`70a2c6f6c`) to resolve the translation conflict, completed the new labels in all 33 supported languages, and reran the release build and focused checks successfully. Strict translation, font coverage and text wrapping checks also passed. Ready for another review; CI will run on the updated head.

Giszmo:
The game loop is 25fps but the rendering loop could very well be decoupled from that. I don't know how much effort that would be but it would be helpful in both directions. If the cpu is lagging due to background processes, drop rendering a few frames to keep the simulation at pace and prevent network games from getting stuck for all and when resources are plenty, let the 100Hz roll.

I recently dug deeper into motion blur and realized that my concept was too simple. You can't just add motion blur at 25Hz and the eye no more notices the low frame rate. For one, something moving fast like the arms of a maxed out worker advances several pixels between frames, so you need blur to avoid the retina from perceiving successive hand borders but an observer's eye can follow that hand in which case you don't want it to be blurry. Imagine a ball flying through the image. If you stare at the tree behind, you want blur but if you follow the ball you want no blur at all. The only way to both is with a continuous picture and the closest we can get to that is a high refresh rate.

I also learned that animation movies sometimes predict where the viewer is looking, so they might blur the tree behind the sharp ball in the previous example even though the trees are still in the frame.

Anyway, glob2 is dead. Long live glob2 :wine_glass:

genixpro:
@Giszmo I agree that decoupling rendering from the 25 Hz simulation is a worthwhile engine improvement. Higher refresh rates would address something motion blur alone cannot, and being able to skip rendering under load while keeping the simulation on pace also makes sense.

I want to take this step by step and avoid turning this into a single monster PR. I've tried the current build again and am happy with this as an incremental improvement. The plan is to merge this as it stands, then tackle independent rendering and the interpolation needed for smooth motion between simulation ticks in a separate follow-up, preserving simulation and multiplayer/replay behavior.

I'm also running low on GPT-6 Astra credits after the web browser and mobile ports, so I might not have enough left for that larger engine change until my account renews. I agree the bigger improvement is warranted; I just want to keep its scope and timing separate from this PR.

genixpro:
@Giszmo would you be happy to approve this PR as it stands so we can merge this incremental improvement? The current changes include the optional cached motion blur and keep the simulation/render timing unchanged, and all CI checks are passing.

I agree with your recommendation: the plan for an upcoming separate PR is to decouple rendering from the simulation, including interpolation between simulation ticks, while preserving multiplayer and replay behavior. Let's keep that engine work separate so this PR stays manageable.

Giszmo:
https://github.com/user-attachments/assets/31601b6d-9d6c-4158-b8ca-0a57bc66bbd1

That motion blur though ...

Giszmo:
https://github.com/user-attachments/assets/20666b03-c491-474e-8ff8-486cb0bb37f2

I don't know. This is not how I imagined the globules to look like close up. That glossy skin deforming when they bend? @stephanemagnenat what do you say?

Also it's a bit weird how some things are high res now but others like the "inside" indicators scale terribly.
image

Giszmo:
It's a bit rough around the edges still and the resources needed are a pain. Loading a saved game it is too much to load all at once, wetting the cache. I think that has to be throttled if the game loop is not running at the target speed. For example, load 8 frames first and then, if we have air to breath, load more. Or more aggressively, load low res 8 frames first.


https://github.com/user-attachments/assets/52a75532-f193-4cd7-bcfd-3f999d72514c

That speed indicator on top goes red here. It goes deep green later. Like down to 5% during later play but the initial burst is definitely reducing the game speed and I have quite the machine here.

genixpro:
@Giszmo thanks for testing this. I've pushed a follow-up (`22534a7bd`) addressing the cache warm-up burst and rebased onto current master, including the merged entering-unit draw-origin fix (#231).

World-unit drawing now has a shared soft 2 ms budget per displayed frame for generating new composites, including initial texture uploads. Existing cache hits draw immediately; other units temporarily use their current sharp native pose while requested composites become available over subsequent frames. The cache remains unbounded, with no additional retained fallback cache. Transparent pixels are also skipped during compositing. A single composite can exceed that budget, so this is a mitigation rather than a hard frame-time guarantee.

In a synthetic cold-cache test of 300 fast workers, the 95th-percentile sprite frame time fell from about 681 ms to 41 ms (worst frame 1,118 ms to 48 ms). This defers work: after 128 frames, the budgeted run had generated 128 composites rather than 3,072. It doesn't prove your particular save now runs smoothly, and the earlier warmed-cache benchmark missed this startup cost.

We also have #234, “Load artwork once and retain it across matches,” which addresses repeated HD loading and invalidation separately. Its larger-unit-pack tests reported one map's startup dropping from about 3.28 seconds to 49 ms. I also hit repeated HD decoding during this branch's settings/replay tests; those passed with an extended timeout. #234 isn't included in the numbers above, so we should retest the combined changes before calling loading solved.

On appearance, these renders preserve the original models, rigs and materials. Improving how the glossy skin and deformations look close up is worthwhile, but it's a separate artwork project. The mismatch with the existing indicators also deserves a UI/zoom consistency pass. I haven't redesigned either here. Motion blur remains optional in General Settings or via F8, which should help distinguish blur artifacts from the underlying sharp artwork.

The client build, cache checks on both backends, settings/replay assertions, game/editor integration and translation checks pass locally; fresh CI will validate the pushed revision. If the warm-up slowdown persists, could you share the save used in the video so we can test that exact case? Decoupling rendering from the simulation remains the planned separate engine PR.

Giszmo:
**Concern up front: the composite cache introduced here has no upper bound, and its key space for a 12-team game with the HD pack is about 31 GB of system RAM plus 31–110 GB of texture memory.** It will never actually reach that — the process dies first — and that is the problem: nothing in the code stops it growing. At native resolution the same ceiling is ~1.9 GB, which is reachable and survivable but still worth a bound. Details, method and suggested fixes below.

Everything else I checked looks good: simulation and replays are untouched (no `src/unit`, no `Race.cpp`, action bases keep their serialized values), the game-speed tables index correctly against `GAME_SPEED_MINIMUM = -3`, and CI is green on all three platforms.

---

## Why the key space is large

`Game::drawUnit` routes **every** unit draw through `Sprite::getCachedComposite`, both the blurred branch (`GameRenderUnits.cpp:102`) and the plain one (`GameRenderUnits.cpp:105`), and entries are retained for the sprite's lifetime by design (`Sprite.cpp:406`, `compositeCache.emplace` at `Sprite.cpp:535`, cleared only by `reloadHighResolution`).

The cache key is the whole shutter — the vector of `(frame, alpha)` pairs — and the shutter start is `from = delta - span + 1` (`UnitAnimation.h`), so the key depends on the exact `delta`, not on the pose. That matters because `Unit::step` never restarts `delta`: it carries the remainder across tile boundaries with `delta += stepSpeed - UNIT_DELTA_QUANTUM` (`Unit.cpp:299`), i.e. `delta_{n+1} = (delta_n + step) mod 256`. A unit therefore visits `256 / gcd(step, 256)` distinct delta values, which is 256 for any odd step:

| walk case | step | distinct deltas | tile-crossing phases |
| --- | ---: | ---: | ---: |
| level 0 straight | 16 | 16 | 1 (exact loop) |
| level 0 diagonal | 11 | 256 | 11 |
| level 1 straight | 21 | 256 | 21 |
| level 1 diagonal | 14 | 128 | 7 |
| level 2 straight | 26 | 128 | 13 |
| level 3 straight | 30 | 128 | 15 |
| standing / harvest level 0 | 8 | 32 | 1 (exact loop) |

That drift is not new — master's 8-pose indexing does the same thing — but it used to be invisible. Now each of those delta values is a separate cache entry, which is the gap between "1,792 frames exist" and 27,312 keys per team colour.

## The numbers

Enumerated from the shipped tables: action bases from `UnitSkin.cpp` `SKIN_OFFSETS`, speeds from `Race.cpp` `kDefaultUnitTypes` (all four upgrade levels, so racetracks and swimming pools are included), the diagonal rule from `UnitTiming.h`, direction parity from `UnitGeometry.cpp:157`, and every delta the recurrence above reaches.

**27,312 distinct keys per colour → 327,744 entries at 12 colours** (any single game-speed setting).

| 12 colours | entries | RAM | texture, rectangle | texture, power-of-two + mips |
| --- | ---: | ---: | ---: | ---: |
| motion blur off, native | 21,504 | 0.13 GB | 0.13 GB | 0.32 GB |
| motion blur off, HD | 21,504 | 2.0 GB | 2.0 GB | 5.0 GB |
| motion blur on, native | 327,744 | 1.9 GB | 1.9 GB | 5.1 GB |
| motion blur on, HD | 327,744 | **31 GB** | **31 GB** | **110 GB** |

RAM is the RGBA surface (38x38x4 native, 152x152x4 at HD); texture is the GL copy each `DrawableSurface` allocates on top of it, sized exactly on drivers with `GL_TEXTURE_RECTANGLE` and rounded 152 → 256 plus mips elsewhere. The HD rows apply when OpenGL is enabled, since `loadExperimentFrame` returns early without `USEGPU` (`Sprite.cpp:364`).

**Turning motion blur off does not avoid the cache**, it only shrinks the key set — the plain branch still caches one composite per (frame, colour), which is the first two rows: 2.0 GB at HD.

About the speed slider, since it is easy to overstate: presets 0.25x through 2.5x all use render interval 1 (`Settings.cpp`), so they produce the *same* keys and change nothing. Only 4x and above skip frames (interval 2, 4, 5, 16 — confirmed at `EngineRun.cpp:48-49` and `503-509`, where `nextGuiStep` gates `gui.step()`), and `span = step * getGameSpeedRenderInterval()` (`GameRenderUnits.cpp:97`) then widens the shutter. Each individual preset lands at 27–28k keys per colour; only *changing* speed mid-game accumulates them, up to 133,744 per colour if a player visits every preset. Live network games are unaffected — `adjustableGameSpeed` is false there.

## How I verified it

The enumerator is a standalone program over `UnitAnimation.h`; to check it counts the same thing the engine does, I built the branch and drove `Sprite::getCachedComposite` over the worker-walk subset with the real delta recurrence. The enumerator predicts 5,184 keys per colour for worker walk; the live `compositeCache` held exactly 5,184 entries at 28.6 MiB native. Per-entry sizes and generation cost were measured the same way.

Per new entry at HD: about 440 KB and **4.45 ms** to build (0.27 ms native), against a 40 ms frame budget at 25 fps. `getCachedComposite` does `getHSV`/`setHSV` in doubles over 23,104 pixels x 2 layers x up to 5 frames. So roughly every 2,400 new entries costs another gigabyte and about 11 seconds of accumulated stutter.

## Suggested fixes, cheapest first

1. **Keep the per-frame recolour cache and composite at draw time.** `RotatedImage::rotationMap` has a natural ceiling of frames x colours = 21,504 surfaces (0.13 GB native, 2.0 GB HD) and this PR deletes it in favour of a key that is 15x larger. Drawing the shutter as N alpha blits is what master does and what the review sketch proposed; the composite only pays off when the same exact shutter recurs, and the delta carry means it often does not.
2. **Bound the cache** — a byte-budgeted LRU. A 64 MiB cap existed and was removed; the test now asserts unbounded retention.
3. **Quantise the shutter start** to `delta & ~7`. The drawn pose is `delta >> 3`, so it is bit-identical; only the sub-pose weighting rounds. That cuts 27,312 keys per colour to 7,352. Helpful, not sufficient on its own.
4. At minimum, **do not route the non-blur branch through the cache** (`GameRenderUnits.cpp:105`), so users who turn motion blur off get master's memory behaviour back.

## Smaller things noticed while reading

- `getColoredSurface` in `compositeOnly` mode clones the frame and hue-shifts it on **every** call with no caching (`Sprite.cpp:210`). The unit info panel and the map editor's unit picture still draw via `drawSprite`, so at HD that is a 152x152 clone plus a full HSV pass every frame one is on screen.
- Loading the HD unit set takes 1.25 s against 0.10 s native, partly because `loadExperimentFrame` rescans the whole pack manifest per frame (`Sprite.cpp:370`) — about 5M stream reads for 1,792 frames.
- `direction == 8` still maps to `(delta >> 5) * 32`, the first pose of each of the eight direction blocks, and motion blur is disabled for it (`GameRenderUnits.cpp:94`). Units at a resource are exactly this case (`MapPathfindRessource.cpp:53`), so the 4x pose count does not reach them.
- One merge conflict with master, in `src/render/GameRenderUnits.cpp`.

Giszmo:
image
At 40x speed, units jump several tiles per frame so the blur over their full rotation twice ... makes no sense. Also pausing the game shows me how the renderer fills in the missing assets over 3s until much later in the game when it's mostly saturated.
Also I suspect the blur to be off in that it takes one image at full opacity and then adds the others at the correct opacity. Still+blur. The blur should be the result of sampling over a longer shutter time with all images contributing equally.

genixpro:
@Giszmo Following up on the composite-cache memory concern flagged in review: it's gone. Pushed a redesign (028993f99) that replaces the final-image cache with GPU shader-based team coloring, plus a bounded CPU fallback. Summary in the updated PR description above; the short version:

- Each pose (sharp draw, or one step of a motion-blur shutter) is now a single ordinary `drawSprite` call — the exact sequence your sketch describes, since there's no cache sitting in between anymore.
- On GPU, a GLSL 1.20 shader does the HSV recolor and base/team compositing per-fragment from the *unrotated* team layer, so no team-coloured or composite texture is ever created. Verified against an independent CPU reference over all 1,792 poses (native+HD) and 28 team hues: max HSV error 1/255, sharp/blur framebuffer max error 2.27/255.
- Without a working shader (software renderer, or a forced fallback for testing), recoloring is backed by one 64 MiB byte-accounted LRU shared by world units and every UI preview, in place of the unbounded per-frame map. Verified it never exceeds the bound across 20,000+ distinct colors.
- New 12-team, 5,000-tick synthetic soak benchmark (`twelve-team-benchmark`, real render path, not a full AI/map-gen run): cache bytes and GPU-allocated bytes flat throughout, RSS growth zero-to-negative from tick 2,500→5,000 on both a river-seed and crater-lakes-seed run, blur on and off, versus the prior baseline's 4.0 GiB / 5.79 GiB peak. Checksums match exactly between blur variants.

CI is running on the pushed head now (linux 22.04/24.04, windows mingw-w64) — I'll follow up once it reports. Would appreciate another look whenever you have a chance.

Giszmo:
Amazing work! I tried it out and it has no initial lag, no rolling rendering, cpu is bored.

Please let the PR rest a bit this time to get @kylelutze and @stephanemagnenat also to try it out!

Bugs found on a quick smoke test:

With software rendering, units carry squares. This is more visible on speed x40.

image

Also in game:

image

probably independent bug of this branch: the scale doesn't scale with software renderer. Either remove it or make it work.

image

Edit: Yeah, "GL only" ... I can't read it on my high res screen.

genixpro:
@Giszmo Following up with the memory work and the bugs from your last smoke test.

**Memory.** The HD pack's actual root cause: this engine never uses GL rectangle
textures, so every individually-uploaded texture pads to the next power-of-two.
Units rendered at exactly 4x native -- 128px for the 32px-native explorer set, but
152px/160px for the 38px/40px-native worker and warrior sets, both padding up to
256x256 and wasting most of the allocation (86% of poses were in that wasteful
class). Fixed by pinning every unit HD texture to a fixed 128x128 canvas
regardless of native size, then re-rendering the six affected sets through the
Blender pipeline (explorer's own output is unchanged at 128 = 32*4 and confirmed
byte-identical against a fresh render). Layered S3TC/DXT5 compression on top --
`glTexImage2D` with a compressed internal format lets the driver compress
directly from the same RGBA8 pixels the mip-chain builder already produces, no
offline encoder or vendored library needed, same fixed 4:1 ratio a precomputed
DDS would give. Falls back to the previous uncompressed behavior when
`GL_EXT_texture_compression_s3tc` isn't present.

Measured via `DrawableSurface::allocatedTextureBytes()`, isolating the HD-only
delta from the constant baseline (fonts, cursor, UI chrome) present in every
measurement:

| Configuration | HD-only GPU texture bytes |
| --- | ---: |
| Old 128/152/160 mix | 917,150,720 (874.6 MiB) |
| Uniform 128, uncompressed | 246,062,080 (234.7 MiB) |
| Uniform 128, S3TC/DXT5 | 61,501,440 (58.65 MiB) |

**93.3% reduction** from the original baseline. `UnitTeamShaderTest`'s
shader-vs-independent-CPU-reference tolerance is unaffected (mean 0.0586/255,
max 1/255, identical to pre-compression) -- these are flat-shaded sprites, not
photographic content.

I also went back and compared against literal master (`84a9b8b6`, this branch's
fork point), not just this PR's own earlier commits, using the existing
`twelve-team-benchmark` scenario (12 teams, 156 units, full-viewport redraw every
tick):

| Config | GPU texture memory | RSS |
| --- | ---: | ---: |
| Master (8-pose, old unbounded per-color CPU cache) | 85.25 MiB | 247.4 MiB |
| This PR, classic/no-HD | 41.0 MiB | 164.5 MiB |
| This PR, HD on (default) | 195.6 MiB | 573.1 MiB |

In classic mode we're already using *less* memory than master, despite 4x more
animation poses -- master's architecture CPU-recolored per team color and
uploaded a separate GPU texture per (frame, color) pair, so GPU cost scaled with
team-color count; the shader recolors live from one shared texture, so GPU cost
is flat regardless of team count. (Side note: the "4.0 GiB / 5.79 GiB" figure
quoted earlier in this thread wasn't literal master -- it was a worse
intermediate state partway through this PR's own history, before the shader
rewrite. Real master, under the same load, plateaus around 85 MB.) With HD on,
memory is higher than master, fairly, since master never had this capability --
that gap is exactly what the two changes above spent today shrinking.

**Bugs from your smoke test:**

- *Units carrying squares in software rendering.* Root cause:
  `DrawableSurface::drawSurface`'s manual per-pixel alpha blend computed `x>>8`
  (divide by 256) instead of an accurate `x/255`, including at fully-transparent
  source pixels where the bias should be exactly zero. A single blend's ~0.4%
  error is invisible, but each motion-blur shutter step is one more
  `drawSurface` call onto the same footprint, and a wide shutter at high game
  speed stacks dozens of these, compounding into the visible darkening you saw.
  Verified directly: 60 stacked partial-alpha draws over an opaque background
  drifted 200 -> 140 before the fix, hold exactly at 200 after. Replaced the
  divide with the standard `(x+1+(x>>8))>>8` identity (exact `round(x/255)`).
  New `DrawableSurfaceBlendTest` (wired into CI) checks this directly and is
  confirmed to fail against the pre-fix code.
- *Illegible "GL only" zoom label.* Checked -- `MapZoomControls.h` predates this
  PR entirely (`3ad40cf3b`, already on master via #218), matching your own
  "probably independent bug of this branch" guess. Left it alone here rather
  than folding an unrelated fix into this PR's diff; happy to pick it up
  separately if useful.

Everything above is pushed and CI-covered. Also fixed two things CI itself
needed: `MapRenderResizeHarness`'s small-desktop bail-out was returning exit
code 1 (treated as a hard failure) instead of 0 for an environment too small to
run that specific check, and the workflow still referenced
`UnitCompositeCacheTest`/`UnitCompositeCacheGPUCheck`, removed when the
composite cache became the shader -- both were masking whether the actual
unit/HD tests were passing at all.

Given the above, I think this is ready -- would appreciate an approval whenever
you get a chance. Fully respect the ask to also get @kylelutze and
@stephanemagnenat's eyes on it; not trying to route around that, just wanted the
efficiency concern and the smoke-test bugs closed out first.

genixpro:
@Giszmo — dug into the two things from your 40x smoke test. You were right about one and, I think, not about the other; both with numbers below.

## 1. The full-rotation blur is real, and it's fixed (`33653b22d`)

The shutter span was `step * getGameSpeedRenderInterval()`, unbounded. Running the real `drawUnitMotionBlur` against the actual speed presets and `kDefaultUnitTypes` walk speeds:

| speed | interval | step | span | draws/unit/frame | distinct poses | cycles |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1x, worker L0 | 1 | 16 | 16 | 2 | 2/32 | 0.06 |
| 1x, worker L3 | 1 | 30 | 30 | 4 | 4/32 | 0.12 |
| 8x, worker L0 | 4 | 16 | 64 | 8 | 8/32 | 0.25 |
| 13x, worker L0 | 5 | 16 | 80 | 10 | 10/32 | 0.31 |
| **40x, worker L0** | 16 | 16 | **256** | **32** | **32/32** | **1.00** |
| **40x, worker L3** | 16 | 30 | **512** | **64** | **32/32** | **2.00** |

Exactly what you saw: at 40x the shutter covers the entire direction cycle, twice for a maxed worker, averaging every pose into a uniform smear — and paying 64 alpha blits per unit per frame to do it.

Scaling by the render interval was wrong in principle, not just unbounded. The sprite is drawn at **one** position, derived from the current delta. The ticks a fast preset skips have no swept position to blur along, so folding their poses in smears the unit in place rather than conveying motion — which is your "units jump several tiles per frame, so this makes no sense" point, stated as a code defect.

The shutter now spans one simulation tick at every speed. Fastest WALK/SWIM is 30 delta/tick, so it stays under a quarter cycle, never wraps, and costs at most 5 draws — down from 64. `UnitMotionBlurTest` now asserts that directly: for every step speed 1..30, every direction and every delta, no pose is visited twice and there are at most 5 draws.

## 2. The weighting is already an equal-weight box filter

> I suspect the blur to be off in that it takes one image at full opacity and then adds the others at the correct opacity. Still+blur.

I think this one is a false alarm — it's a running average, not still+blur. Drawing sample *k* with alpha = *w_k* / (sum of weights so far) makes each sample's **final** contribution equal. The first draw is at alpha 255 because it establishes the accumulator, not because it's privileged; it ends up weighted 1/N like the rest.

Effective final contribution per sample, accumulating `a*new + (1-a)*old` with the integer alphas the code actually emits:

| span | samples | ideal each | actual range | max deviation |
| ---: | ---: | ---: | --- | ---: |
| 16 | 2 | 0.5000 | 0.4980 – 0.5020 | 0.0020 |
| 32 | 4 | 0.2500 | 0.2471 – 0.2520 | 0.0029 |
| 64 | 8 | 0.1250 | 0.1216 – 0.1270 | 0.0034 |
| 256 | 32 | 0.0313 | 0.0275 – 0.0331 | 0.0038 |

Contributions sum to 1.0 and the worst sample is 0.4 percentage points off equal — that residual is 8-bit truncation in `255 * weight / drawn`, nothing structural. So the shutter was sampling correctly all along; what made 40x look wrong was the span, not the weighting.

## 3. Two things I have *not* addressed

- **HD/native mismatch.** You flagged the building "inside" indicators scaling badly next to HD units. That's real and it's inherent to this PR's scope: the HD pack is units only (all 2,819 files are `unit*`), so buildings and UI chrome stay native. Making that consistent is a separate art effort, not something I can close here.
- **Decoupling rendering from the 25 Hz simulation.** Still agreed, still follow-up work, per the earlier exchange.

Also worth flagging since your tests predate it: **motion blur now defaults to off** (`bbbb289b1`), so everything above is opt-in polish rather than default behavior.

Verified on the new head (macOS / Apple M3): full client builds clean; `UnitMotionBlurTest`, `UnitTeamShaderTest`, `UnitTeamColorCacheTest`, `UnitHighResolutionCacheTest` all exit 0; `run-game-speed-tests.py` passes all 6 checks. CI is running.

Still outstanding for a human call, not a test: @stephanemagnenat on the art direction — the glossy/deforming close-up look you raised on 2026-09-09.

genixpro:
## Motion blur is removed; the 32 poses stay

Following up on my last comment — rather than leave blur in as a default-off option, I've taken it out entirely (`bde2e284f`). It made movement look worse, and the 32 poses carry the smoothness on their own.

Gone: the shutter helper, the blur branch in `drawUnit`, the `motionBlur` setting and its round-trip, the General Settings toggle, the F8 binding, `UnitMotionBlurTest`, the sharp-vs-blur benchmark and `MOTION-BLUR.md`, and the three `[motion blur*]` strings across all 34 translation files. `drawUnit`'s only remaining difference from master is the pose indexing itself.

Kept, because none of it was blur-specific: the GPU team-color shader and bounded fallback cache; the x/255 software alpha-blend fix (a single blend's 1/256 bias is wrong regardless of what stacks the blends — `DrawableSurfaceBlendTest` covers it independently); and `UnitTeamShaderTest`/`UnitTeamColorCacheTest`, which used the shutter only to generate stacked alpha draws and now build those from a local helper.

## Why 32 and not 16

@Giszmo — your review argued the extra poses are wasted without blur, since at default worker speed only 16 of 32 are displayed. With blur gone that deserves a direct answer, and I think the 16-pose conclusion doesn't hold once **diagonal** movement is accounted for.

Diagonal travel advances delta by `(speed * 181) >> 8` (`UnitTiming.h`), so a diagonal tile crossing takes **more ticks** than a straight one — and therefore needs **more** poses to fill without repeating a frame. Running the real per-tick advance against the `kDefaultUnitTypes` walk speeds, counting rendered frames that repeat the previous pose:

| case | step | ticks/tile | 8 poses | 16 poses | 32 poses |
| --- | ---: | ---: | ---: | ---: | ---: |
| worker L0 straight | 16 | 16 | 8 held | 0 held | 0 held |
| **worker L0 diagonal** | 11 | 24 | 16 held | **8 held** | **0 held** |
| worker L1 straight | 21 | 13 | 5 held | 0 held | 0 held |
| **worker L1 diagonal** | 14 | 19 | 11 held | **3 held** | **0 held** |
| worker L2 straight | 26 | 10 | 2 held | 0 held | 0 held |
| worker L2 diagonal | 18 | 15 | 7 held | 0 held | 0 held |
| worker L3 straight | 30 | 9 | 1 held | 0 held | 0 held |
| worker L3 diagonal | 21 | 13 | 5 held | 0 held | 0 held |

"held" = a rendered frame showing the same pose as the one before it, i.e. visible stutter.

16 poses fixes straight movement but still stutters diagonally — and **worker L0 diagonal is the single most common movement in the game**: 8 of its 24 frames repeat, a third of the animation. 32 poses is the first count that reaches zero held frames in every speed and direction combination. That's the smoothness, and it's there with no blur at all.

The margin also isn't waste. Pose count is what lets the engine change unit speeds, or add a level or a movement type, without reintroducing stutter — at 16 there's no headroom, since the most common case is already short.

Verified on `bde2e284f`: full client builds clean; `DrawableSurfaceBlendTest`, `UnitTeamShaderTest`, `UnitTeamColorCacheTest` and `UnitHighResolutionCacheTest` (GPU and software) all exit 0; `run-game-speed-tests.py` passes all 6; `check_translations.py --strict` reports 0 structural errors and 0 missing keys across all 34 files.

Still open and still yours/@stephanemagnenat's call, not something tests settle: the close-up art direction, and the units-only HD pack leaving building indicators at native resolution.

genixpro:
Merged current master (`90f6bb262`) via `7643b2ab7`. Three conflicts; two were genuine reconciliations rather than mechanical resolutions.

**`src/render/GameRenderUnits.cpp` merged cleanly** — the important one. The branch had drifted far enough to be silently *reverting* two master changes: the failing-unit markers (#259) and the `isViewingGame()` rename. Both are back, with the 32-pose `unitAnimationFrame` indexing sitting on top of them. `drawUnit`'s only remaining difference from master is that indexing.

**`SettingsScreen.cpp` / `SettingsScreenGeneral.cpp`** — master replaced the widget-based settings screen with the form-based one (#236). This branch's only remaining content there was the motion-blur toggle (now removed) plus the below-1x speed presets. Took master's implementation wholesale and re-applied the slower presets on top: the choice row is built from `GAME_SPEED_MINIMUM` rather than 0, and since a preset's row index and its stored speed now differ by that offset, the setter maps the index back before clamping.

**`test/GameSpeedTest.cpp`** — master rewrote the harness for the form UI (`rows()`/`changeSetting`), dropping the `Selector`/`Text` lookups this branch asserted against. Kept master's harness, taught `select()` to convert a speed value to its row index, and restored this branch's coverage of the below-1x presets against the new `Row` API.

Verified locally on the merged head (macOS / Apple M3):
- Full client builds clean.
- `run-game-speed-tests.py` passes all 6, including the restored slower-preset assertions.
- `DrawableSurfaceBlendTest`, `UnitTeamShaderTest`, `UnitTeamColorCacheTest`, `UnitHighResolutionCacheTest` (GPU and software) all exit 0.
- `check_translations.py --strict`: 0 structural errors; `test_translations.py` passes.

CI is running on `7643b2ab7`. @Giszmo — with motion blur gone and the pose-count question answered with the diagonal-stutter measurements in the comment above, this should be ready for another look. The art-direction question for @stephanemagnenat is still the one thing tests can't settle.

stephanemagnenat:
Some notes (I do not have time to go in the code or the AI generated text in details unfortunately):
- On OpenGL I suggest (explore) doing the per team colorisation in a shader. This would save a lot of memory.
- I guess it is possible to indeed decouple the simulation from the rendering, and probably interpolate/compute the animation frame independently to take advantage of the more animation frames.
- I am not convinced by the motion blur, it makes the animation hard to read and feels over cheesy.

Giszmo:
> * I am not convinced by the motion blur, it makes the animation hard to read and feels over cheesy.

Motion blur is only over-cheesy on the main menu where it actually does not belong and I said so before. If we abandon the low 25fps, motion blur will not make so much of a difference but it does look just factually smoother and closer to what the eye would see if it saw the real motion and never cheesy at all.

PR #202, opened 2026-09-08
Refresh front-end menus with a live colony background.

genixpro:
Glob2’s opening screens currently put glossy, heavily framed controls over a repeated grass texture. This proposal gives the first impression a warmer cream-sage-and-gold refresh and shows a peaceful, living colony behind the menus.

The colony is a real Reach to Infinity game, starting from a bundled semi-mature save. It keeps developing while menus are open. The existing renderer, simulation, AI, widgets, labels and translations do the work.

## Why this direction

- Show the game’s colorful world immediately, with movement and activity that reflect what playing Glob2 feels like.
- Give text a warm, readable cream-sage panel, and use gold sparingly to emphasize the main action and selected tabs.
- Bring back a recognizable game identity with a rounded wordmark: the first “o” is the faceless, four-limbed Glob, with a small central opening that preserves the letterform while giving the Glob more visual weight.
- Carry the same treatment through the existing front-end flows, while preserving the gameplay and editor interfaces.

This is a substantial stylistic proposal, and I’d like broad feedback before merging.

## Before and after

| Current main menu | Proposed main menu |
| --- | --- |
| Before | After |

Eight seconds of the actual simulation at normal speed, with the new wordmark, warmer panels, and unit/building indicators. Only this GIF loops; the colony runs continuously in the menus:

Live colony

| Settings | Custom-game setup |
| --- | --- |
| Settings | Custom game |

| Post-game results | German at 640×480 |
| --- | --- |
| Results | Compact German settings |

## Scope and implementation

The shared theme covers main/campaign/tutorial menus, map and saved-game selection, custom-game options, all settings tabs, LAN/online front-end widgets, editor entry/map creation, credits, results and replay-saving dialogs. Existing layouts remain where practical. Team colors and post-game charts remain intact.

The background enables the existing renderer’s unit and building indicators: carried resources, health/hunger, worker assignment, occupancy, and resource bars. This makes colony activity easier to read while the menu panel keeps controls legible.

The new wordmark was developed with image generation using the repository’s app icon and worker sprite as references. It is bundled as a PNG; menu initialization fits it once and recovers alpha from its pale background for clean compositing, including the letter openings. The normal text heading is retained as a loading fallback.

`FrontendTheme` extends the existing `Style` interface; scoped activation restores previous fonts and styling around gameplay/editor sessions. `MenuColony` owns its own `Game`, view, clock and RNG, advances at 25 ticks/sec, and draws with `Game::drawMap`. Updates are limited to two ticks per menu frame. It pauses during games, replays, editing and focus loss/minimization, then resumes without hidden-time catch-up.

There is no new rendering engine, UI framework, shader, production worker thread, network session, or second engine loop. Core simulation changes are three null-GUI guards. A small software sprite-tint fix preserves transparent pixels. Menu simulation temporarily isolates the shared RNG and replay/training-data writers. The colony is silent, keeps the existing menu music, and has no periodic reset, added population cap or persistence across launches. A bundled still covers first paint and load failure. The AI rename and strategy changes are outside this PR.

## Validation

Latest wordmark/panel revision: normal client and harness builds pass, as do the presentation-count, bounded-text, RNG/global isolation, determinism, scoped-style, and fallback checks. Updated software screenshots cover the screens shown above, with the wordmark also inspected at 640×480. Earlier OpenGL, server, and soak results are recorded separately below; those longer checks have not been repeated for the wordmark revision.

- Normal optimized client and server-only builds passed locally; resource installation includes the save and still.
- Software and OpenGL checks passed for RNG/global restoration, replay/training-record isolation, real-match checksum determinism, nested theme restoration, missing/incompatible assets, bounded updates and pause/resume.
- Actual SDL event-loop cancellation passed across 14 front-end screens. Main-menu mouse/keyboard routes, map selection, replay-dialog typing/editing/cancellation, and real game/replay/results/editor return paths passed.
- Captures inspected at 640×480, 1152×720 and 1920×1080, plus German at 640×480. Live/fallback resolution changes passed in both renderers without changing simulation state.
- Existing game-speed/pause/replay tests passed, including all seven deterministic engine checksums. Focused colony/navigation tests are added to both Linux CI toolchains.
- The generator reproduced the bundled save byte-for-byte on its original engine base (`f2b6d4329`). Its generation revision is documented; the upstream pathfinding fix changes freshly generated worlds, while the bundled snapshot remains compatible.
- Warm-cache samples on Apple M3: theme/still setup 13–22 ms; snapshot load 27–32 ms. The final-background ten-minute run averaged 0.40 ms per simulation update call and 3.9 ms per complete software menu frame; peak resident memory was 60 MiB for the whole harness. Injected clicks continued to work.
- One-hour integration soak passed: 79,282 frames, tick 12,000 → 101,989, mean frame 4.98 ms, maximum 119 ms, peak resident memory 60.5 MiB. This run began before final text/sprite visual fixes; the separate ten-minute run covers the corrected background and injected input. Both preceded the final rebase onto `88934ecfb`; builds, isolation/navigation/session/display checks, existing speed tests and a one-minute smoke passed again on the rebased tree.

The desktop relocked during native testing. **Before merging:** manually check live lobby/multiplayer setup, physical wheel input, real minimize/focus transitions and display-setting controls. Automated renderer and session tests do not replace those checks.

[Design, generation commands, detailed coverage and measurements](https://github.com/Globulation2/glob2/blob/2669bf1cb2719588c4ee27df2fced82a1d1b18dd/docs/frontend-refresh/README.md).

## Feedback requested

Do the new Glob wordmark and warmer palette feel like Glob2? Is the panel opacity/readability right? Is the main menu’s density and hierarchy comfortable, particularly at small sizes? Does the moving colony with its status indicators add life, or draw too much attention? Please also flag any front-end widget or flow that still feels inconsistent.


Giszmo:
Visual feedback: Not seeing resources carried by the units tells me there is something going on about the simulation. It's not the real game as a background for the screen. If we don't use the actual game engine for this background, visuals will diverge as we change the game. The 8s video is a looped video but the game background is continuous I assume.

I'd support moving to a menu widgets library to better deal with high resolutions but I see that's not what's happening here. I would not mind the menus being less playful. It looks a bit cleaner but also sterile.

genixpro:
Thanks for the feedback. This is intended as an incremental improvement over what we have today. Introducing a new widget library is a bigger lift, but I agree that is the direction we should take next, particularly for better layout and high-resolution support.

The background does use the actual game engine: Reach to Infinity issues orders, the normal simulation advances, and the existing map renderer draws it. It runs continuously while the menus are open; only the eight-second GIF in the PR loops.

We deliberately hide game stats and some resource-related overlays to keep the background from becoming too distracting. That presentation is separate from the simulation, so it continues to follow the engine as the game changes. The goal here is to improve the first impression without taking on a full UI replacement in the same PR.

genixpro:
@Giszmo Following up on your point about carried resources: I've enabled the existing unit and building indicators in the menu background. That brings back carried-resource icons, unit health/hunger bars, and the building status indicators using the normal renderer. This was just enabling an existing draw flag; the background continues to run the same real simulation as before.

Here's the updated main menu:

Main menu with unit and building indicators

And an unobstructed view from the same colony renderer, without the menu panel or translucent wash, so the indicators are easier to inspect:

Colony with the existing game indicators

The client build and focused regression checks pass. I think the extra visual feedback makes the colony's activity easier to read. Does this feel closer to what you'd expect, or do the full indicators make the menu background too busy?

Giszmo:
I'm a bit speechless you merged this without any approval. Give us a bit time to breath. I get it, you are excited and want to move fast but this change was not agreed on.

I'm not generally against it but it's inevitably breaking stuff. I don't have much time but you haven't tried the credits screen for example.

stephanemagnenat:
I agree with @Giszmo , I am fine with the new layout, but **not at all with the new visual look**. I find that the new flat design completely loose the nice Globulation 2 visual vibe. And it won't be that hard to keep it, in the graphics dumps I provided there were high-resolution UI element that could be used to keep the nice Globulation 2 UI visuals with the new layouts.

So please @genixpro try bring back the nice Globulation 2 style buttons, check boxes, progress bars and window corners. Thank you!

Giszmo:
@genixpro please reply here as I might also try to apply the cute style to the new menus. Reverting would be the wrong approach as the menus changed a lot for new features. Also menus shouldn't be rendered in the game loop as they were.

stephanemagnenat:
I fully agree that the way forward is to try to re-use the old style graphics, in high resolution, over the new menu layout and code structure!

PR #203, opened 2026-09-08
Add browser builds with persistence, WebGL2, and YOG cross-play

genixpro:
Globulation 2 runs as a full-page WebAssembly browser client, sharing the native game's simulation, menus, editor, file formats and YOG protocol. Desktop browsers can play single-player games, persist and import/export files, and connect to native YOG games through a WebSocket gateway.

This branch is rebased onto master `4c16a8406`, including the current custom-game lobby, landscape picker, start-quality view, map-generation service, interface scale, artwork and rendering changes. Mobile playability is a separate PR.

### Behavior

- Hardware-accelerated WebGL2 is selected automatically; software rendering is the fallback. `?renderer=software` and `?renderer=webgl2` allow explicit selection. WebGL context restoration and the torus overview are supported.
- Browser screens and loading jobs run through scheduled lifecycles. Landscape and start-quality dialogs retain their parent setup. Queued cancellation is handled before advancing a loader, and editor replacement preserves the original map until success.
- Resizing preserves the selected interface scale and sends logical dimensions to screen layouts. Desktop secondary-click input cancels construction placement and cycles sidebar views without opening the browser context menu. This behavior is covered by real mouse input tests; the earlier missing-right-click report was not reproduced on desktop.
- Settings, progress, maps, saves and replays use durable browser storage, with visible failures, retry and import/export. New profiles start muted.
- Browser multiplayer uses WSS through the host gateway; native clients retain TCP/WSS. Browser LAN discovery/listening is unavailable.
- Generation follows master's `GenerationService` and `Fertility::forMap`. Browser previews advance one candidate per frame, but each generation roll and fertility calculation is synchronous and can stall a frame. This PR does not promise bounded latency within those calculations.
- The browser import replay fixture lives separately from the repository's shared determinism baselines. Save, replay and network compatibility floors follow current master.

### Menu presentation

Desktop playtest fixes use the browser's system cursor (also overriding older saved cursor preferences), keep the colony visible around the custom-game lobby, and activate menu theming when returning from a finished game. The old grass was reproduced on the returned setup screen, where it had inherited the still-live game's disabled theme. Results and their replay-save dialog now present together. The custom lobby's extra fullscreen panel also affected master's full drawing path, so that correction is shared.

Visual regression tests in `browser/tests/menu-presentation.spec.js` fail on the pre-fix build and pass with both renderers. Focused menu checks pass across Chromium, Firefox and WebKit, including normal/high-density cursor behavior; native `MenuColonyHarness` verifies presentation and menu/simulation isolation. Windows CI's two stale Linux harness paths are corrected. User playtesting and renewed approval are still pending.

### Validation and review

The September 14 desktop revision has local native and WebAssembly release builds and focused browser coverage for imports, storage, cancellation, editor flows, input, resizing, landscape selection and start-quality details. Native screen/session, custom-setup and savegame-safety harnesses pass.

For `games/cross-replay.game`, seed 42, 1,500 ticks, the per-tick checksum artifacts from master, the rebased native build and WebAssembly are byte-identical (SHA-256 `0eacdb13c144e8a92c55baff9289587c1509a444d136f641e59f4f3989dcdaa2`; final checksum `32e96f66`). This covers macOS ARM64 versus WebAssembly for that scenario; Windows/Linux equivalence and the full multiplayer matrix have not been re-established locally after this rebase.

Regression sources are in [`browser/tests`](https://github.com/Globulation2/glob2/tree/codex/browser-experiment/browser/tests) and [`test/EngineSessionHarness.cpp`](https://github.com/Globulation2/glob2/blob/codex/browser-experiment/test/EngineSessionHarness.cpp). The interface-scale test fails before the resize fix and passes afterward. Hardware graphics checks use Chromium's Metal backend on an Apple M3; forcing WebGL on the headless SwiftShader backend produces loading timeouts, while renderer auto-selection falls back to software there. Firefox and WebKit automation are additional coverage, not substitutes for actual Safari/Edge release testing.

The author will desktop-playtest this rebased build before requesting renewed maintainer approval. CI and independent review remain pending; this is not a merge request based solely on local automated checks.

### Build and play

```sh
python3 browser/setup.py
scons target=web release=1 -j6
python3 -m http.server 8765 --bind 127.0.0.1 --directory build/emscripten/client/release
```

Open `http://127.0.0.1:8765/`. [`browser/README.md`](https://github.com/Globulation2/glob2/blob/codex/browser-experiment/browser/README.md) documents the pinned SDK, tests, renderer selection, persistence and gateway deployment. The previously shared hosted demo predates this rebase; use a fresh build for this playtest.


Giszmo:
This is genuinely exciting. A browser build is the single biggest thing we could do for
reach: "click a link and you are playing" removes the packaging, distro and driver
problems that cost us most new players. Being able to put a Play button on
globulation2.org is something I have wanted for a long time, and I appreciate that you
went at it by compiling the real engine rather than reimplementing the game in JS —
around 130 lines of shipped JS glue and only 8 `__EMSCRIPTEN__` sites across `src/` and
`libgag/` is a much better outcome than I expected. The shared `scons/sources.py`
manifest is the right call.

Some things I want to be explicit about before this lands, mostly because the cost of
this port is not in the browser directory — it is in the shared code.

**Generated map layouts change.** ADR 005 states this outright. I accept the reasoning:
mixing the synchronized generator with libc `rand`, time-based reseeding and shared
static Perlin tables meant a seed never actually reproduced a map, and a noise helper
could perturb unrelated noise including clouds. That was a latent bug and the new
seed-to-map contract is worth having. But it is a user-visible change for existing
desktop players, so it should be called out in the release notes rather than discovered.
Related: the ADR says bit-exact generation across native and Wasm is not yet established
because of floating-point height maps. What happens today if a browser and a native
client generate a map in the same session?

**Two screen-execution models now coexist.** ADR 003 is honest that `execute()` remains a
compatibility polling host, overlay modal loops are unconverted, and calling a legacy
child `execute()` from a callback still blocks. That is reasonable for an incremental
migration, but half-finished migrations tend to stay half-finished once the motivating
work ships. Could you write down what is left as a tracked list, so it does not become
folklore?

**C++20 coroutines in the loader.** `CooperativeTask` brings new lifetime rules — jobs
must outlive nothing they borrow, no reentrant advancing — into the load path, which is
exactly where subtle bugs are most expensive. You raise the hand-written state machine
alternative yourself in ADR 004; I would rather keep the coroutines, but I want that
tradeoff visible to whoever debugs it in two years.

**Networking was rewritten.** `NetConnectionThread` and its message queues are gone in
favour of the `NetTransport` interface. I checked that `TcpTransport` still runs a real
worker thread, so desktop keeps its threading model — good. Still, every desktop
multiplayer path now runs through new code, and multiplayer is the area where we have
the fewest users to catch regressions for us.

**Small thing:** three surviving files lose their `// Copyright (C) ... Bradley Arsenault`
line (`src/FertilityCalculator.h`, `src/net/NetConnection.h`, `src/net/NetConnection.cpp`)
and `NetConnection.h` loses its `///` documentation comments in the rewrite. They are
your lines to remove, but in a GPL project I would rather keep headers intact and
document the new interface.

What reassures me most is the shape of your evidence, not its volume: freezing the old
fertility algorithm in `LegacyFertilityReference.h` as an equivalence oracle is exactly
the right instinct, and the new native harnesses mean desktop is not just being trusted
to still work.

One expectation to set: the description defers production hosting and operations, and the
current demo is a single-player bucket at about 31 MB with no loading indicator. So this
PR is what makes globulation2.org hosting possible, not what delivers it. I am glad to
treat hosting as separate work — I just do not want anyone reading this thread to think
the Play button arrives with the merge.

Thank you for the ADRs and for the candour in `status.md` about what is still
experimental. That made this much easier to review.

Giszmo:
So what's the goal of this? Instant custom games and campaigns (single player) from our website or YOG and LAN, too? The former would already be awesome and could be its own branch if the consequences for merging this to master were somehow too controversial.

genixpro:
Thanks — the intended scope is now clearer after cleaning the branch up.

This PR makes custom games, campaigns, saves/imports, and the existing YOG flow work in the browser. YOG goes through the WebSocket gateway and supports browser/native cross-play. LAN is compiled out of the WebAssembly UI: browser sandboxing does not provide the direct TCP/listen/discovery model that Glob2 LAN uses.

For generated maps, YOG clients do not independently regenerate the selected map. The host supplies the `MapHeader` and file ID, and a client that does not already have that exact map downloads the map file before the match starts. The floating-point caveat in ADR 005 therefore affects portability of offline seed-to-map generation; it should not create browser/native map divergence inside one YOG match. I agree the desktop-visible generator change belongs in release notes.

Longer term, I would like the browser release to grow a web-native multiplayer entry flow on top of this cross-play foundation: shareable match links, guest or lightweight identity, and instant matchmaking without requiring account creation first. That product flow is follow-up work rather than part of this implementation cleanup.

I also restored the three copyright lines you identified and added API documentation to the new `NetConnection` interface in the rewritten branch. The PR is now five commits based on current `master`; the migration/status artifacts from the exploratory branch are gone.

Giszmo (changes requested):
Requesting changes, but I want to lead with what this review is *not* about.

I was worried about the +16,330 / −4,203 and about what a browser port would cost the desktop
game. I built both sides and measured rather than guessed, and on those two questions the PR
holds up: **the simulation is bit-identical and the engine gets slightly simpler.** The four
changes below are dependency and packaging issues, not architecture.

## Requested changes

### 1. OpenSSL is now a hard requirement for every desktop build — please put it behind a flag

`SConstruct:166` appends to `missing` and fails the build if OpenSSL is absent, for any
`server=0` configuration. There is no flag and no fallback, so every packager on every
platform now needs OpenSSL to build a client that may never open a `wss://` connection.

The size cost is concentrated in one file. Attributing the binary growth per object on the
link line:

```
src/net/WssTransport.o     828,479 bytes of .text
all 362 other shared .o    +162,873 bytes total (+5.5%)
```

`WssTransport.cpp` is 168 lines, but it includes `boost/asio.hpp`, `boost/asio/ssl.hpp` and
four `boost/beast` headers. The linked binary goes 5,486,152 → 7,656,528 bytes (+40%, `.text`
+1.50 MB) and picks up `libssl.so.3` and `libcrypto.so.3`.

I am not asking you to drop it — `NativeTransport::open` (`src/net/NetTransport.cpp:106`)
selects `WssTransport` for a `wss://` address, so this is reachable code that desktop
cross-play needs. I am asking for a build option (default on is fine) so a distro or a
minimal build can compile the client without OpenSSL and Boost.Beast, with
`NativeTransport` falling back to TCP-only.

### 2. `debian/control` was not updated

`Build-Depends` still has no `libssl-dev`, and nothing under `darwin/` changed. The CI jobs
install `libssl-dev` and `mingw-w64-x86_64-openssl` explicitly, so CI stays green and the
first thing that breaks is a package build. Please add it (and whatever the macOS bundle
needs) in this PR.

### 3. Explain the regenerated determinism baselines

`games/cross-replay.game` is unchanged, but both baselines were replaced:

- `tests/baselines/cross-replay.replay`: 256,668 → 610,606 bytes, diverging at byte 24
- `tests/baselines/cross-replay.checksums`: same size, identical for the first 4,668 bytes,
  then 9,209 differing bytes

That reads alarming, so I checked it before writing this: the engine is bit-identical on that
exact save (evidence below), so this is a new, longer recording rather than a behaviour
change. But no in-repo test compares against `cross-replay.checksums` — the only in-repo
consumers of `tests/baselines/` are two Playwright specs that use the `.replay` as an import
fixture. The oracle these files serve lives in the sibling cross-replay tooling, so replacing
them silently rebases something outside this repo.

Please say in the PR description why they were re-recorded and confirm the sibling tooling
was regenerated to match. If they did not need to change, revert them.

### 4. Hoist the per-tile work out of the `FertilityCalculator` kernel loop

`Job::advance` (`src/FertilityCalculator.cpp:82`) spends its `operations` budget per *kernel
cell*, and `kKernelSide` is 31 — so 961 iterations per tile. Each of those calls
`State::coordinate()` (`:72`), which is `cursor / map.getH()` and `cursor % map.getH()`.
`Map::getH()` returns a runtime member (`src/map/Map.h:155`), so that is real integer division
961× per tile where the old nested loop did it once, plus a `coordToIndex` and `distance[]`
lookup that used to be hoisted out of the kernel loop.

I did not benchmark this, so treat it as a code reading. The scope is narrow — only maps older
than `FILE_FORMAT_VERSION_PRE_FERTILITY` reach it, at load time (`src/Game_io.cpp:306`) — but
computing `x`, `y` and the grass/reachability test once per tile instead of 961 times is a
small change and keeps the incremental version honest against the one it replaces.

## Evidence for the parts I am *not* asking you to change

Merge base `84a9b8b6` vs head `7962c30e`, both built `scons -j4 release=1 server=0`, identical
flags (`-O3 -std=gnu++20`), native Linux.

**Simulation is bit-identical.** Headless `--nox`, `SDL_VIDEODRIVER=dummy`:

| run | master | this PR | final checksum |
|---|---|---|---|
| 20,000 ticks, `cross-replay.game` (SmallForTwo, 2 AI) | 1680 / 1650 / 1656 ms | 1672 / 1664 / 1668 ms | `c44b9dfa` both |
| 20,000 ticks, `gd-large-4ai.game` | 6643 / 6558 ms | 6599 / 6632 ms | `3d4ce420` both |
| 1,500 ticks with `GLOB2_CHECKSUM_SIDECAR=1` | — | — | 3,660,536 bytes of per-tick, per-unit and per-building checksums, `cmp`-identical |

Save files cross-load in both directions: a save written by this branch loads on master and
yields the same `checkSum() = 302ef289` after 500 ticks.

**The line count is not engine code.** By destination: `browser/` +3,251, `data/` +2,614, tests
+2,593, docs and ADRs +1,153, build system +840, deploy and CI +385. Shared engine code moves
+5,494 / −4,070 across 225 files, net +1,424. `src/` + `libgag/` total goes 134,145 → 136,076
lines, +1.4%. `Unit.cpp`, `Map.cpp` and `Building.cpp` are untouched; `Game.cpp` is −1 line.
There are 11 `__EMSCRIPTEN__` sites in shared code across 6 files, and `browser/` ships ~267
lines of JavaScript against 2,031 lines of Playwright tests.

**Two things I expected to be debt and are not.** `src/net` excluding the test suite goes
2,507 → 2,088 lines, and that 2,088 already includes the new `WssTransport.cpp` and
`gateway/Gateway.cpp`; native-only networking is about a third smaller, and
`add_net_thread_message.py` — a 103-line code generator — is gone. And the screen migration is
not the half-finished state I assumed from ADR 003: `Screen::execute` is now implemented on top
of the nonblocking primitives rather than duplicating them, and direct blocking `execute(` /
`runModal()` call sites drop from 44 to one (`src/Engine.cpp:95`, the end screen).

`scons release=1 server=0 screen-test session-test` builds clean here, and
`ScreenExecutionHarness` plus `test/run-engine-session-test.py` give 12 PASS and exit 0.

## Non-blocking notes

- `libgag/src/GraphicContextCompound.cpp:85` adds
  `if (!surface->textureInfo) glState.setTexture(surface->texture);`, but the `else` branch
  eight lines down binds the same texture, and `GLState::setTexture`
  (`libgag/src/GraphicContext.cpp:80`) early-returns when unchanged. Harmless, just redundant
  on the native path.
- `CooperativeTask::promise_type` allocates a `shared_ptr<State>` in `get_return_object()` that
  `Awaiter::await_suspend` immediately overwrites with the parent's for every child task. One
  wasted allocation per subtask.
- The cmd+Q / alt+F4 handling moved from `SDL_GetModState()` to `event.key.keysym.mod`. That
  reads intentional — synthetic and replayed events need the event's own modifier state — but
  it is a macOS/Windows behaviour change that no CI job covers.
- The new `web` job (Emscripten, `npm ci`, Playwright across chromium + firefox + webkit,
  pulseaudio, 60-minute timeout) is a real recurring CI cost and a new flake surface. Worth
  deciding on deliberately rather than inheriting.

Thank you for the ADRs and for `status.md`. Being able to read your reasoning for screen
execution, cooperative loading and generation randomness before reading the diff is what made a
349-file PR reviewable at all.

*Opus 5 helped authoring this review.*

genixpro:
Addressed the requested changes in the rebased five-commit series:

- added an optional `wss=0` native build that omits OpenSSL and Boost.Beast and fails closed for WSS addresses; verified the TCP-only link contains neither `ssl` nor `crypto`;
- added the Debian OpenSSL build dependency and macOS packaging guidance;
- hoisted the fertility kernel’s per-tile coordinate and eligibility work while preserving equivalence in the engine-session harness;
- rebased onto current `master` and regenerated the deterministic replay fixture at format 93 because master now rejects older replay formats. The complete-import and truncated-stream regression passes against the new fixture.

CI is running on `81f91fb5e`.

genixpro:
On item 3 (the regenerated determinism baselines):

`tests/baselines/cross-replay.replay`/`.checksums` changed because `ReplayReader::open` hard-rejects anything below `REPLAY_MINIMUM_VERSION_MINOR`, and that floor has been climbing on master independent of this branch — every merge that changes simulation behavior bumps it:

- 90 — weighted pathfinding/diagonal timing (#211)
- 92 — hiring-bucket iteration order (#225)
- 93 — trapped-colony elimination (#180)
- 94 — fetch-job apportionment (#223, master's current tip)

The fixture predates all of that, so it fails the floor check outright once rebased past any of those merges. That's not cosmetic: `browser/tests/import.spec.js` and `session-reload.spec.js` both import these exact bytes and assert a successful import, so an unloadable fixture is a real, in-repo test failure, not just an external mismatch. There's precedent for this exact regeneration too — `e2f4da9e1` did the same thing in July when the format moved 84→87.

Separately: this branch is still pinned at `VERSION_MINOR 93`, one bump behind master's actual tip (94, from #223, which landed after my last rebase). So this fixture needs one more re-recording regardless, once I rebase past that.

On the sibling cross-replay tooling: that's @kylelutze's Rust reader, not something in this repo, so I can't regenerate or verify its oracle myself. Kyle — could you either grant visibility into it so this can be checked directly in future PRs, or regenerate/confirm its side against the baseline once this branch lands on the new bytes? Happy to ping you again once the next rebase (to version 94) is up.

I don't think this last piece should gate the merge decision here — the in-repo tests that actually load this fixture are covered, and the Rust-side sync is tracked as a follow-up rather than something this PR can resolve on its own — but I wanted it on record rather than left implicit.

genixpro:
Rebased onto master's current tip (`e3a01b051`, picking up #223's `VERSION_MINOR` 93→94 bump among others). The branch was further behind than I'd realized — 5 master commits, not 1, including the Settings (#236) and custom-game lobby (#237) redesigns, which had independently rewritten `CustomGameScreen` and `SettingsScreen` around the same time this branch converted them to the non-blocking `ScreenStack` architecture. Kept master's redesigned implementations and re-applied this branch's architecture on top of them (constructor now takes `ScreenStack&`; `SettingsScreen::persist()`/`onTimer` now also track the browser's `ApplicationHost::persistStorage()` flush, so a failed durable write still surfaces through the existing `saveFailed()`/retry path instead of being silently dropped).

Verified locally: native client and server both build clean (`scons release=1 server=0` / `server=1`), a headless `--nox` run completes with a deterministic checksum, and `data/check_translations.py --strict` is clean across all 34 language files.

On item 3: `tests/baselines/cross-replay.replay` is re-recorded at the new floor (version 94), from the unchanged `games/cross-replay.game`, run to its natural end. `cross-replay.checksums` came out byte-identical to the prior recording, consistent with the simulation being unchanged in the first 100 ticks the sidecar covers.

CI is running on `750ea6d37`.

genixpro:
@Giszmo — this should be ready for another look now. Summary of everything since your review:

**Your 4 requested changes:**
1. OpenSSL behind a flag — `wss=1` (default on), `wss=0` builds TCP-only without OpenSSL/Boost.Beast
2. `debian/control` — added `libssl-dev`
3. Determinism baselines explained — see the earlier reply on this thread; `cross-replay.replay`/`.checksums` track `REPLAY_MINIMUM_VERSION_MINOR`, which master keeps bumping independently of this branch. Re-recorded again since then at version 94 (was 93), because master's tip (#223) bumped the floor again after my last rebase.
4. `FertilityCalculator` kernel loop — hoisted the per-tile coordinate/eligibility work out of the 961-iteration inner loop

**Also since your review**, the branch turned out to be 5 master commits behind (not the 1 I first assumed), including the Settings (#236) and custom-game lobby (#237) redesigns landing around the same time — full rebase onto master's current tip is up now, with `CustomGameScreen`/`SettingsScreen` reconciled to keep your redesigns and re-apply this branch's `ScreenStack` architecture on top. Also fixed a stale CI path, a decade-old literal merge marker sitting in `texts.br.txt` on master itself, and translated 39 new UI strings across all 32 non-English catalogs that a newer master-only test now checks.

Native client and server both build clean locally, headless run is deterministic, and all four translation checks pass. CI is running now on `7662aec13` — should be visible on the PR by the time you look.

Giszmo (commented):
My judgement is colored by wanting this to be a thing on globulation2.org so I don't give approval but I hope @stephanemagnenat does ;)

genixpro:
## Browser follow-ups: torus overview, random maps, software map preview, WebGL2 by default

Four commits after the presentation fixes, deployed to the test bucket as `index-350ee6bdb7de5639`.

**Torus overview in WebGL2** (`3d4a0152a`). Pressing G in a WebGL2 match now shows the torus overview, which was compiled out of every Emscripten build. It is a viewing mode only; simulation is unchanged. WebGL builds save and restore the GL state each torus block changes instead of using attribute stacks, use 16-bit mesh indices (Emscripten's legacy-GL bridge draws emulated client-array elements as `GL_UNSIGNED_SHORT`), request a depth buffer, and recreate the overview's GL objects after context loss. The software renderer keeps the flat map. The native attribute-stack path was not built or run.

**Random maps launch again** (`0733a7182`). The ScreenStack destroys `CustomGameScreen`, whose destructor deleted the generated preview file, before `GameLoadScreen` reads the map, so every random-map launch ended on "the file is either damaged or an incompatible version". Native builds of this branch share the flow; master's blocking flow keeps the screen alive. The launch now holds the generated map until loading finishes. `CustomGameSetupHarness` loaded inside the completion callback, which is why it missed this; it now loads after the stack destroys the screen and checks the file is removed afterwards.

**Software map preview** (`4d71872d9`). The software renderer's scaled `DrawableSurface::drawSurface` overload was an empty TODO, so the lobby's stretched map thumbnail never drew and players saw a flat panel. It now uses `SDL_BlitScaled`. The only other software caller is GUIBase's blocking `execute()` background; the stretched magic effect stays GPU-only.

**WebGL2 is the default when the browser accelerates it** (`4a626e150`). The shell probes a scratch canvas and uses WebGL2 unless the browser reports a major performance caveat or a CPU rasterizer (SwiftShader, llvmpipe, softpipe, Software Rasterizer, Basic Render Driver); Chrome does not report SwiftShader as a caveat, hence the name check. `?renderer=webgl2` and `?renderer=software` still force either renderer.

### Feel and performance impact, for maintainer review

Most players now get WebGL2 presentation, including map zoom and the torus overview. Local measurement: Apple M3, headless Chrome 152 with Metal, FourSquares1 custom match, 10 s samples, 3 trials, median shown. Frames are paced to 25 fps, so CPU is the comparison.

| Window | Main thread per frame (software → WebGL2) | Match CPU | Menu CPU |
|---|---|---|---|
| 1280×720 | 5.5 → 3.6 ms | 21% → 27% | 24% → 24% |
| 1920×1080 | 9.0 → 4.3 ms | 30% → 37% | 36% → 25% |
| 2560×1440 | 13.0 → 5.3 ms | 42% → 46% | 47% → 25% |
| 3840×2160 | 23.8 → 12.0 ms | 72% → 83% | 84% → 27% |

Both renderers held 25 fps and 25 ticks/s at every size. WebGL2 roughly halves main-thread time and keeps menus cheap, but matches cost 3–11 points of one core more in total, because the legacy GL emulation does its work in Chrome's GPU process. With SwiftShader instead of a GPU, WebGL2 needed 3–4.3 cores and fell to 16 fps at 1920×1080 while software held 25 fps at under half a core, which is why emulated WebGL2 falls back.

CPU is summed across Chrome processes from CDP `SystemInfo.getProcessInfo`; main-thread time is `Performance.getMetrics` TaskDuration. The laptop was on battery, and these are averages without frame-time percentiles. `browser/benchmarks/rendering.cjs` still clicks pre-redesign menu positions and only reports ticks/s, which the frame pacing caps, so the numbers come from a separate local script.

### Verification

Headless Chromium unless noted:

- `single-player.spec.js` 16/16 with software rendering. The new random-map launch and map preview tests each fail on the previous build and pass now.
- `rendering.spec.js`, `viewport.spec.js` and `torus.spec.js` 17/17, including new default-selection tests: accelerated → WebGL2; unavailable or emulated → software.
- `runtime-build.spec.js` and `viewport.spec.js` 18/18 in Firefox and WebKit, both of which selected WebGL2 on macOS.
- Against the deployed bucket: random-map launch, map preview and torus toggle pass. Without a renderer parameter, Chrome with the M3 GPU, Firefox and WebKit chose WebGL2, and Chrome with SwiftShader chose software; both URL overrides were honoured.

Not verified: none of these changes were built or run natively (the changed C++ for the last two commits passes a macOS syntax check, and CI runs `CustomGameSetupHarness`). Real Safari and Edge, and Linux or Windows GPUs, were not tested. Headless Chromium in CI emulates WebGL2, so its default-renderer runs keep exercising software; which renderer CI's Firefox and WebKit runs select was not checked.

https://claude.ai/code/session_01Kem7kazWRvTVQXKfmvX5Tn

Giszmo:
This is even playable in the mobile browser somewhat. Main obstacles are half the screen being an always-on menu and the missing right mouse button.

PR #209, opened 2026-09-08
Refresh a resource-fetch target when its gradient is rebuilt stale

Giszmo:
When a worker's resource target disappears, the rebuilt resource gradient can route it elsewhere while targetX/targetY still refer to the depleted tile. Refresh the stored destination during resource movement when that tile is no longer marked GRADIENT_AT_GOAL.

The check preserves an existing goal. It does not attempt to retarget every still-valid resource whenever another route becomes cheaper; round-trip gradient behavior belongs to #210/#193.

The regression invokes the real movement method and resource-gradient rebuild directly across all seven swim classes. It verifies that a valid target stays unchanged, then removes the near resource, rebuilds the gradient and checks that the target switches to the remaining resource. This is a focused method regression, not a full-match fixture.

The branch incorporates current master, and Linux CI now builds and runs the regression through the shared disposable-profile runner, including preference-preservation checks:

```sh
scons -j8 release=1 server=0 resource-fetch-target-test
python3 test/run-savegame-safety-tests.py --check-preferences build/src/ResourceFetchTargetHarness .
```

Verification on d3f2dfe89388b9f0f4931445b509bbb4bc479810, integrated with master 8babf83c5:

- Unfixed master plus the regression fails at the intended stale-target assertion.
- macOS release build and all seven fixed movement-method cases pass, with preference bytes/mtime unchanged.
- A real G2 match completes 15,000 ticks. This is a smoke check, not a claim of exhaustive gameplay coverage or before/after replay equivalence.
- Ubuntu 22.04, Ubuntu 24.04 and Windows CI all pass on the final head. Both Linux jobs execute the new regression; Windows checks the client, save compatibility and server build.

 Giszmo's production fix is retained; follow-up work adds CI/profile isolation, swim-class coverage and a shorter explanatory comment. Sonnet assisted the original implementation; Codex assisted the integration and verification.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,6 +86,11 @@ jobs:
           scons -j$(nproc) release=1 server=0 immobile-unit-gradient-test
           ./build/src/ImmobileUnitGradientHarness
 
+      - name: Build and run the resource-fetch target regression
+        run: |
+          scons -j$(nproc) release=1 server=0 resource-fetch-target-test
+          python3 test/run-savegame-safety-tests.py --check-preferences build/src/ResourceFetchTargetHarness .
+
       - name: Test savegame loading and atomic autosaves
         run: |
           scons -j$(nproc) release=1 server=0 savegame-safety-test
--- a/src/SConscript
+++ b/src/SConscript
@@ -465,7 +465,9 @@ source_files.append("../libusl/src/libusl.a")
 p1 = local.Program("glob2", source_files)
 
 if not env['server']:
-    stats_sources = [source for source in source_files if source != 'Glob2.cpp']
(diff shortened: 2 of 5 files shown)


genixpro (approved):
Approved on d3f2dfe89388b9f0f4931445b509bbb4bc479810. The focused fix refreshes a stored resource target only once it ceases to be a gradient goal. Verified the regression fails on unfixed master, passes all seven swim classes with the fix, and preserves disposable-profile preferences. A release G2 simulation completes 15,000 ticks. Both Linux CI jobs pass with the new regression wired in. Waiting for Windows CI to pass before merging. Follow-up edits added test/CI integration and shortened the comment; the author’s production behavior is retained.

PR #211, opened 2026-09-08
Bump map, network and replay versions for the pathfinding change

Giszmo:
Follow-up to #184, prompted by @stephanemagnenat's review comment there.

#184 changed no saved byte (gradients are never written to disk; saves and maps from before it load fine), but it changed the simulation: units path and time their steps differently. Any recorded order stream from before it therefore diverges from what happened, and nothing refused those streams:

- `ReplayReader` accepted replays from format 86 up to the current version, so a pre-#184 replay played back with the new pathfinder. With checksums it aborts on the first order ("checksums don't match"); without them it drifts silently.
- The network compared only `NET_PROTOCOL_VERSION` (28), so a client built before the merge could join a game hosted by one built after and desync from the first tick.

This PR makes the incompatibility explicit:

- `VERSION_MINOR` 89 → 90, documented in `Version.h` as the version that marks the pathfinding change. (#169 bumped to 89 for the team statistics thirteen hours after #184 merged; no release sits between the two, but a dedicated number keeps the meaning of each bump clear.)
- `NET_PROTOCOL_VERSION` 28 → 29 and `YOG_MIN_CLIENT_NET_PROTOCOL_VERSION` 27 → 29, so mixed clients are refused before the game starts instead of desyncing in it.
- `REPLAY_MINIMUM_VERSION_MINOR` 86 → 90. With that floor the Uint16 step-counter path for pre-87 replays is dead, so it goes: `wideStepCounter`, `readStepCounter()` and `REPLAY_UINT32_STEP_COUNTER_VERSION_MINOR` are removed and the reader reads Uint32 directly. `ReplayStepCounterTest` loses its version-86 case and keeps the round-trip, floor/ceiling and decode-version-plumbing cases; the last one no longer requires the floor to be below the current version, since right after a floor bump they are equal.

Tested: release build; `ReplayStepCounterTest` and the other `test/` harnesses; the CI test targets.

### DIFF
--- a/src/ReplayReader.cpp
+++ b/src/ReplayReader.cpp
@@ -21,7 +21,6 @@ ReplayReader::ReplayReader()
 	ordersProcessed = 0;
 	numOrders = 0;
 	stepsUntilNextOrder = -1;
-	wideStepCounter = false;
 	versionMinor = VERSION_MINOR;
 	checksum = 0;
 }
@@ -87,9 +86,6 @@ bool ReplayReader::loadReplay(GAGCore::InputStream *inputStream, bool skipToOrde
 		return false;
 	}
 
-	// Replays written before version 87 store step counters as Uint16
-	wideStepCounter = (version_minor >= REPLAY_UINT32_STEP_COUNTER_VERSION_MINOR);
-
 	versionMinor = version_minor;
 
 	// If there are no orders, this is also not a valid replay (there should be at least a NullOrder)
@@ -108,7 +104,7 @@ bool ReplayReader::loadReplay(GAGCore::InputStream *inputStream, bool skipToOrde
 	std::shared_ptr<Order> order;
 	numSteps = 0;
 	numOrders = 0;
-	stepsUntilNextOrder = readStepCounter();
+	stepsUntilNextOrder = stream->readUint32("replayStepCounter");
 	do
 	{
 		try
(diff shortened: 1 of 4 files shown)


genixpro:
Good catch, and thanks for following this up. We should have checked the replay/network compatibility gates when landing #184: unchanged saved bytes do not mean unchanged simulation behavior.

This is also a concrete example for the shared repository guide in #173. I have added an explicit reminder: save-format changes must bump VERSION_MINOR with version-gated loading for older saves, or an explicitly approved/documented compatibility break. Simulation changes that invalidate old replays or mixed-client games must also update replay acceptance and network/YOG version gates, with boundary tests. An agent may usually infer this, but a few repository-specific lines should make it more consistent and prevent omissions.

I support addressing the compatibility gap. This comment is not yet approval of the exact implementation or its old-replay removal; those still need review and verification before merge.

genixpro (approved):
Approved following Bradley’s acceptance of the old-replay compatibility break. Reviewed the replay floor/ceiling checks and protocol constants, including YOG login and registration rejection paths. Rebuilt and ran ReplayStepCounterTest (17 assertions) and NetSendOrderDecodeTest (8 assertions) locally on macOS: all passed. Linux 22.04, Linux 24.04 and Windows CI are green for 26fb8387e17cddaffb9ff52d552b058c1809efed. The minimum save/map version is unchanged. The decode-version plumbing assertions cannot distinguish the header version from a hardcoded current version while the replay floor equals the current version; the production header-version plumbing remains intact.

PR #223, opened 2026-09-09
Apportion fetch jobs, and price a loaded candidate instead of refusing it

Giszmo:
A building wanting several resources at once sent every fetcher to whichever one was nearest, kept hiring past what it could hold, and refused a loaded candidate outright however close it stood.

Two commits, straight against master.

## 1. Apportion the jobs

`Building::considerUnitForResources` scored each candidate resource by `neededResource()`, derived from `resources[]` alone, which does not fall when a unit subscribes to fetch one. `Team::updateAllBuildingTasks` hires in a `while(cont)` loop that calls `subscribeToBringResourcesStep()` repeatedly within a single tick, so every pass re-read the same unchanged number. The score is `distance / need`, so the nearest resource won every pass.

Reported from a market built next to trees: **eight** wood fetchers for its four wood, **none** for its four stone. The surplus carriers then arrived at a full site, could not deposit, were re-hired for the stone still holding their wood, and the next harvest overwrote it.

`computeWishedResources` is subscription-aware, but it only sizes the worker count; it never picks which resource a given worker fetches.

The resource is now chosen by apportionment. Sainte-Lague priority `target / (2 * served + 1)` picks the resource whose served share sits furthest below its target share, with `served` counting subscribed units alongside deliveries that landed. Distance then chooses which unit fills that job.

- targets of 4 and 4 interleave one for one
- targets of 4 and 1 give wood, wood, algue, wood, wood

## 2. Price a loaded candidate instead of refusing it

Empty-handed candidates were a *tier*: a unit carrying something the building cannot take was only considered once that tier came up empty, so it lost to an empty unit anywhere on the map and the job went to the longer trip.

They are now scored together, with the loaded unit charged `CARRIED_RESOURCE_PENALTY_TILES = 5` of detour. Enough that an empty-handed unit a short walk further away still wins — which leaves the cargo with its carrier long enough for a building that *does* want it to hire the unit first — and small enough that a unit far enough ahead is still the better hire.

Merging the passes also picks up a case that fell between them: a unit carrying a resource this building wants but has already fully subscribed. It is now hireable for a different resource, at the penalty.

`bringResourcesLevel` still dominates the comparison, as before; the penalty only orders candidates of equal level.

## Tests

- `FetchApportionmentTest` (in `TestsRunner`) covers the ranking, including that hiring stops exactly at the targets.
- `FetchApportionmentHarness`, real-engine, three cases:

| case | result | without the change |
|---|---|---|
| market site, wood 5 tiles away, stone 21 | 4 wood, 4 stone | **8 wood, 0 stone** |
| two candidates placed symmetrically, both scoring 19 | the empty-handed one is hired, whichever of the two holds the cargo | same (this is the guard, not the discriminator) |
| loaded candidate at 12, empty at 18 | the loaded one is hired | the empty one at 18 takes the job |

Verified: `TestsRunner` (177 tests), `WinningConditionsHarness`, and every engine harness CI runs — selection, building-footprint, entering-unit-save, immobile-unit-gradient, resource-fetch-target, terrain, global-gradient, savegame-safety, buffered-file, team-stats-save, including the version88 and version84 expected-stdout fixtures. `scons server=1` builds. A 47,298-tick headless AI game runs to a clean win.

## Not included

`.github/workflows/build.yml` needs a step for the new `fetch-apportionment-test` target, mirroring `resource-fetch-target-test`:

```yaml
      - name: Fetch apportionment regression
        run: |
          scons -j$(nproc) release=1 server=0 fetch-apportionment-test
          python3 test/run-savegame-safety-tests.py --check-preferences build/src/FetchApportionmentHarness .
```

Pushing that file needs a token with the `workflow` scope, which this one lacks — same limitation #209 hit.

## Compatibility

`VERSION_MINOR` and `REPLAY_MINIMUM_VERSION_MINOR` move to 91. #225 also takes 91; the two are independent, so whichever merges second needs a one-line bump to 92.


Giszmo:
**Hold this PR.** Measured through a strained-economy scenario (Playground, 8 Nicowars, warmed to tick 25000, all builds resumed from the same save and run to tick 40000, six seeds) this branch is *worse* than master:

| | median deliveries | median starvation deaths |
|---|---|---|
| master | 2656 | 22 |
| this branch | 2562 (-3.5%) | **54** |

Starvation deaths are worse in **6 of 6 seeds** (76/39, 31/13, 47/25, 62/14, 41/19, 67/45). That column is not noise.

Likely cause: `WISHED_RESOURCE_NUM/DEN` is 4/3, so a building deliberately hires about a third more carriers than it has room for. The apportionment here caps hiring at exactly the deliveries outstanding, which removes that headroom. The surplus was absorbing losses — units that die, get stuck, or arrive after a slot is filled — and for an inn an unfilled corn slot is a unit that starves.

The reported bug (8 wood fetchers for 4 wood while stone got nobody) is over-subscription of *one resource*, not of the building. Those are separable and this PR conflated them: the apportionment should decide **which** resource the next carrier fetches and leave the 4/3 headroom on the total intact.

Reworking on that basis and re-measuring before this is worth reviewing.

Giszmo:
**Hold lifted.** I isolated this branch from the task-switch/poaching work that was stacked on it in my measurement build, and measured it alone across six seeds of the strained-economy scenario (Playground, 8 Nicowars, warmed to tick 25000, all builds resumed from the same save, run to tick 40000):

| | median deliveries | median starvation deaths |
|---|---|---|
| master | 2656 | 22 |
| this branch alone | 2674 (+0.7%) | **14** |
| this branch + #225 + poaching | 2562 (−3.5%) | 54 |

The apportionment on its own is delivery-neutral and cuts starvation deaths by about a third, better in 4 of 6 seeds. The regression I flagged earlier comes entirely from the unmerged poaching work that was stacked on top, which reassigns units mid-job at a rate of ~54% of all hires. That work is not part of this PR and is being reworked separately.

My earlier comment blaming the `WISHED_RESOURCE_NUM/DEN` headroom was wrong, and I retract it. `desiredMaxUnitWorking` is already `min(maxUnitWorking, WISHED_RESOURCE_NUM * neededResourcesSum / WISHED_RESOURCE_DEN)` and is checked before hiring, so the per-resource targets in this PR never governed the total. Restoring the 4/3 inflation on them produced byte-identical results in all six seeds.

genixpro (approved):
Read the two-commit diff. The Sainte-Lague apportionment in `FetchApportionment.h` is a clean, integer-only, order-independent ranking (unit-tested directly, including the tie-breaking and satisfied-resource-dropped cases), and `selectFetcher`'s fixed detour for a loaded candidate is a sensible replacement for the old hard tier split. Your isolated A/B evidence (delivery-neutral, ~1/3 fewer starvation deaths across 6 seeds once separated from the unmerged poaching work) is convincing. `FetchApportionmentHarness` reproduces the original 8-wood/0-stone bug and the two loaded-vs-empty tie-break cases. Version bump to 91 is correctly documented. CI green. Merging.

PR #224, opened 2026-09-09
Fix the two distance defects in the fetch-candidate scorer

Giszmo:
Targets `feat/pathfinding-lanes` (#193) — both defects are in c5250839, so they belong here, not on `feat/pathfinding-swaps` which only inherits them.

Two commits, two separate defects in `considerUnitForResources`.

## 1. The hunger check measured the wrong thing

It refuses a unit that cannot reach a resource before it goes hungry, and took that distance as:

```cpp
distResource = roundTrip - distBuilding;
```

which is `(unit→resource) + (resource→building) − (unit→building)`. Not a length between two places, and it can come out negative.

- Unit **at the building**: `distBuilding ≈ 0`, so the value is about `2 × (unit→resource)` and the gate is twice as strict as intended. Units right next to a building were refused resources they could comfortably reach.
- Unit **beside the resource, far from the building**: the value collapses toward zero and the gate stops rejecting anything.

Traced instance: `roundTrip=2 distBuilding=2 distResource=0` for a resource one tile away.

Fixed by asking the resource gradient for the walk and the round-trip field for the score, instead of deriving one from the other. The round trip still ranks candidates, so the intent of c5250839 is untouched.

## 2. The two arms were on different scales

With a field, `roundTrip` is a real round trip. Without one, the fallback was `distBuilding + distResource` — the walk out plus however far the unit happens to be from the building, which for a unit standing at the building is *half* a round trip. Which arm a resource gets turns on whether anything has fetched it for this building yet, so a resource already being fetched was ranked against a fresh one on two different scales.

Fixed by estimating the carry leg rather than substituting the building distance:

```cpp
roundTrip = distResource + std::max(distBuilding, distResource);
```

| unit is | true round trip | old `dB+dR` | this |
|---|---|---|---|
| at the building | `2·dR` | `dR` — half | `2·dR` ✅ |
| at the resource | `dB` | `dB` ✅ | `dB` ✅ |
| far from both, they are close together | `~D` | `2D` | `2D` |

Exact at both ends, no worse in between. **No gradient is built** — this reads only the two numbers already in hand. Building a field here would cost one per resource of every hiring building, nearly all never fetched, which is exactly what the lazy construction exists to avoid.

## Tests

`RoundTripHungerGateHarness`, new (`round-trip-hunger-gate-test`), two cases, each of which fails on `feat/pathfinding-lanes` unchanged:

| case | fixed | unfixed |
|---|---|---|
| unit with time to reach the wood but not to carry it home is hireable | hired | `FAIL: a unit that can reach the wood is hireable for it` |
| no field built, unit 2 tiles from the site and 12 from the wood | scores 1536, the round trip of 24 over a need of 4 | scores 896, the one-way 14 |

Also on this branch: `TestsRunner` (172 tests), `WinningConditionsHarness`, and every engine harness it has — selection, building-footprint, entering-unit-save, immobile-unit-gradient, terrain, global-gradient, savegame-safety, team-stats-save. `scons server=1` builds. A 46,850-tick headless AI game runs to a clean win.

## Not included

`.github/workflows/build.yml` needs a step for the new target; pushing that file needs a token with the `workflow` scope, which this one lacks — same limitation #209 hit.

```yaml
      - name: Round-trip hunger gate regression
        run: |
          scons -j$(nproc) release=1 server=0 round-trip-hunger-gate-test
          python3 test/run-savegame-safety-tests.py --check-preferences build/src/RoundTripHungerGateHarness .
```


Giszmo:
Folded straight into feat/pathfinding-lanes as 900f8a5 — no more feature-branch-on-feature-branch, and it drops a stray .local/state/gh/device-id that git add -A had swept in. Closing this in favour of #193.

PR #226, opened 2026-09-09
Price a loaded fetch candidate instead of refusing it

Giszmo:
Stacked on #223 — review that one first. Targets `fix/proportional-fetch-jobs`; retarget to `master` once #223 lands.

## The problem

Hiring consulted empty-handed candidates as a *tier*, and only looked at units carrying something the building cannot take once that tier came up empty. A loaded unit standing at the door therefore lost to an empty one anywhere on the map, however far away, and the job went to the longer trip.

## The change

Score them together and charge the loaded unit a fixed detour:

```cpp
/// Tiles of detour a candidate pays for turning up with a resource this
/// building cannot take.
constexpr int CARRIED_RESOURCE_PENALTY_TILES = 5;
```

Five tiles is enough that an empty-handed unit a short walk further away still wins — which leaves the cargo with its carrier long enough for a building that *does* want it to hire the unit first, which is the point — and small enough that a unit far enough ahead is still the better hire. A price, not a veto.

Merging the two passes also picks up a case that fell between them: a unit carrying a resource this building wants but has *already fully subscribed*. The delivery pass had no room for another of that resource, and the fetch-out pass only looked at units carrying something unneeded, so nothing could hire it. It is now hireable for a different resource, at the penalty.

**Note on levels.** `bringResourcesLevel` still dominates the comparison — a higher-level candidate beats a lower-level one regardless of distance or penalty. That is pre-existing and I have not touched it; the penalty only orders candidates of equal level. Say the word if you want the level tier priced too.

## Tests

Two new cases in `FetchApportionmentHarness`:

| case | result |
|---|---|
| two candidates placed symmetrically about the site, both scoring 19 | the empty-handed one is hired, **whichever of the two holds the cargo** — so neither position nor scan order decides it |
| loaded candidate at 12, empty at 18 | the loaded one is hired despite the penalty |

The second is the discriminating one: on `fix/proportional-fetch-jobs` without this change it reports

```
lopsided: empty=18 loaded=12 hired=0
FAIL: the loaded unit is hired anyway
```

The first passes either way by design — it is the guard against over-correcting, not the discriminator.

Also verified: `TestsRunner` (177 tests), `WinningConditionsHarness`, every engine harness CI runs, `scons server=1`, and a 47,298-tick headless AI game to a clean win.

## Not covered here

Your idea of dropping the cargo at a nearby building that *does* want it, when the detour is small, is a separate and larger change — it needs a cheap "who wants this near here" lookup and a detour budget, and it lives in the walking code rather than the hiring code. Not in this PR.

## Version numbering

This takes `VERSION_MINOR` 92. #225 also takes 92; they are independent, so whichever merges second needs a one-line bump to 93.

### DIFF
--- a/src/ReplayReader.h
+++ b/src/ReplayReader.h
@@ -26,7 +26,7 @@ static constexpr Uint32 REPLAY_MIN_VALID_ORDERS = 5;
 //! changed the simulation (fetch-job apportionment), as version 90 did before
 //! it (weighted pathfinding, diagonal timing), so earlier replays would diverge
 //! from what happened.
-static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 91;
+static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 92;
 
 /// This class is used for reading replays.
 /// The replay stream is kept open and read every time you do retrieveOrder.
--- a/src/Version.h
+++ b/src/Version.h
@@ -6,7 +6,7 @@
 // This is the version of map and savegame format, and all of the recorded data on the server
 #define VERSION_MAJOR 0
 #define MINIMUM_VERSION_MINOR 58
-#define VERSION_MINOR 91
+#define VERSION_MINOR 92
 // version 10 adds script saved in game
 // version 11 the gamesfiles do saves which building has been seen under fog of war.
 // version 12 saves map name into SessionGame instead of BaseMap.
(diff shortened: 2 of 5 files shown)


Giszmo:
Folded into #223 instead of stacking a PR on a PR. Same commit, collapsed onto VERSION_MINOR 91 since both changes now ship together.

PR #239, opened 2026-09-10
Fix conversion abort after a blocked occupant's building is destroyed

genixpro:
Destroying a building while an occupant has finished service but cannot find an exit released the unit into random activity without a map slot. A hungry worker could then select an enemy inn and abort on the map-occupancy assertion during conversion.

Treat `DIS_EXITING_BUILDING` with `MOV_INSIDE` as still indoors when the building dies, using the existing occupant-death policy. Units that successfully exited remain alive. Conversion assertions remain intact, and no other unit's map slot is overwritten.

Validation:
- Added a real-engine regression covering blocked and successful exits for ground and air units, next-step collection, and preservation of other air units' occupancy; wired it into Linux CI.
- The regression fails against the original destruction code and passes with the fix on Linux. A clean build from current master plus this change also passes the new regression on macOS, plus the adjacent entering-unit save/load regression (40 directions/positions and corruption controls).
- Replayed the original failing 2v2 scenario (FourSquares1, seed 805328015) using an isolated binary with only this translation unit changed: original abort at tick 57950; corrected run exits normally at tick 92802 under a 100000-tick cap.

Tournament queues and the frozen engine were not changed. Diagnostic outcomes are not tournament observations. This PR contains only the core fix and its regression/build integration.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,6 +122,11 @@ jobs:
           ./build/src/EnteringUnitSaveHarness
           ./build/src/EnteringUnitSaveHarness --load test/fixtures/entering-explorer/reproducer.game
 
+      - name: Build and run the building occupant death regression
+        run: |
+          scons -j$(nproc) release=1 server=0 building-occupant-death-test
+          ./build/src/BuildingOccupantDeathHarness
+
       - name: Build and run the entering unit draw regression
         run: |
           scons -j$(nproc) release=1 server=0 entering-unit-draw-test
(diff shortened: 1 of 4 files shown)


genixpro:
Closing as superseded: the fix here (`u->isDead=true` for a blocked-exiting occupant, i.e. `DIS_EXITING_BUILDING` + `MOV_INSIDE`) is now in direct conflict with #185/#186 ("Let units survive building destruction"), which merged into master afterward and takes the opposite position — it expels that exact unit alive instead of killing it.

Master's own `BuildingExpelHarness.cpp` (`destroyedInnExpelsEveryone`) explicitly constructs a unit in this exact state and asserts it survives, gets expelled, and ends up registered on the map. So the underlying bug this PR set out to fix — a blocked-exiting occupant released into random activity with no map slot, causing a map-occupancy assertion during conversion — is already fixed upstream via a broader mechanism (expel-with-fallback in `Building::kill()`), not the narrower kill-it fix this branch took.

Rebasing this as originally written would regress already-merged, already-tested behavior, so closing rather than carrying it forward.

PR #238, opened 2026-09-10
Unify map generator controls and tune playable defaults

genixpro:
Random-map defaults now leave more grassy building room while retaining each generator's shape. The lobby, map editor, command-line stress tool and study tooling share one definition of generator names, available controls, ranges, steps and defaults.

**Stacked on #237** (`codex/custom-game-ux`, 82943d96e). The diff contains only the generator follow-up.

### Changes

- Tune all eight procedural presets. Against #237's lobby settings, immediately free tiles rise from 9.8% to 40.7% for River and 9.2% to 57.1% for Crater Lakes.
- Add measured lake-size, channel-width, neutral-island, island-size and bridge-width controls. Both UIs use the same definitions, remember terrain settings per mode, and share dimensions/colony/worker settings. Grass 75 and island size 65 are selectable; bounds and five-unit steps agree everywhere.
- Remove ineffective wheat/wood/stone/algae ratio widgets; retain working fruit controls. Restore legacy resource placement and fix the invalid team index during island base placement.
- Rename the legacy modes to **Shattered Coast** and **Rugged Archipelago**, with matching labels in all 33 languages.
- Export the compiled catalog to generate the control reference and chart labels. Add reproducible study seeding without changing normal UI random seeding.

### Validation

- Release game build, shared-control regression, and existing custom-setup headless/native SDL tests pass, including previews, controllers, saves and replays.
- Every selectable editor value, lobby dropdown/stepper interactions, per-mode memory, and descriptor serialization checked. Add shared-control coverage to Linux CI.
- **24,000 fixed-seed attempts**: 1,000 per generator for tuned, previous lobby, and previous editor settings. **72 independently repeated seed/configuration pairs** match. Inspect 24 final map previews plus native UI captures.
- Strict translation audit, English-fallback regression, Python syntax and diff checks pass.

Both baseline cohorts include the same legacy correctness fixes. Placement failures remain: River has 18/1,000 (versus lobby 4 and editor 145), and Shattered Coast 14/1,000 (versus lobby 0 and editor 2), alongside substantially better building room and starting-resource access. Small crowded maps remain a limitation; this is not a full-match balance study.

Coverage comparison

Generated reports, study data and screenshot files are excluded from the branch tip and final source changes. Existing image embeds reference earlier review commits; future artifacts stay outside tracked source.

PR #237, opened 2026-09-09
Redesign custom-game lobby with automatic maps, teams, and rules

genixpro:
Custom-game setup now keeps map selection, players/teams, and rules in a native tabbed lobby. Players edit controllers and alliances directly in colored colony rows, tune rules inline, and see random maps generated automatically before launching the exact previewed world.

### Behavior

- Use **Econo** in implementation classes, source filenames, CLI help and all language catalogs. Display **Econo - Easy - No warriors** in the selector and profile header. Preserve numeric AI ID 4 and binary save layout. Source files, class, enum, CLI spelling and text section names all use Econo; historical binary fixtures retain their original recorded data.

- Automatically remember the lobby configuration in an atomically written, versioned user settings file: map/library choice, generator controls, all colony/AI/team assignments, rules and expanded sections. Reload on return or app restart. Random mode generates a fresh preview; missing maps preserve the draft, and malformed preferences fall back to defaults.
- Map / Players & Teams / Game Rules, with persistent summaries and a gold launch action. Four-player FFA defaults to FourSquares1, you and three Numbi AIs. Compact layouts support 640×480; larger rosters and settings scroll independently.
- Premade and user libraries use canonical map identities, deduplicate symlink/search-path aliases, and preserve selection/scroll position. Distinct same-name maps show their source folder. Windows/shared-folder installations use one combined library so shipped maps stay visible.
- Random previews generate automatically after edits settle for 500 ms. Terrain, resources, dimensions, landscape and layout controls stay inline; dragging defers generation. Owned temporary snapshots launch through the existing map-loading path, with cleanup and retry handling.
- Explicit You / AI / You + AI / Closed controllers, FFA / 2 vs 2 / You vs all presets, retained hidden assignments, and a validated twelve-controller limit. AI choices show difficulty and provide browsable strategy/counterplay profiles.
- Standard, Quick clash, Open book and Last colony standing rulesets expose victory, visibility, alliances, pace and generated workers. Premade starting units remain authored; map-script victory still applies; session speed preferences are restored.
- Live AI-only watching supports whole-map/colony viewpoints, pause, navigation and inspection while blocking gameplay orders. Every AI, including the local bookkeeping slot, is polled once per tick. Shared-control and AI-only saves retain existing encodings; saves/replays remain self-contained after temporary-map cleanup.
[...]

PR #240, opened 2026-09-10
Map generation improvements: modular framework and twelve new generators

genixpro:
Map generation was spread across central dispatch code, overloaded descriptor fields, global random state and duplicated helpers, which made generators hard to add, compare or trust. This PR makes map generation modular, adds twelve new generators, makes starts measurably fairer, gives every generator tunable resource amounts and on/off switches, and gives the custom game lobby better control over random maps.

## What changes

**Framework.** Each generator is a registered module with named controls, a revision, and optional request and world validation. The lobby, map editor, compatibility adapter, study tools and tests all read the same definitions. `GenerationService` validates the request, generates, validates the finished world, scores it, and keeps the best of five seeded rolls. Randomness comes only from named per-generation streams. See `docs/map-generators/`.

**Catalog.** 20 playable generators plus editor-only Uniform. New:
- **Contested commons:** home islands around a large shared island.
- **Maze:** a pen-and-paper maze of grass passages, stone-and-water walls and sand roads, with every colony in its own cul-de-sac.
- **Fjord continent:** fjords between colonies, a central lake that can connect to them, and outlying resource islands.
- **Watershed:** a branching river network into a delta, with fertile floodplains, dry uplands and sand fords.
- **Stone highlands:** grass valleys walled by stone ridges, with passes and ponds.
- **Symmetric arena:** exactly symmetric starts for 2, 4 or 8 colonies around a moated orchard.
- **Ring world:** one continent belt wrapping the torus, so every colony has exactly two neighbours.
- **City states:** a large shared commons in the middle ringed by a strait, and round it one big walled home per colony, each joined to the commons by a single stone-lined causeway. Every roll differs in its coasts, its home layout (lake, creek and ford, ridges and passes, marsh, sand flanks) and the heart of its commons (lake, crag, island, delta, forest), and every home is the same wedge turned round the centre, so any colony count is fair by rotation. Played and approved over four iterations.
- **Tidal flats:** grass islands on a wide expanse of walkable sand that holds no building and no deposit, so open-field battles from the first minute and no forward bases. Every home island has a pond, fields and a quarry; every neutral island is an oasis ringed with wheat out to its shore that has to be cleared to be taken; an island at the centre carries the orchard. Fair for any colony count by rotation. Played and approved over three iterations.
- **Everglades:** a dense wetland, the same everywhere and across the wrap: pools and sloughs on a jittered lattice, grass between them thick with wood and wheat that grow back at the engine's top rate and close the lanes unless kept cut, sand beaches as the lanes that stay open, and every colony in a levee-ringed clearing with a pond and a kit. The generator bends rather than refuses (clearings shrink, the ring widens, routes are opened to any boxed-in colony), so no size and colony count fails.
- **Spider web:** an orb web of land spun over open water: spokes from a hub with an orchard out to a frame thread, capture threads sagging between them, and every colony on a pad where its spoke meets the frame. Every ground route follows the threads, so knots are the places to hold and the water between is crossed only once colonies can swim; dew-drop islets carry prizes. The capture threads are a spiral with one arm per colony, so the whole web is fair by rotation. Approved from rendered results.
- **Coral:** every colony is a sea fan of land: a trunk from its pad on the rim forks level after level towards the middle, so each colony's coral is a triangle opening inwards and neighbouring fans reach into each other's gaps. Land gets thinner and richer away from home (stone at forks, fruit on tips). The fan is grown once and turned round the map, and every branch keeps a strait from every colony's copy, so it is fair by rotation. Approved from rendered results and early play.

Lattice was added and then removed within this branch, and never shipped.

**Resource amounts and switches.** Every playable generator now exposes the resources it places as amount controls, and its own sub-behaviours as on/off switches: 56 amount controls and 22 switches across 14 generators.
- An **amount** is a percentage of that generator's own default, 0 to 300 in steps of 25, default 100. It scales the numbers that already decided that amount, so at 100 every map is exactly what it was.
- **Switches** turn one behaviour on or off while leaving the map playable either way — Fjord's lake-to-fjord connection, Isles' land bridges and sandy beaches, Maze's sand roads and dead-end treasure, River's winding course, stone on hilltops for the height-field generators, and so on.
- **Fairness placements are never scaled** — starter kits, the 1:1 guaranteed wheat and wood, Fjord's per-bank guarantees and Ring world's starter patches — so an amount of 0 empties the ambient layer but still leaves every colony a start.
- High amounts used to wall a colony in with nowhere left to build, so generation now reopens a cramped start, and only at non-default amounts.

**Fairer starts and resources.**
- **Choosing starts:** height-field generators score candidate sites as they will be built.
- **Scoring:** every finished map's colonies are scored on fertility, wheat and wood reach, room, resource depth and isolation, and that score picks the best of five rolls.
- **Reachability backstop:** `guaranteeStartingResources` clears resource walls that seal a colony off from wheat or wood, and tops up whatever is still missing. Designed walls can be protected from it.
- **Scatter:** ambient resources follow fertility, and algae now actually places, by water body.
- **Fertility overlay:** generated maps get a populated fertility overlay.

**Lobby.**
- **Controls:** random maps expose each landscape's controls, with amounts and switches alongside them.
- **Checkboxes:** every on/off option is a check button in both the lobby and the map editor, reachable by mouse or keyboard.
- **Randomize** under the preview rolls the same settings with a new seed.
- **Reset to defaults** restores the landscape's registered values.
- **Default size:** random maps now default to 256×256.
- **Catalog order:** Contested commons and Maze lead; the four newest generators sit mid-list.
- **Landscape picker:** the Landscape field opens a full-window modal (`LandscapePickerScreen`) that shows every landscape as a freshly generated map at the draft's current size and colony count, with colony markers and a Regenerate all button. Previews roll on background threads (`LandscapePreviewer`, at most four). The map you pick is the map you play: the lobby rolls the seed behind the tile you chose instead of sampling five, until the next edit. The picker takes only names and requests, so the editor and multiplayer lobby can reuse it.

**Fixes found along the way.**
- **Queue overrun:** `computeDistances` overran its queue on repeated sources, which made Isles depend on leftover heap memory. Isles moves to revision 2, and its 128×128 seed 5 changes.
- **Algae scatter:** algae scatter never placed anything.
- **Rectangular maps:** previews and wrapped start coordinates were wrong.
- **Old random:** starts had a wood-poor bias.
- **Resource walls:** colonies could be walled off from wheat or wood by resources.
- **Lobby settings:** a Fjord lake size of 50 reloaded as 30, and a Fjord lake size of 0 or 90, a Ring world lake density of 0, or an Old growth lake size above 100 made the whole settings file fail to load.
- **Contested commons past four colonies:** its whole-map colony spread cost map area × colonies² per pass behind a fixed budget of 20 million evaluations, so it failed on every seed past four colonies at 256 and at any colony count at 512. The search now scores each tile from its two nearest weighted distances, kept current as points move: the same minimums in the same scan order, so every seed that generated before gives a byte-identical map (147 configurations checked), while 8 colonies at 256 now rolls in about 0.2 s and 12 at 512 in about 1 s. Revision stays at 2.
- **Concurrent game construction:** libusl's static per-type native prototype is repointed at every `Game`'s heap on construction; two games built at once corrupted each other's heaps. That update is now under a mutex.

**Translations.** Every new label and message is in all 33 text tables — English plus 32 translations. Words a catalog's own language spells as English does (Polder, Coral, Spiral, Plazas, Canals, Grain, Diagonal in the languages concerned) are listed as reviewed shared vocabulary. The Brazilian Portuguese catalog's entries for this branch were rewritten after an earlier pass had used Breton.

## Cleanup and safety net

A second pass over the generator tree, every step verified byte-identical against a golden table:
- **Golden maps and colony sweep:** `MapGeneratorGoldenTest` keeps a per-platform table of the map every generator produces for fixed seeds, sizes and colony counts, keyed by revision; a map that changes at an unchanged revision fails CI. `--sweep` rolls every playable landscape at every colony count the lobby offers and fails any cell where no seed generates. Its first run found Shattered coast unable to seat 12 colonies on most seeds; that placement now relaxes its spacing on a crowded map. macOS arm64 and Linux x86-64 rows are committed, and CI runs the check with `--require-rows`, so a platform CI builds on cannot pass it without rows; `--print` runs first so a new platform's rows are in its log.
- **One style:** a `.clang-format` in the house style (tabs, Allman, 100 columns) and the whole tree reformatted, with no map changed. The height map's raw arrays, debug dumps, noise macros and dead diagnostics are retired.
- **A shared terrain layer:** `shared/Grid` holds the torus arithmetic, the eight-connected flood, the walkable mask, units by team and the colony-reachability check that seven generators and their validators each carried a copy of; 530 lines net removed.
- **Structure:** the five functions over 200 lines (Contested commons, Rugged archipelago, Isles, Shattered coast, City states) are split into named stages around small layout structs, with direction tables replacing hand-written switch cases where the draw order allowed it.
- **Lobby responsiveness:** the preview's five candidate rolls run on the landscape previewer's workers and only the winning seed is rolled on the UI thread, so a 512x512, 12-colony draft no longer freezes the lobby.
- **Measured, not done:** caching random-stream lookups by name gained one to two percent and was reverted; the legacy preferences adapter stays as a reader, since old files must still load.

## Fairness in real play

`tools/map_fairness_tournament.py` plays Nicowar free-for-alls on one generated map across many engine seeds and asks whether the same starting positions keep winning, which is a different question from the structural score `scoreStarts` reports. Per-start win counts get an exact multinomial or chi-square test, Wilson intervals, and Benjamini-Hochberg correction across generators; `GLOB2_TEAM_RESULTS` reports per-team elimination order, so a game that reaches the tick cap still yields a ranking instead of being discarded. Presets live in `tools/map-fairness/`, the method is written up in `docs/map-generators/FAIRNESS_TOURNAMENT.md`, and `test/test_map_fairness_tournament.py` covers the statistics offline.

What a partial run of 309 games across all 15 generators found:
- **The win rate follows the map's colony index, not the engine's team index.** Fit to the same games, a slot-only model holds (p=0.91) while a team-only model is rejected (p=0.005).
- **Colonies 2 and 3 win about twice as often as colonies 0 and 1** - 16.5 / 18.4 / 32.4 / 32.7 % pooled over all 15 generators. It holds among decisive games alone (16.5 / 19.2 / 34.4 / 29.9), so tick-cap adjudication is not what produces it.
- **No individual generator departs significantly from that shared pattern.** Measured against the pooled profile rather than a uniform quarter, the largest departures are Fjord continent and Old random, and neither survives correction across 15 generators. Per map, the harness reports no biased map after Benjamini-Hochberg.
- **Earlier reports of an engine team-index bias are withdrawn.** A 96-game smoke run put team 1 at 38.5% (6.5 pp, p=0.034), a second 96-game run did not replicate it (p=0.43), and over 309 games team 1 is the weakest index.
- The **start-quality scorer does not predict wins**, so `scoreStarts` is best read as a structural check rather than a fairness guarantee. It is also not rotation-invariant: `build_sites` counts by top-left anchor.

That run covered 2 of the 4 rotations and about 20 games per generator, so the pooled result is solid but per-generator conclusions are underpowered. The remaining games are not part of this PR.

Two full-rotation runs on this head (`tools/map_fairness_tournament.py`, Nicowar in every slot, 128×128, 4 colonies, best of 5 lobby rolls per seed, tick cap 90,000; every rotation of team numbers over starts is played, so wins by start carry no team-index effect and wins by team index carry no map effect):

- **`baseline`** (Symmetric arena only, six maps, 96 games, 69 decided by elimination, 2 of 2 re-run games reproduced exactly): position bias 0.0 pp against a fair-map floor of 8.7, no map biased after Benjamini-Hochberg, wins by colony index 26 / 15 / 29 / 26 (p 0.17). Wins by team index 25 / 18 / 18 / 35 (p 0.053), team 3 at 36.5% [27.5%, 46.4%]: the engine's processing-order effect sits at the edge of detectability on a map with nothing else in it, and is the thing to keep measuring.
- **`smoke`** (Symmetric arena, Contested commons, Watershed, Crater lakes; three maps each, 96 games, 82 decided): the same numbers as the 2026-09-12 run in `FAIRNESS_TOURNAMENT.md`, game for game. Team index pooled over all four generators 28 / 21 / 23 / 24 (p 0.80). Symmetric arena shows no start bias (0.0 pp, floor 14.4). The three asymmetric generators do, per map rather than per colony index: Contested commons 35 pp with 3 of 3 maps biased, Watershed 33 pp with 2 of 3, Crater lakes 29 pp with 2 of 3, and the best start on those maps wins 2.5 to 3.5 times its fair share whichever team plays it. The start scorer only partly predicts which (rho 0.59 on Contested commons, 0.68 on Crater lakes, none on Watershed).

The `standard` preset (every generator, 2,160 games, about 5.5 hours at 3 jobs) is not part of this PR; its generator list predates the landscapes and it belongs with the per-generator tuning that follows this merge. Reports and CSVs for both runs are under `artifacts/map-fairness/` locally and available on request.

## Compatibility

- **Saves and replays** contain full worlds, so generator changes don't alter existing files.
- **Generator ids:** legacy numeric ids are unchanged. The new ids are 9 and 11–21; 10 was used only by Lattice inside this branch and stays retired.
- **Lobby settings:** existing `custom-game-settings.txt` files still load, bridged through the legacy descriptor; the file gains an optional options section for the new controls. Round trip, older files and corrupt-file recovery are tested.
- **Determinism:** generation is deterministic for a given seed, revision and platform. Cross-platform identical generation is not a promise of this framework.
- **Fertility rewrite:** `Fertility::Field` also backs the editor's fertility tool and old-save migration; its effect on migrated saves has not been separately checked.
- **Per-thread random stream:** the generator behind `syncRand()` is now `thread_local`, so the picker's background previews neither disturb nor race the menu's live colony. The simulation runs on one thread and its sequence is unchanged: headless `--nox` replays from the previous and the new client binary are byte-identical over 1200 ticks (macOS arm64).

## Verification

Run on the head before City states (`82e01557f`), macOS arm64, release build, and repeated for the contract tests and translations on the final head (`7b221ff64`):
- **Build:** full `scons` build.
- **CI:** linux (ubuntu:22.04), linux (ubuntu:24.04) and windows (mingw-w64) all pass. `near` and `far` are macros in Windows' `windef.h`, and two generators had used them as identifiers, which kept the mingw job red; both are renamed, verified as a pure rename.
- **`MapGeneratorDefaultsTest`:** all four groups pass. That covers registry and control contracts, seed repeatability including after an intervening map, RNG isolation, legacy sentinels, algae placement, and lobby/editor control memory.
- **`CustomGameSetupHarness`:** passes headless, and windowed at 640×480 and 1000×700 with all five groups green at each size, including clicking Randomize, Reset to defaults and the new checkbox rows through the production controls.
- **`TestsRunner`:** 198/198.
- **Offline tournament statistics:** `test/test_map_fairness_tournament.py` passes.
- **City states:** 999/1000 single rolls over ten configurations (2 to 12 colonies, 128 to 512 maps) pass generation and its own world validation; the one failure is a 12-colony 512 roll whose landing walks spread past the bound, which the lobby's best-of-five absorbs. Rendered rolls, the sweep and fairness measurements are on a page linked from the review thread.
- **Tidal flats:** 1000/1000 single rolls over ten configurations (2 to 12 colonies, 128 to 512 maps) pass generation and its own world validation.
- **Translations:** `data/check_translations.py --strict` reports 0 structural errors; `test/test_translations.py`, `test/test_font_coverage.py` and `test/test_text_area_layout.py` pass.
- **Generation sweeps with `MapGeneratorStudy`:**
  - all 15 generators at the 256×256 default, 10 seeds each, with 0 failures;
  - the four newest generators at 50 default seeds each, with 0 failures, plus every control's endpoints, colony counts, and sizes from 64 to 512 including rectangles;
  - every new amount's extremes and every switch's other state generate cleanly.
- **Defaults unchanged:** 45 map dumps (15 generators × 3 seeds) are byte-identical across both merges into this branch and across the Windows rename, so no generator revision moves beyond the Isles change noted above.
- **Shared-code changes** were checked against before/after map dumps:
  - the protected-wall mask, algae fix, catalog reorder and overrun fix leave all sampled maps byte-identical, except the Isles seed noted above;
  - Symmetric arena maps were independently checked for exact rotational symmetry.

- **Cleanup pass** (`57e424cde` to `e2071ba27`): golden check 126 rows with 0 failures after every step, sweep 0 failing cells, `MapGeneratorDefaultsTest` and `CustomGameSetupHarness` (headless, 640x480, 1000x700, SDL-driven flow, preferences modes) green on the final head; the full-rotation `smoke` tournament reproduced 2 of 2 re-run games exactly.
- **Landscape picker and Contested commons fix** (`adf81b19b`, `517ea16d9`): `CustomGameSetupHarness` passes headless and windowed at 640×480 and 1000×700 (three runs each), with new coverage of the picker's previews on background threads, selection by click and keys, regeneration, and the lobby rolling the very seed the picker showed; a ThreadSanitizer build of the visual harness reports only a pre-existing `VoiceRecorder` teardown race; `MapGeneratorDefaultsTest` gains a regression rolling Contested commons at 256/8, 512/4 and 512/12 that fails on the previous code and passes now.

Screenshots and example maps for every generator are in [this comment](https://github.com/Globulation2/glob2/pull/240#issuecomment-5640670249). Seed studies were generated locally under ignored `artifacts/` and are available on request.

- **Spider web and its primitives** (`745098856`): new shared `Drawing` (strokes, Bezier threads, turned shape fills), `KitFrame`, `plantFields`, `scatterClumps` and `secureStartingCrops`, with Tidal flats, City states, Everglades and Ring world moved onto them byte-identically (133 existing golden rows unchanged); golden 140 rows with 0 failures, `--sweep` 0 failing combinations, a stress run of 1,662 / 1,742 rolls over 64–512 maps, 1–12 colonies and every control extreme (the misses are the up-front refusal of 6+ colonies on 64×64), median start fairness 0.965; `MapGeneratorDefaultsTest` with new toolkit checks and `check_translations.py --strict` pass.
- **Coral, rectangular maps and Spider web tuning** (`e95f98865`): new shared `growBranches`, `pathClearance`, `tracePath`, `Stretch`, `sprinkleSand` and `algaeGrowthChance`; golden 168 rows (a 512×256 row per generator added) with 0 failures and every other generator's square maps byte-identical; `--sweep` 0 failing combinations; stress runs of 1,642 / 1,862 Coral rolls (median start fairness 0.966, every miss an up-front refusal) and 1,842 / 1,922 Spider web rolls (median 0.960) over 64–512 maps, 1–12 colonies and every control extreme; rectangle stress median fairness 0.96 for both; `MapGeneratorDefaultsTest`, `check_translations.py --strict` and clang-format pass.

- **Review response (`ab95800c2`, 2026-09-14):** the mingw `near` clash, the `test/` link surface, the Stone Highlands pass sorting and GCC `-Warray-bounds` noise, `Glob2.cpp` style, 47 untranslated keys and the Old growth preferences bound, with master (#194, #269) merged in and the CORN → WHEAT rename applied branch-wide. Locally on macOS arm64: full build, `MapGeneratorDefaultsTest`, golden 248 rows with 0 failures (with and without `--require-rows`), `--sweep` 0 failing combinations over 32 generators, `CustomGameSetupHarness` in all three modes, `MenuColonyHarness`, `test/TestsRunner` and every `test/` harness, and all four translation checks. CI: [run 34844458453](https://github.com/Globulation2/glob2/actions/runs/34844458453) green on all three jobs, including the savegame harness on Windows, the `test/` suite on Linux and the golden check against committed rows.

**Not run:** cross-platform per-tick checksum comparison; maintainer playtesting of each generator; native-speaker review of the new translations, which are machine-assisted.

## Play-note batch (2026-09-14)

Seventeen tweaks from playtesting the branch, each explained in the generator's header comment and in the docs:

- **Tidal flats:** six oases and eight sandbars per wedge by default (were three), smaller, rounder and with tighter gaps so crowded wedges seat more; the smallest oasis keeps its pond inside a ring of wheat. A 128 map with six colonies is the limit and fits only an oasis or two.
- **Carousel:** the sea inside the ring stays a lagoon with algae, the inward farm and the home kit are gone; lanes and courts on the lagoon keep a 5-tile stone band (a diagonal spoke on the old 3-tile margin lost every grass tile to its beach and seal, which cut colonies off from the plaza).
- **City states:** an archipelago of round islets in the sea outside the design circle, where the torus wraps (a square's corners, a rectangle's bands): one on the wrap point and rings of eight, sixteen and so on round it with the map's own four-fold symmetry, each islet radius 11 with a 10x4 building plot and a small prize, three tiles of water from everything; the Barrens home layout is gone.
- **Standard farms:** sand bridges run clean across the whole farm, crop rows and water alike (Old town, Carousel, Switchbacks).
- **Amphitheatre:** starting towers stand against the arena's outer wall by the colony's own ramp, covering the arena.
- **Starting tower level** defaults to 1 on Carousel, Amphitheatre, Switchbacks and Canals.
- **Switchbacks, Carousel, Polder:** no wheat and wood kit in the homes; Polder's wood share 8% (was 15%).
- **Rain shadow:** streams 7 deep, pass roads 18 tiles into each valley and joined to the streams' beaches.
- **Canals:** a `block-shape` choice of squares or hexagons.
- **Maze and Stone highlands** amounts are percentages like every other landscape; the resource-amount audit is in the framework doc.
- **Catalog order** shuffled once (no bias towards the landscapes written first); the lobby and editor open on the first entry, Fingerprint.
- **Lobby:** random maps are the default (the premade library preselects FourSquares1 when first opened); Reset to defaults moved to the top of the map column with Random parameters beside it, which redraws any set the generator or the world refuses (up to six times) until a map exists; the landscape picker gains Randomize parameters and Reset to defaults and hands the lobby the shown map's parameters as well as its seed; a generated map's fairness and score show under its preview with an (i) opening the per-colony breakdown. Reloading every saved setting, generator options included, is covered by the preferences harness modes.

Verification on macOS arm64 (release): full build; `MapGeneratorDefaultsTest`; golden 248 rows with 0 failures under `--require-rows`, with linux-x86_64 rows regenerated on a Linux x86-64 box (gcc); `--sweep` 0 failing combinations over 30 playable landscapes at 128 to 512 and 2 to 12 colonies; `CustomGameSetupHarness` headless with captures, `preferences-write` and `preferences-read`; `MenuColonyHarness check`; all four translation checks. Screenshots of the compact lobby with the new buttons and the fairness line, the breakdown screen, the randomized picker and example maps are on a results page linked from the review thread, and every one reproduces with `build/src/CustomGameSetupHarness artifacts/custom-game/compact` (captures) and `tools/render_map.py one <landscape> --size 256 --set teams=N`.

## Feel changes to playtest

- **256×256 default:** random maps start bigger, so games take longer and colonies spread out further.
- **Choosing a landscape:** the dropdown is gone; the Landscape field opens the picker, and picking a tile plays that exact map. Previews take a few seconds to fill in on a large map with many colonies.
- **Amounts and switches:** defaults reproduce today's maps exactly, but they open up deliberately poorer or richer maps, and switches like Isles' land bridges or Fjord's lake connection change a map's shape as much as its economy.
- **Watershed:** the economy follows the rivers, and before swimming pools first contact happens at the fords.
- **Stone highlands:** passes are the only ground routes all game, stone is effectively unlimited, and first contact is slow.
- **Symmetric arena:** the early game is calm, and the orchard in the centre is the map's only fruit.
- **Ring world:** every colony fights on two fronts, and 5 colonies on 128×128 is crowded.
- **Spider web:** ground routes are narrow threads, so knots and the hub decide early fights; swimming changes the map entirely. A sand road (on by default, like Maze's) runs down every spoke and whole thread, and threads now carry about twice the wheat and wood.
- **Coral:** the front between colonies is a zigzag of fingertips, a short swim apart; bases start on a pad at the rim with one trunk to hold, and supply lines thin at every fork. Sand roads run down every forking branch.
- **Rectangular maps:** Fjord continent, Tidal flats, Spider web and Coral now stretch to fill a rectangle as an oval instead of sitting in a disc on the short side; City states stays round, because stretching it made its homes unequal.
- **Maze and Fjord continent:** both changed substantially through playtesting on this branch.
- **Play-note batch (2026-09-14):** Carousel's middle is a lagoon, so its farms are all outside the ring and the plaza is reached only down the spokes; City states' outer sea is an archipelago of forward posts for whoever swims first; Tidal flats has far more oases and sandbars; starting towers everywhere are level 1, so players upgrade their own; three landscapes' homes start without a wheat and wood kit and rely on their fields and farms; the lobby opens on a random map.

## Known limits

- **Symmetric arena** accepts only 2, 4 or 8 colonies. `scoreStarts` counts build sites by top-left anchor, so its fairness score reads just under 1.
- **Coral** with 12 colonies joins its trunk tips into a small ring of bridges around the centre.
- **Watershed** has one river network, so a 512×512 map keeps a desert strip far from the delta.
- **Untested AI behaviour:** how the AIs cope with Highlands passes and Watershed fords.
- **Some starts win more often than others, and this PR does not fix it.** A partial run suggested colonies 2 and 3 win twice as often as 0 and 1; a full-rotation run (96 games) shows no colony-index skew and no engine team-index bias, and start metrics by colony index are flat over 680 maps (`tools/colony_start_metrics.py`). What is real is per map: on seven of nine asymmetric maps one start won seven or eight of eight games whichever team played it, and the start scorer only partly predicts which. The fairness reference records the numbers; teaching the scorer what a dominant start has is the open work.
- **Amount and switch combinations** are covered by generation sweeps, not by playtesting. Extreme amounts are deliberately unbalanced.

Features and gameplay changes need a human maintainer's review and playtest: @Giszmo, @kylelutze.

https://claude.ai/code/session_014TBmePvnWuhjRHwGdT5UPq
https://claude.ai/code/session_01GpUW6Ny43aUA1EzThn6cYE


genixpro:
Rebased onto the (also just-rebased) #238 branch — it forked from the same `codex/custom-game-ux` point as #238, so I used the same `rebase --onto` approach to replay only this branch's own 6 commits.

Most conflicts were pure append/append (translation files) or a stale-vs-final doc mismatch, resolved by taking this branch's own final version of `docs/map-generators/README.md` directly rather than trusting the intermediate per-commit auto-merge (one commit in this branch's own history trims that doc down over several later commits; replaying it against my manual resolution of an earlier commit briefly duplicated content — using the actual final committed version sidesteps that entirely). `test/CustomGameSetupHarness.cpp` had one large genuine conflict (`model()`/`engine()`/`sessionReplay()`/`reload()`/`main()`); after stripping whitespace, master's and this branch's versions turned out to be functionally identical except for the map-generator loop itself, which I merged by keeping master's already-independently-added content (two new test methods from `#237`'s tail commits, extra `AINames` assertions) and splicing in this branch's `GeneratorRegistry`-based loop in place of the old hardcoded 8-method one.

**The real find, after conflicts resolved:** the branch didn't compile. `#238` (merged in the meantime) added `CustomGamePreferences.h`, which serializes `CustomGameSetup::generator` via direct `Sint32 MapGenerationDescriptor::*` member pointers. This branch replaced `generator`'s type with `GenerationRequest`, whose per-method options live in a generic `std::map<std::string,int>` — an architecture change, not a rename, so `#238`'s file had no path forward. Fixed by routing `encode()`/`decode()` through `toLegacyDescriptor()`/`fromLegacyDescriptor()` (the compatibility adapter this PR already built for exactly this), keeping the on-disk format and corruption-recovery bounds unchanged, and widening a few of those bounds to match the registry's current ranges (they'd gone stale — e.g. `oldIslandSize`'s own default already exceeded the field's declared max for any method other than Isles/Old Islands, so a completely ordinary preferences save could fail to load back).

While tracing that, I also found and fixed an unrelated crash in the compatibility adapter itself: Lattice and Maze register `wheat`/`wood`/`stone`/`algae` controls that `legacyField()`'s mapping table doesn't know about, so converting either method through the adapter threw an uncaught `std::invalid_argument`. Added the four mappings (they match pre-existing legacy struct fields exactly) and changed the fallback for every other option with no legacy slot (loopiness, home-radius, cell-size, ...) from throwing to using the control's default value, so a newer generator's full option set can't take the adapter down.

Verification on this head:
- Full `scons -j8 release=1 server=0` client build is clean.
- `CustomGameSetupHarness` passes in default, `preferences-write` and `preferences-read` modes.
- `MapGeneratorDefaultsTest` and `MapGeneratorStudy --catalog` pass, unaffected by the adapter fix.

Requesting review from @Giszmo and @kylelutze (same as #238, since this stacks directly on it). No merge performed.

genixpro:
Retargeted onto master and rebased now that #238 has merged. Master's tree was byte-identical to #238's old branch tip (clean squash merge), so `git rebase --onto origin/master origin/codex/map-generator-defaults` replayed this branch's own 7 commits with **zero conflicts** — no manual reconciliation needed this time.

That surfaced a real compile break in CI, though: `tools/MenuColonyHarness.cpp` still called `Map::oldMakeIslandsMap`/`Game::oldMakeIslandsMap`, which the registry-driven rewrite removed. Fixed by routing it through `MapGenerator::generateMap(game, d, seed)`. Two follow-on issues surfaced once it actually ran end-to-end (CI only exercises `check`/`navigation`, not `generate`, so these were latent):

- The shared `workers` control is now capped at 1–8; the harness's old `nbWorkers=48` fails validation. Clamped to 8 — verified the 12,000-tick AI warmup still grows the decorative colony to 56 units/11 buildings by the end, so this doesn't defeat the intent.
- `island-size` on Rugged Archipelago (the renamed Old Islands) is now 50–70; the old `oldIslandSize=35` fails too. Switched to `d.setMethodDefaults(...)` first and only overriding what the tool actually needs to differ, so it stops rotting every time a generator's tuned ranges move.

Verified on this head (`db88fb0c9`):
- Full `scons -j8 release=1 server=0` client build is clean.
- `scons -j8 release=1 server=0 menu-colony-harness` builds clean.
- `check`, `navigation`, and `generate` subcommands all pass.

CI is running now. Requesting review from @Giszmo and @kylelutze (unchanged from the original description). No merge performed.

genixpro:
## Screenshots and example maps for every generator

Each panel has two parts:
- **Lobby Map tab** for that generator, captured by `CustomGameSetupHarness` at 1000×700. The tab headers show "2 vs 2" and "Quick clash" left over from earlier harness steps, not the game's defaults.
- **Three example maps** at the default settings (256×256, 4 colonies, seeds 1–3) from `MapGeneratorStudy`, each with its start-quality score.

The maps are drawn tile for tile: grass, sand, shore, water, wheat, wood, stone, algae, fruit, and red squares for each colony's swarm. Everything was generated from `95e24a761`, in catalog order.

**Update:** the lobby halves of these panels predate the resource amounts and on/off switches. Refreshed Map tab screenshots for every generator, showing the amount sliders and check buttons, are in [this comment](https://github.com/Globulation2/glob2/pull/240#issuecomment-5641660824). The example maps below are unchanged: map defaults are byte-identical after that merge.

<details open><summary><b>Contested commons</b> (id 9): home islands around one large shared island, joined by bridges</summary>

01-contested-commons

</details>

<details><summary><b>Maze</b> (id 11): grass passages and stone-and-water walls, with every colony in its own cul-de-sac</summary>

02-maze

</details>

<details><summary><b>Fjord continent</b> (id 12): one continent with fjords between colonies and a central lake</summary>

03-fjord-continent

</details>

<details><summary><b>Old random / Shattered coast</b> (id 7): a fragmented patchwork of land and water</summary>

04-shattered-coast

</details>

<details><summary><b>Isles</b> (id 6): home islands linked by land bridges</summary>

05-isles

</details>

<details><summary><b>Watershed</b> (id 13): branching rivers into a delta, with fertile floodplains, dry uplands and sand fords</summary>

06-watershed

</details>

<details><summary><b>Stone highlands</b> (id 14): grass valleys walled by stone ridges, joined by passes, with ponds</summary>

07-stone-highlands

</details>

<details><summary><b>Symmetric arena</b> (id 15): exactly symmetric starts around a moated central orchard</summary>

08-symmetric-arena

</details>

<details><summary><b>Ring world</b> (id 16): one continental belt wrapping the map, so every colony has two neighbours</summary>

09-ring-world

</details>

<details><summary><b>Old islands / Rugged archipelago</b> (id 8): separate jagged islands</summary>

10-rugged-archipelago

</details>

<details><summary><b>Concrete islands</b> (id 5): land plates split by channels, with extra islands</summary>

11-concrete-islands

</details>

<details><summary><b>Crater lakes</b> (id 4): open land with round lakes</summary>

12-crater-lakes

</details>

<details><summary><b>Islands</b> (id 3): organic islands with no land passage between them</summary>

13-islands

</details>

<details><summary><b>Swamp</b> (id 1): water and land interleaved</summary>

14-swamp

</details>

<details><summary><b>River</b> (id 2): one winding river through connected land</summary>

15-river

</details>

genixpro:
## Lobby Map tab for every generator, with resource amounts and switches

Refreshed from `5b967beb9`, now that resource amounts and on/off switches are merged. Each shot is the lobby's Map tab at 256×256 with 4 colonies, captured by `CustomGameSetupHarness` in a 1000×1200 window so every control group fits in one frame: **Terrain**, the generator's **switches as check buttons**, **Resources** with its amount sliders at their default 100, **Layout**, and **Reset to defaults** — plus the preview and **Randomize**.

Two notes:
- The **example maps** in [the panel comment above](https://github.com/Globulation2/glob2/pull/240#issuecomment-5640670249) are still current. Defaults are byte-identical after the merge, so only the lobby half of those panels aged; the maps themselves are unchanged.
- The tab headers show "2 vs 2" and "Quick clash" left over from earlier harness steps, not the game's defaults.

Every amount sits at 100, which reproduces exactly the maps this PR already generated. In catalog order:

<details open><summary><b>Contested commons</b> (id 9): home islands around one large shared island, joined by bridges</summary>

01-contested-commons

</details>

<details><summary><b>Maze</b> (id 11): grass passages and stone-and-water walls, with every colony in its own cul-de-sac</summary>

02-maze

</details>

<details><summary><b>Fjord continent</b> (id 12): one continent with fjords between colonies and a central lake</summary>

03-fjord-continent

</details>

<details><summary><b>Old random / Shattered coast</b> (id 7): a fragmented patchwork of land and water</summary>

04-shattered-coast

</details>

<details><summary><b>Isles</b> (id 6): home islands linked by land bridges</summary>

05-isles

</details>

<details><summary><b>Watershed</b> (id 13): branching rivers into a delta, with fertile floodplains, dry uplands and sand fords</summary>

06-watershed

</details>

<details><summary><b>Stone highlands</b> (id 14): grass valleys walled by stone ridges, joined by passes, with ponds</summary>

07-stone-highlands

</details>

<details><summary><b>Symmetric arena</b> (id 15): exactly symmetric starts around a moated central orchard</summary>

08-symmetric-arena

</details>

<details><summary><b>Ring world</b> (id 16): one continental belt wrapping the map, so every colony has two neighbours</summary>

09-ring-world

</details>

<details><summary><b>Old islands / Rugged archipelago</b> (id 8): separate jagged islands</summary>

10-rugged-archipelago

</details>

<details><summary><b>Concrete islands</b> (id 5): land plates split by channels, with extra islands</summary>

11-concrete-islands

</details>

<details><summary><b>Crater lakes</b> (id 4): open land with round lakes</summary>

12-crater-lakes

</details>

<details><summary><b>Islands</b> (id 3): organic islands with no land passage between them</summary>

13-islands

</details>

<details><summary><b>Swamp</b> (id 1): water and land interleaved</summary>

14-swamp

</details>

<details><summary><b>River</b> (id 2): one winding river through connected land</summary>

15-river

</details>

https://claude.ai/code/session_014TBmePvnWuhjRHwGdT5UPq

Giszmo:
For these map generators we need people trying them out. Some might be completely biased or even unplayable most of the time. Even the one called symmetric is obviously not perfectly fair as the cardinal directions matter. Inns expand down and to the right for example.

That said, those maps look fun to try out. I'll only not have much time the coming week. Also I'd like to draw your attention to my #261 tiled map idea. Its main goal is to have perfectly fair maps. Toroidal maps are perfect for tiling.

genixpro:
City states, the eighth generator, is in as of 43426ae49. Rendered rolls for every home and heart layout, the 1,000-seed sweep and the fairness measurements are here: https://claude.ai/code/artifact/1dc2c167-e5df-4893-94bc-5755051767eb

genixpro:
Tidal flats, the ninth generator, is in as of 7b221ff64: grass islands on walkable sand, oasis islands ringed with wheat, a central orchard island; played and approved over three iterations.

Giszmo:
## Review at `b86c59b` (merge base `95c6a60`, 120 commits)

Built and ran this branch on Linux x86-64, GCC 15, `scons release=1`. Findings below are reproduced locally unless marked otherwise.

### Where the 41k lines actually are

| area | + | − | files |
|---|---:|---:|---:|
| `src/map/generator` — framework + 21 generators | ~23.5k | ~3.9k | 121 |
| `data/texts.*` — 34 catalogs | 8,226 | 0 | 34 |
| `test/` | 4,516 | 470 | 13 |
| `tools/` — fairness tournament + plots | 1,599 | 23 | 9 |
| `docs/map-generators` | 1,382 | 6 | 8 |
| `src/` other | 1,326 | 745 | 25 |

Within `src/map/generator`: 14.4k is the generator implementations, 6.6k the shared toolkit, 1.2k the legacy ports, 786 the core framework. So the headline is roughly 14k of new map-drawing code, 10k of framework and tests around it, and 17k of translations, docs, tooling and UI.

Worth stating plainly because it affects who should review what: this is not only a map-generator change. It also lands `LandscapePickerScreen` with a threaded `LandscapePreviewer`, rewires `CustomGameScreen` / `NewMapScreen` / `EditorMainMenu`, and changes the engine's global sync RNG.

---

## Blocking

### 1. CI is red on all three jobs

[Run 34764038562](https://github.com/Globulation2/glob2/actions/runs/34764038562).

**Linux 22.04 and 24.04** die in the *first* step, `Validate translations`, at ~35s — 7 English fallbacks:

```
[Spiral]  in dk, id, sv, tr
[Coral]   in es, pt, ro
```

Reproduced: `python3 test/test_translations.py` → `FAILED (failures=7)`.

**Windows mingw-w64** fails to compile [`Tessellation.cpp:340`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/shared/Tessellation.cpp#L340):

```cpp
std::vector<std::vector<std::pair<int, int>>> near(obstacles.size());
```

`near` is a legacy `windef.h` macro that expands to nothing, so the uses at [L354-L355](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/shared/Tessellation.cpp#L354) and [L387](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/shared/Tessellation.cpp#L387) become `expected '{' before '.' token`. This PR already documents the trap at [`StoneHighlandsGenerator.cpp:749`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/generators/StoneHighlandsGenerator.cpp#L748) — "Not near/far: both are legacy macros in Windows' windef.h". One rename fixes it.

### 2. The `test/` unit-test build no longer links

`MapHomogen.cpp` was removed and its body now lives in [`src/map/MapTerrain.cpp`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/MapTerrain.cpp#L238), which pulls `globalContainer` into a translation unit that `test/SConstruct` deliberately links against the stripped `libgagserver.a` (see the comment above `explored_area_objs`, "the link surface is Map.cpp's own dependencies"):

```
MapTerrain-saveload.o: undefined reference to `globalContainer'
                       undefined reference to `ResourcesTypes::get(unsigned int) const'
                       undefined reference to `Map::isResourceAllowed(int, int, int)'
collect2: error: ld returned 1 exit status
scons: *** [MapExploredAreaSaveLoadTest] Error 1
```

Checked both sides: `cd test && scons -j5 MapExploredAreaSaveLoadTest` succeeds at `95c6a60` and fails at `b86c59b`. CI has not surfaced this because the Linux jobs die at the translation gate (step 1) long before `Build and run the tests`. Fixing the catalogs will expose a second red.

Cleanest fix is probably to keep those three functions in their own translation unit rather than in `MapTerrain.cpp`.

### 3. The new golden gate does not gate the platform CI runs on

`test/map-generator-golden.txt` has 168 rows, every one tagged `macos-arm64`. [`check()`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/test/MapGeneratorGoldenTest.cpp#L191) returns 0 when no row matches the running platform. Running exactly what CI runs, on Linux:

```
$ ./build/src/MapGeneratorGoldenTest glob2-golden-ci
No golden rows for linux-x86_64; nothing to compare. Run --update on this platform and commit its rows.
$ echo $?
0
```

So "a generator whose output changes without its revision moving fails the check" holds on the author's Mac and nowhere else. `--update` on a Linux runner and committing those rows would make it real.

To be fair to the same CI step: `--sweep` **is** a genuine gate, and it passes — see below.

---

## Worth fixing before merge

### 4. `nearbyValleys` skips its sort on the early return

[`StoneHighlandsGenerator.cpp:287-302`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/generators/StoneHighlandsGenerator.cpp#L287) returns at `count == limit` *before* `std::sort(ids, ids + count)`:

```cpp
if (count == limit)
    return count;          // unsorted
ids[count++] = v;
...
std::sort(ids, ids + count);
```

That path is reached whenever a 4th distinct valley is in the window. The caller at [L755-L757](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/generators/StoneHighlandsGenerator.cpp#L755) then builds `anyPair` keys from unsorted ids, so `{a,b}` and `{b,a}` become two separate map entries and the fallback pass-carving sees a split, smaller candidate tile list than intended. It is deterministic, so no desync — but it makes `"passes could not join every valley"` fire more often than it should.

### 5. Six new `-Warray-bounds` warnings

All from [`StoneHighlandsGenerator.cpp:751`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/map/generator/generators/StoneHighlandsGenerator.cpp#L751) — GCC 15 loses the `count <= 3` bound across the `std::sort` call and warns about `nearIds`/`farIds`. A false positive, but it is new noise under `-Wall`.

### 6. `src/Glob2.cpp` style, in a PR that also adds `.clang-format`

Two `#include`s sit [above the SPDX/copyright header](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/Glob2.cpp#L1), and [L180](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/Glob2.cpp#L180) / [L197](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/Glob2.cpp#L197) are space-indented in a tab-indented file.

---

## Design calls worth a conscious decision, not defects

**7. `randomGenerator` is now `thread_local`** ([`Utilities.cpp:29`](https://github.com/Globulation2/glob2/blob/b86c59bb97015458a6990c875d4daec494333011/src/Utilities.cpp#L29)). This is a change to the engine's lockstep sync RNG, made so the threaded previewer can generate without disturbing the menu's live colony. It is correct only as long as the simulation never runs on more than one thread, and that invariant is currently carried by a comment. Flagging it because it is the one change here that could bite far from map generation.

Everything else on the determinism side checks out: no `rand()`, no `std::uniform_*_distribution` (not portable across standard libraries), an own FNV-1a seed derivation and own rejection-sampling `bounded()`, and `GenerationService` saves and restores the engine stream around generation.

**8. Generator IDs 7 and 8 are reused.** `shattered-coast` and `rugged-archipelago` take the numeric IDs *and* the translation keys of Old random and Old islands. A stored lobby preference pointing at method 7 now silently selects a different landscape. Documented in `MAP_GENERATOR_FRAMEWORK.md`, so I read it as deliberate — noting only that it is a substitution rather than a rename.

**9. Preferences are forward- but not backward-compatible.** `CustomGamePreferences` gains an `options` section; a file written by this build is rejected by an older binary, which then falls back to defaults. No crash, just worth knowing.

---

## What holds up

- **The save and wire format really is untouched.** I diffed `getData` / `setData` / `save` / `load` / `checkSum` against master: identical byte layout. The one substantive change, `resource[i] << (3 * i)` → `<< ((3 * i) % 32)` (UB for `i >= 11` with `MAX_NB_RESOURCES == 15`), is inert, because those five functions have no callers on master either.
- `scons release=1` builds clean, zero errors.
- `MapGeneratorDefaultsTest` passes all five groups.
- `MapGeneratorGoldenTest --sweep` passes: **0 failing combinations** across 20 playable generators × 10 size/colony cells. Two details from the table:
  - `concrete-islands` is marginal at high colony counts — `256/12: 4/5`, `512/12: 2/3`. The lobby keeps the best of five rolls, so 12 colonies there is one unlucky seed from a visible failure. Every other generator is 5/5 wherever it accepts the request.
  - `symmetric-arena` refuses 3, 6 and 12 colonies outright and `ring-world` refuses 12. Those are `invalid`, not failures, so by design — but it implies the lobby must gate those counts per generator, which I did not verify.
- The framework shape is good: a registry of `GeneratorDefinition`s with revisions, named per-generator RNG streams, request/world validation hooks, and a compatibility adapter that keeps the old descriptor alive for the callers that still want it.

---

## Recommendation

Not mergeable as it stands, but none of the four blockers is deep: the translation catalogs, the `near` rename, moving `makeHomogenMap`/`controlSand`/`smoothResources` out of the stripped test link surface, and committing `linux-x86_64` golden rows.

I would not ask for this to be split. The framework switch is essentially one atomic change and splitting it would mostly move risk around.

The thing that is genuinely unanswered is playtesting. This PR ships a fairness tournament tool and a `Symmetric arena` baseline config, but no committed tournament *results*. `--sweep` proves a map generates and seats its colonies; it says nothing about whether a map is fair or fun. Baseline and standard tournament output in the PR body would close that gap.

<sub>Reviewed with Opus 5.</sub>

Giszmo:
Trusting the clanker on code quality and pending fixed unit testing I ACK this change but also wonder who will test and play these maps. It's probably hundreds of hours of ~work~ play before we know what's worth keeping and what needs tweaking and there problem with that is that we let this feature ripe with users that we don't know if they will ever report back issues with the different models.

genixpro:
## Review response, at `ab95800c2`

Thanks for the thorough pass. Every item is addressed on this head; master (#194, #269, #181, #278) is merged in as well, with the CORN → WHEAT rename applied across this branch. What changed, in your numbering:

### Blocking

1. **CI red.** The translation gate was the nine landscapes and three arenas registering 47 strings that no catalog had: every one is translated in all 32 catalogs (850ce2cec), aligned with each catalog's existing terms for ridges, passes, homes and resources. Where a catalog's own word is spelled as the English is (Polder in the Germanic and most Slavic catalogs, Coral in es/pt/ro, Spiral in dk/id/sv/tr, Plazas in es, Canals in ca, Grain in fr, Diagonal in de/id) it is listed as shared vocabulary rather than reworded. Seven keys no generator refers to any more are removed. While in there: `texts.br.txt` is Brazilian Portuguese and an earlier pass had written about fifty of this branch's entries in Breton; they are rewritten. The mingw break is the `near` vector in `Tessellation.cpp`, renamed `nearby` (953fe431c); once that compiled, `Territories.h` stopped on `const int far`, so every remaining `near`/`far` identifier in the generators and their test headers is renamed too (1617f7ecb).
2. **`test/` link.** `makeHomogenMap`, `controlSand` and `smoothResources` are back in their own translation unit, `src/map/generator/MapHomogen.cpp`, listed with the generator sources; `MapTerrain.cpp` no longer pulls `globalContainer` into the stripped link surface. `cd test && scons MapExploredAreaSaveLoadTest` builds and passes again, and the whole `test/` suite (`TestsRunner` plus every harness) is green locally.
3. **Golden gate.** `test/map-generator-golden.txt` now carries `linux-x86_64` rows, taken from this CI run's `--print` output, and the check gained `--require-rows`: the same comparison, but a platform with no rows fails instead of reporting and passing. CI passes that flag, so the Linux jobs cannot pass this step vacuously any more; the plain check keeps passing on a rowless machine so a new platform can still be bootstrapped. The workflow prints the rows before it checks, so the next platform's rows are in the log even when the check fails. The ubuntu 22.04 (g++ 11) and 24.04 (g++ 13) jobs printed identical rows, so one platform tag covers both; 156 of the 248 rows match macOS hash for hash and the other 92 are the generators that go through floating point (Fjord continent, Watershed at 512, Ring world and others), which is the per-platform caveat the table was designed around.

### Worth fixing

4. **`nearbyValleys`.** The helper keeps a sorted three-slot set and always returns it sorted, so a valley pair keys `anyPair` in one order. Stone highlands moves to revision 3. The golden rows for the sampled seeds are unchanged, so the fix only bites where the fallback ran.
5. **`-Warray-bounds`.** The same rewrite removes the `std::sort` over a runtime-bounded raw array; there is no runtime count for GCC to lose track of any more.
6. **`Glob2.cpp`.** The two includes are in the client include block under the header, and the three space-indented lines are tabs.

### Design calls

7. **`thread_local` sync RNG.** This one bit, though not where either of us looked. Once the Windows job compiled, `SavegameSafetyHarness` failed there on its check that a save leaves the sync RNG untouched, with the same harness and save code master passes on the same toolchain (GCC 16.2, Boost 1.92, MSYS2). The one RNG change on this branch was `extern thread_local boost::mt19937 randomGenerator`: a thread_local with a constructor that other translation units name directly is initialised through a weak per-unit wrapper, and mingw's emulated TLS is the platform that gets it wrong. The engine now lives behind `syncRandEngine()`, a function-local thread_local initialised in one place in `Utilities.cpp`, and every former use of the object calls the accessor (ab95800c2). With that the Windows savegame step passes, and Linux and macOS behave exactly as before (golden table, defaults test, savegame, trapped-unit and setup harnesses all unchanged). Reproduced outside CI too, on a Windows Server 2022 VM with the same MSYS2 toolchain: `SavegameSafetyHarness.exe` fails at 1617f7ecb with that exact assertion and passes at the same tree plus this patch. The harness also prints both RNG states now instead of a bare assert. On the invariant itself: the simulation still runs on the main thread only and the previewer threads never touch a live `Game`; a debug-time thread check on the simulation step is a reasonable follow-up for its own PR.
8. **Ids 7 and 8** are a substitution, as you read it: Shattered coast and Rugged archipelago are the old generators reworked under the same ids and keys, so an old lobby preference selects the reworked landscape. Documented in `MAP_GENERATOR_FRAMEWORK.md`.
9. **Preferences** are forward-compatible only, yes. An older binary rejects the new options section and falls back to defaults, which is the recovery path the harness tests.

### Two more things the fixes turned up

- **Old growth's lake size did not survive a lobby restart.** Its range reaches 160 and it shares the legacy `riverDiameter` slot, bounded at 100, so a saved lobby with a big lake failed to load and fell back to defaults. `CustomGameSetupHarness`'s whole-range round trip caught it once the harness could run again; the envelope is 0-160 (c6fb9d51c).
- **The lobby does gate the colony counts a generator refuses.** `CustomGameScreen` shows `setup.validation()` as the message and only starts when it is empty, so Symmetric arena at 3, 6 or 12 colonies and Ring world at 12 are refused with their translated reason rather than failing at generation. Concrete islands at 12 colonies is still marginal in the sweep (4/5 at 256, 2/3 at 512), as you found; the lobby's best of five absorbs it, and tightening its placement is on the follow-up list.

### Fairness tournament

Two full-rotation runs on this head (`tools/map_fairness_tournament.py`, Nicowar in every slot, 128×128, 4 colonies, best of 5 lobby rolls per seed, tick cap 90,000; every rotation of team numbers over starts is played, so wins by start carry no team-index effect and wins by team index carry no map effect):

- **`baseline`** (Symmetric arena only, six maps, 96 games, 69 decided by elimination, 2 of 2 re-run games reproduced exactly): position bias 0.0 pp against a fair-map floor of 8.7, no map biased after Benjamini-Hochberg, wins by colony index 26 / 15 / 29 / 26 (p 0.17). Wins by team index 25 / 18 / 18 / 35 (p 0.053), team 3 at 36.5% [27.5%, 46.4%]: the engine's processing-order effect sits at the edge of detectability on a map with nothing else in it, and is the thing to keep measuring.
- **`smoke`** (Symmetric arena, Contested commons, Watershed, Crater lakes; three maps each, 96 games, 82 decided): the same numbers as the 2026-09-12 run in `FAIRNESS_TOURNAMENT.md`, game for game. Team index pooled over all four generators 28 / 21 / 23 / 24 (p 0.80). Symmetric arena shows no start bias (0.0 pp, floor 14.4). The three asymmetric generators do, per map rather than per colony index: Contested commons 35 pp with 3 of 3 maps biased, Watershed 33 pp with 2 of 3, Crater lakes 29 pp with 2 of 3, and the best start on those maps wins 2.5 to 3.5 times its fair share whichever team plays it. The start scorer only partly predicts which (rho 0.59 on Contested commons, 0.68 on Crater lakes, none on Watershed).

The `standard` preset (every generator, 2,160 games, about 5.5 hours at 3 jobs) is not part of this PR; its generator list predates the landscapes and it belongs with the per-generator tuning that follows this merge. Reports and CSVs for both runs are under `artifacts/map-fairness/` locally and available on request.

### Verification on this head

- CI: [run 34844458453](https://github.com/Globulation2/glob2/actions/runs/34844458453) is green on linux (ubuntu:22.04), linux (ubuntu:24.04) and windows (mingw-w64), every step included: translations, the full build, the golden check with `--require-rows` against the committed Linux rows, `--sweep`, the setup harness, the `test/` suite, the savegame harness on Windows, and the menu-colony navigation test.
- Locally (macOS arm64, release): full build with no warnings in this branch's files; `MapGeneratorDefaultsTest` all groups; `MapGeneratorGoldenTest` 248 rows, 0 failures, with and without `--require-rows`; `--sweep` 0 failing combinations over 32 generators; `CustomGameSetupHarness` in default, `preferences-write` and `preferences-read` modes; `MenuColonyHarness` check and navigation; `test/TestsRunner` and every `test/` harness; `check_translations.py --strict`, `test_translations.py`, `test_font_coverage.py`, `test_text_area_layout.py`.

On playtesting: I have been playing these maps meticulously as they were built (City states over four iterations, Tidal flats three, the nine landscapes through six feedback rounds, the arenas three), and I am not done. The plan is to merge this as the framework plus the current catalog and keep tuning individual generators in smaller follow-up PRs, where each change is one map and one revision bump rather than forty thousand lines.

https://claude.ai/code/session_01GpUW6Ny43aUA1EzThn6cYE

genixpro:
Play-note batch pushed as `81622ca2a` (generators) and `65ae97262` (lobby); the PR description has the full list under "Play-note batch (2026-09-14)".

Screens and example maps: https://claude.ai/code/artifact/a10eb8c8-8dc2-49d6-9f73-f6645ff03755 (the compact lobby with Reset to defaults and Random parameters under the landscape chooser and the fairness/score line with its (i); the start quality breakdown; the picker after Randomize parameters; City states 256/4 with its archipelago; Carousel 256/3 with its lagoon and stone-banded spokes; Tidal flats 256/8 with six oases and eight sandbars per wedge).

To reproduce locally: `build/src/CustomGameSetupHarness artifacts/custom-game/compact` writes the lobby captures (`random-parameters-640.bmp`, `start-quality.bmp`, `landscape-picker-randomized.bmp`), and `python3 tools/render_map.py one city-states --seed 1 --size 256 --set teams=4 --out cs.png` (likewise `carousel` with `teams=3`, `tidal-flats` with `teams=8`) renders the maps.

Verification on this head, macOS arm64 release: `MapGeneratorDefaultsTest`; golden 248 rows, 0 failures with `--require-rows`, linux-x86_64 rows regenerated on a Linux x86-64 box (gcc); `--sweep` 0 failing combinations; `CustomGameSetupHarness` headless with captures plus both preferences modes; `MenuColonyHarness check`; all four translation checks.

Two things worth a maintainer's eye in play: Carousel's spokes now cross a lagoon between bands of stone, and the catalog shuffle makes Fingerprint the lobby's opening landscape.

genixpro:
Follow-up `459093f3d` from the last play notes: City states' archipelago now lies in the sea outside the design circle, where the torus wraps (an islet on the wrap point, a ring of eight and four more along the axes on a 256 map, radius 11 so the 10x4 plot leaves room to build), with the commons and strait back at 55 and 4 percent; the landscape picker gains Reset to defaults beside Randomize parameters; switching to the premade library while the first random preview is still rolling no longer lets that preview replace the premade choice. The results page linked above shows the new map.

The ubuntu 24.04 failure on the previous run was the visual harness: a random parameter set the world refused is redrawn by the preview itself, and the check only gave it one round. It now gives it the redraw rounds. Locally on macOS arm64: golden 248 rows with 0 failures under `--require-rows` (Linux rows regenerated on the x86-64 box), `--sweep` 0 failing combinations, City states 5/5 seeds at 128 and 256 for 2 to 8 colonies, and the harness's visual, preferences-write and preferences-read modes.

PR #243, opened 2026-09-10
Stop Map::syncStep from spinning when no gradient is allocated

Giszmo:
`Map::syncStep`'s round robin is `while (!updated)` with `updated` never set; it only exits through the returns inside its scans. Since gradients are allocated lazily (cac37e99), a team set without a single allocated field never reaches a return: a fresh world, a game with no teams, or a loaded game before any unit asks for a field hangs the whole game step. #186 hit it when its expulsion harness was rebased onto master (357d5fd7 papered over it by touching one gradient at setup, same as `TrappedUnitLifecycleTest`).

Fix: two passes instead of an open-ended loop. The first pass finds the next field still due; the second runs on freshly reset flags and finds the first allocated field. Finding nothing on the second pass means nothing is allocated.

Also drops the `getResourceGradient` workarounds from `BuildingExpelHarness` and `TrappedUnitLifecycleTest`, and adds a bare-world stepping scenario to `BuildingExpelHarness` so a regression fails the harness rather than the CI job timeout.

Verified locally (release build): `BuildingExpelHarness` 4/4 scenarios, `TrappedUnitLifecycleTest` via `run-savegame-safety-tests.py` all PASS with identical trace digests before and after save/load.

Note for #212: that branch already restructures this loop into `pickRoundRobinField` with the same two-pass shape, so its rebase resolves by keeping its own version.

### DIFF
--- a/src/map/MapStep.cpp
+++ b/src/map/MapStep.cpp
@@ -93,9 +93,10 @@ void Map::syncStep(Uint32 stepCounter)
 			updateExploredArea(team);
 	}
 	
-	// We only update one gradient per step, round robin over the gradients in use:
-	bool updated=false;
-	while (!updated)
+	// We only update one gradient per step, round robin over the gradients in use.
+	// Fields are allocated lazily: the second pass runs on freshly reset flags,
+	// so finding nothing there means no gradient exists yet and there is nothing to do.
+	for (int pass=0; pass<2; pass++)
 	{
 		int numberOfTeam=game->mapHeader.getNumberOfTeams();
 		for (int t=0; t<numberOfTeam; t++)
--- a/test/BuildingExpelHarness.cpp
+++ b/test/BuildingExpelHarness.cpp
@@ -33,8 +33,6 @@ struct World
 		game.map.setGame(&game);
 		game.addTeam(0);
 		team = game.teams[0];
-		// The gradient scheduler expects an in-use field; allocation is now lazy.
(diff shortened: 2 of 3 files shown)


genixpro (approved):
Independently confirmed on macOS/arm64 (clang), in an isolated worktree at `993e7193d`:

- **Reproduced the bug first.** Reverted just `src/map/MapStep.cpp` back to master's `while (!updated)` loop while keeping this PR's new `stepsWithoutAnyGradient` test, rebuilt, and ran it under a hard timeout: it hung indefinitely (no output, killed at 20s). Confirms the regression actually fails against unpatched code before trusting the fix.
- **Fix verified.** Restored the two-pass version, rebuilt: `BuildingExpelHarness` — all 4 scenarios PASS (0.3–0.4s), including the new hang case.
- **Save/load continuity holds.** `TrappedUnitLifecycleTest` via `run-savegame-safety-tests.py --check-preferences`: PASS on trapped elimination, indoor survival, rescue, service, hatchery recovery, repeated seeds and save continuation; disposable-profile preferences unchanged.
- `Map::syncStep` only compiles under `#ifndef YOG_SERVER_ONLY`, so the client/headless build tested here is the right target; nothing analogous applies to `server=1`.

This adds macOS coverage on top of the already-green Linux/Windows CI. Small, self-contained, no save/replay/network-compatibility surface. Approving and merging.

PR #249, opened 2026-09-11
Fix middle-button panning being dropped, and hold the automatic torus while panning

Giszmo:
Middle-button panning could silently do nothing, and the automatic torus overview retracted while the button was still held.

**Dropped pans.** `GameGUI::step()` coalesces mouse motion and replays it after the poll loop, while button events are handled inline — so within a frame, buttons are always processed before motion. MMB panning lives entirely in the motion handler and the release clears `panPushed`, so any drag whose release landed in the same frame was discarded outright. A flick that fits inside one frame lost all of its travel; slower drags worked, which is why it felt random. Left-button drags were shielded by the existing do-not-skip-map-cells flush, gated on `SDL_BUTTON(1)`. Fix: flush pending motion before a button event.

**Right panel.** Middle-click over the right panel started a pan, except when a building was selected — that case fell into a debug branch whose body has been commented out for years, so the click did nothing. Removed. `panViewX`/`panViewY` went write-only when panning moved to the camera origin and go with it.

**Automatic torus.** `notifyMove` only fires on tile crossings and the overview retracts 250 ms after the last one, so holding the button still mid-pan looked identical to having finished. A held pan now suspends the retract timer; release behaves as before.

### Verification

Driven with xdotool under Xvfb, `-test-games --map balanced`.

Pan, measured by cross-correlating before/after screenshots, for a 200 px middle-button flick that fits in one frame:

| | map content moved |
| --- | --- |
| master | 0 px |
| this branch | 200 px |

Multi-frame drags still move exactly the distance dragged, with no double application.

Torus, `automaticTorus=1`: drag, then hold the button still for 2 s. Share of overview-sky pixels in the map area:

| | idle | dragging | held still | released |
| --- | --- | --- | --- | --- |
| master | 0.3% | 0.3% | 0.3% | 0.3% |
| this branch | 0.3% | 0.2% | 36.8% | 0.5% |

Middle button held still, master then this branch:

master
this branch

`test/TestsRunner` 185 tests OK and `WinningConditionsHarness` clean at 5141903. No simulation code is touched, so saves, replays and network compatibility are unaffected.

**Feel:** the overview now stays unfolded for as long as the middle button is down instead of snapping back after 250 ms of stillness. Worth a maintainer actually panning around with `automaticTorus` on before merging.

Screenshots are committed in 27b1853 and removed at the branch tip, matching #238. Branched from 8b8c9341b; the three master commits since touch a disjoint set of files.


genixpro:
Read this one closely, traced every path it touches, and built and ran it on Linux/x86_64. **No defects found** — which is a real result, not a shrug; I went looking specifically at the places this kind of change usually breaks.

## What I checked and why

**The flush cannot double-process.** The motion branch already replays the *previous* stored motion when a left-drag crosses a map cell, then immediately overwrites `mouseMotionEvent` and re-arms the flag. The new flush clears `wasMouseMotion`, and the post-loop flush is guarded on it, so a pending motion is now processed exactly once — at the button event or after the loop, never both. Walked M1→M2→B and B→M2→end; both are clean.

**The reorder is chronologically correct.** Motion now runs before `lastMouseButtonState` is updated, which is right: the motion genuinely happened first. And flushing before a middle-button *press* is a quiet bonus — previously the press set `panMouseX` from its own coords and the stale pending motion was then applied against it, producing a small spurious jump.

**It also fixes the zoom control and minimap silently losing their last drag increment**, for exactly the same reason the pan was lost: `handleMouseButtonUp` clears `zoomControlPushed`/`miniMapPushed` and early-returns, so a same-frame drag+release was discarded. Worth mentioning in the PR description — it's a second user-visible fix riding along.

**`panViewX`/`panViewY` really were write-only.** `grep` confirms zero remaining references after removal.

**Pushing `setPanHeld(panPushed)` every frame is the right call, not laziness.** `panPushed` is cleared in six places (`GameGUIInput.cpp:123,346,363`, `GameGUIInputKey.cpp:81`, `GameGUITorus.cpp:70`, `GameGUI.cpp:139`). Mirroring it at press/release would have drifted out of sync at the first one anyone forgot. The comment says as much and it's correct.

**The interaction I most expected to be broken isn't.** Holding the overview open during a pan means motion and the middle button now reach the input path *while the torus is active*. I chased that: `TorusView::event` only intercepts the wheel and only when `target`, and `handleTorusPointer` filters to `SDL_BUTTON_LEFT` press/release, so middle-button pan and all motion fall through to the ordinary 2D path. `handleMouseMotion` has no torus early-return either. The pan keeps working with the overview up.

**`panHeld` only suppresses clearing `moving`, it never sets it.** So pressing the middle button without moving does not summon the overview — it only holds out one that a real pan already raised. That's the conservative choice and the right one.

## Verification on my side

Built clean (gcc 15, `-Wall`, no new warnings), `TestsRunner` 185 OK, and every `test/*Harness` and `test/*Test` in CI's loop passes — 0 failures.

## Two things for a maintainer, not for me

1. **The feel change is the actual review question**, as your own description says. The overview now stays unfolded for as long as the middle button is down instead of retracting after 250 ms of stillness. That is a deliberate behaviour change and wants someone panning around with `automaticTorus` on, which no amount of reading substitutes for.

2. **Middle-clicking over the right panel now starts a pan** where, with a building selected, it previously did nothing. That's a consequence of deleting the dead debug branch, and it makes behaviour *consistent* (with no building selected it already panned), so I think it's right — but it is a behaviour change that isn't called out in the description.

Minor: the branch is behind master (#252 and #253 have landed). GitHub reports no textual conflict and the changed files are disjoint, so this is a note rather than a request.

Giszmo:
The "two things for a maintainer" part are deliberate fixes, not bugs holding it back.

I would interpret MMB as "wants to pan for more", so I would not use the slow zoom out an accidental "mouse gets to the edge of the screen" triggers but a fast zoom out and a fast zoom in on release but that would be the behavioral change arguably not being a strict fix, so ... merging this as is.

PR #252, opened 2026-09-11
Shrink the audio device buffer and skip opening it while muted

Giszmo:
Shutdown was gated on the audio device. `SoundMixer::openAudio` asked SDL for
`SAMPLE_COUNT_PER_SLICE>>1` = 16384 frames at 44100 Hz, i.e. a 371 ms buffer, and
`SDL_CloseAudio` has to wait for the in-flight callback before the process can exit.
Muting did not help: the constructor zeroed the volumes and then opened the device
anyway, so a silent single-player session still ran the audio thread, still decoded
Ogg every 371 ms, and still paid the same exit wait.

### What changed

**The fade no longer defines the buffer size.** `interpolationTable` was indexed by
the sample offset *within one callback*, so one buffer was exactly one music fade.
Shrinking the buffer on its own would have cut every fade from 371 ms to 23 ms.
A new `fadePos` member carries the ramp position across callbacks, so a fade spans
`FADE_SAMPLE_COUNT` samples however much the device buffer holds. `MODE_EARLY_CHANGE`
now aligns its two tracks once at fade start instead of re-seeking every callback.
With that in place the device buffer drops to 1024 frames.

**Muted starts leave the device closed.** `setVolume` already implemented
stay-closed-while-muted and open-lazily-on-unmute; the constructor now honours it.
`setNextTrack` keeps its track selection while the device is closed so unmuting
resumes the music instead of waiting for the next track change.

### Measurements

`-test-games 1`, dummy audio driver, Xvfb, ~350 MB resident, SIGTERM to process exit:

| | `SDL_CloseAudio` | total exit |
|---|---|---|
| master, sound on | 204 ms | 292 ms |
| this branch, sound on | 58 ms | 175 ms |
| master, muted | 219 ms | 303 ms |
| this branch, muted | device never opened | 109 ms |

### Correctness

The music is unchanged. Capturing the raw output stream of both builds through SDL's
disk driver and aligning on the first non-silent frame, the first **3 seconds are
bit-identical** (`max |diff| = 0` over 132300 frames) — the fade-in ramp and the
decoded music match sample for sample.

Tracing the state machine confirms both ramps still span the full 371 ms: `MODE_START`
and `MODE_EARLY_CHANGE` each run exactly 16 callbacks, `fadePos` stepping 0 to 30720
by 2048, before transitioning.

Music also starts sooner as a side effect: first audible frame at 24 ms instead of
373 ms, since the first callback is filled 16x faster.

Unmuting mid-session was exercised in the real app under Xvfb: launched with `-m`
(no `SDL_OpenAudio` at all), opened Settings, Audio, toggled Mute off — the device
opens at 1024 frames and the Intro track fades in with the same envelope.

`test/TestsRunner` 187 tests OK, `WinningConditionsHarness` OK.

### Note, not addressed here

`GameGUI::~GameGUI` calls `settings.save()` when `rememberUnit` is set, writing the
whole settings struct, so any command-line override (`-m`, `-g`, `-G`, ...) becomes
permanently persisted after one game. That is a separate correctness issue.

### DIFF
--- a/src/SoundMixer.cpp
+++ b/src/SoundMixer.cpp
@@ -170,7 +170,17 @@ void mixaudio(void *voidMixer, Uint8 *stream, int len)
 		{
 			mixer->fadePos = 0;
 			mixer->actTrack = mixer->nextTrack;
-			mixer->mode = SoundMixer::MODE_NORMAL;
+			if (mixer->pendingTrack >= 0)
+			{
+				// a change asked for while this fade was running: cross into it
+				// from the track that just landed, so the mix stays continuous.
+				// Staying in MODE_EARLY_CHANGE with fadePos == 0 re-aligns the
+				// incoming track on the next callback.
+				mixer->nextTrack = mixer->pendingTrack;
+				mixer->pendingTrack = -1;
+			}
+			else
+				mixer->mode = SoundMixer::MODE_NORMAL;
 		}
 	}
 	else
@@ -293,6 +303,7 @@ SoundMixer::SoundMixer(unsigned musicvol, unsigned voicevol, bool mute)
 	this->voiceVolume = voicevol;
 	mode = MODE_STOPPED;
 	fadePos = 0;
+	pendingTrack = -1;
 	soundEnabled = false;
 	speexDecoderState = NULL;
 	
@@ -377,8 +388,22 @@ void SoundMixer::setNextTrack(unsigned i, bool earlyChange)
 
 	SDL_LockAudio();
 
-	// Select next tracks
-	if (actTrack >= 0)
+	// A fade now spans many callbacks, so a track change can be asked for while
+	// one is still running — GameMusicController can emit on consecutive 40 ms
+	// ticks. Restarting the fade would cut the incoming track off mid-mix, so
+	// queue the request and let mixaudio() start it when this fade lands.
+	if (soundEnabled && mode == MODE_EARLY_CHANGE)
+	{
+		pendingTrack = static_cast<int>(i);
+		SDL_UnlockAudio();
+		return;
+	}
+
+	// Select next tracks. While the device is closed nothing is playing, so the
+	// selection is both the current and the next track: leaving actTrack at the
+	// first track ever selected would make setVolume() resume that one on
+	// unmute, whatever was asked for since.
+	if (soundEnabled && actTrack >= 0)
 		nextTrack = i;
 	else
 		nextTrack = actTrack = i;
@@ -389,12 +414,14 @@ void SoundMixer::setNextTrack(unsigned i, bool earlyChange)
 		if (mode == MODE_STOPPED)
 		{
 			fadePos = 0;
+			pendingTrack = -1;
 			SDL_PauseAudio(0);
 			mode = MODE_START;
 		}
 		else if (earlyChange)
 		{
 			fadePos = 0;
+			pendingTrack = -1;
 			mode = MODE_EARLY_CHANGE;
 		}
 	}
@@ -451,6 +478,7 @@ void SoundMixer::stopMusic(void)
 {
 	SDL_LockAudio();
 	fadePos = 0;
+	pendingTrack = -1;
 	mode = MODE_STOP;
 	SDL_UnlockAudio();
 }
--- a/src/SoundMixer.cpp
+++ b/src/SoundMixer.cpp
@@ -21,7 +21,13 @@ using namespace GAGCore;
 #include <malloc.h>
 #endif
 
-#define SAMPLE_COUNT_PER_SLICE 4096*8
+//! Length of a music fade in Sint16 samples, both channels interleaved.
+//! Independent of the device buffer size: a fade spans as many callbacks as
+//! it takes to cover this many samples.
+#define FADE_SAMPLE_COUNT 4096*8
+//! Frames per device buffer. SDL_CloseAudio waits for the in-flight callback,
+//! so this bounds how long teardown blocks.
+#define DEVICE_FRAME_COUNT 1024
 #define INTERPOLATION_RANGE 65535
 #define INTERPOLATION_BITS 16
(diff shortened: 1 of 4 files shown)


genixpro:
Reviewed this closely — the core of it is right, and I've pushed 1aff6549d with two fixes rather than sending it back, since both are small. Shout if you'd rather I'd left them to you.

**What's good:** splitting `SAMPLE_COUNT_PER_SLICE` into `FADE_SAMPLE_COUNT` and `DEVICE_FRAME_COUNT` is the correct decomposition — those were two unrelated concepts sharing a literal, and nobody could shrink the buffer safely without spotting that. The locking discipline on `fadePos` matches how `mode` is already handled, the `fadeValue` clamp is right (the smoothstep reaches exactly `INTERPOLATION_RANGE` at `l`, so an overshoot lands on full-on/full-off), and moving the crossfade's `ov_pcm_seek` to once per fade is equivalent and cheaper. The bit-identical 3-second capture is exactly the right evidence for the fade rewrite.

**Fix 1 — unmuting resumed the wrong track.** `setNextTrack` now runs with `soundEnabled` false, but kept moving only `nextTrack` once `actTrack` was set. `actTrack` means "playing right now", and with the device closed nothing is playing, so it stayed pinned to the first track ever selected.

`GlobalContainer::loadClient` asks for Intro then Menu, so a muted start left `actTrack=Intro, nextTrack=Menu`. On unmute `setVolume()` resumes `actTrack` — Intro — and that call also overwrites `nextTrack` with Intro, so the EOF handoff loops it: **menu.ogg never plays at all.** Unmuting inside a game is worse: `Engine::run`'s `InGameDefault` only ever reached `nextTrack`, so the intro plays over the match until some later event fires an early change.

I think the verification note is what hid this — "the Intro track fades in with the same envelope" is the symptom. It checked that sound resumed, not that the right sound resumed.

**Fix 2 — a track change mid-fade cut the incoming track off.** A fade used to be one callback, so this window was zero-length; it's now sixteen callbacks, and `GameMusicController` can emit on consecutive 40 ms ticks, so it's reachable in a busy battle. Restarting the fade dropped the partially mixed track instantly — abort at 90% and the output jumps from 10% A + 90% B back to 100% A, a hard cut. The request is now queued in `pendingTrack` and started when the running fade lands.

**Verification:** added `test/SoundMixerTrackSelectionHarness`, picked up by CI's existing `test/` harness loop with no workflow change. It loads the real ogg tracks and drives `mixaudio()` directly with SDL's callback thread parked, so callback counts are deterministic. I checked it actually fails on the unfixed tree: reverting the selection fix alone fails four checks (actTrack 0 where 1 and 2 are wanted — the intro), reverting the queueing alone fails five. `TestsRunner` 187 OK.

Two things I did **not** touch:

- `mute → unmute → mute` never re-closes the device. Consistent with the existing design and you don't claim otherwise, but "skip opening it while muted" only holds before the first unmute.
- The `GameGUI::~GameGUI` settings-persistence bug you flagged. Agreed it's out of scope — but it's arguably worse now that mute means "device never opens", so it's worth its own issue rather than living in a PR description.

One thing I can't check from here: the 1024-frame buffer is a normal game buffer and 371 ms was absurd, but the removed `// Dejan: this is supposed to fix reported problem on Gentoo` comment says somebody once hit platform-specific trouble here. All the measurements are Linux + dummy driver. If you have a Windows box handy, that's the gap.

genixpro:
Ran this on real audio hardware (Realtek ALC1220, Ubuntu, SDL 2.32.10), which turned out to be worth doing: **the PR understates its own benefit by roughly an order of magnitude.** The measurements in the description used SDL's dummy driver, which has no hardware buffer to drain, so it can't show the cost the change actually removes.

Isolated `SDL_CloseAudio`, 3 reps each, requesting the two buffer sizes directly:

| driver | buffer | `SDL_CloseAudio` | callbacks in 2 s (actual/expected) |
| --- | --- | --- | --- |
| pulseaudio | 16384 (master) | 783–784 ms | 5/5 |
| pulseaudio | 1024 (this PR) | **47–51 ms** | 85–86/86 |
| alsa | 16384 (master) | 1739–1742 ms | 7/5 |
| alsa | 1024 (this PR) | **112–115 ms** | 86/86 |

Full game, `-test-games 1` under Xvfb, 12 s of play then SIGTERM, timed to process exit, 3 reps each, throwaway `HOME` per run so the settings-save bug can't leak state between them:

| | master | this PR |
| --- | --- | --- |
| sound on | 870 / 882 / 911 ms | **152 / 164 / 173 ms** |
| muted | 864 / 876 / 886 ms | **91 / 112 / 114 ms** |

The muted row is the clean demonstration that master really did open the device and pay full freight to produce silence.

**On the Gentoo comment.** I was the one who raised it, so: at 1024 frames the callback rate is exactly on target on both drivers (86/86), and all 12 full-game runs logged zero underrun/xrun/ALSA warnings. That's real evidence the smaller buffer keeps up, on modern ALSA and PulseAudio at least. Still nothing exercising Windows audio — mingw builds but never opens a device — so that stays untested by anyone.

**CI**: mingw-w64 and ubuntu:24.04 green on 1aff6549d, and I confirmed `SoundMixerTrackSelectionHarness` actually ran on 24.04 with all 23 checks passing.

ubuntu:22.04 went red on something unrelated: `CustomGameSetupHarness` aborting on `assert(generator.generateMap(g, d))` (test/CustomGameSetupHarness.cpp:834). I ran that harness 25x on the **master baseline** 4e00c2b17, with no audio changes at all, and got **1 failure in 25**. It's an unseeded `eRIVER` / 4-team generation that flakes around 4% of the time — so any PR running both Linux images has roughly a 1-in-12 chance of a spurious red. Worth its own issue and a pinned seed; happy to file it if nobody's already on it.

One process note, since the automatic checks on this PR look stale: the `pull_request` event never fired for my push — no runs triggered repo-wide for over an hour — so I dispatched the workflow by hand. The PR's check list still points at the old SHA.

PR #255, opened 2026-09-11
Let any worker stock a completed building (draft: keep the schooling gate for sites only)

Giszmo:
**Draft for discussion.** Stacked on #256 (first commit is that one; the second is this change). One-line behaviour change plus harness cases; the question is whether the game should work this way.

## What the code does today

`Building::canUnitWorkHere` (src/building/Misc.cpp) refuses a worker for any fetch job at a non-flag building whose `type->level` is above the worker's `level[HARVEST]`. That covers construction and upgrade sites, which is the reading everyone has of the rule ("you need schooled workers to build level 2"), but it equally covers completed buildings that only want stock:

- inn level 2 and 3: wheat and fruit
- defence tower level 2 and 3: stone

A level-0 worker four tiles from a completed level-2 inn, carrying wheat, is refused with `UnitTooLowLevel` (real-engine run on master, see the harness in this PR). Only the school raises the harvest level, and it raises harvest and build together, so in a normal game the gate is invisible: you cannot upgrade an inn without schooled workers, and once you have them they also feed it. It bites when the schooled units are busy, dead, or few: every unschooled worker in the colony is then barred from feeding the upgraded inns and stocking the upgraded towers, and the only sign is the "N units too low level" line in the building panel.

The upgrade button is gated separately on `Team::maxBuildLevel()`; #256 underneath this makes both read the same worker level.

## The change

Keep the gate for sites (`type->isBuildingSite`), drop it for completed buildings. Any worker can carry. `VERSION_MINOR` 96 → 97 and the replay floor with it; no save-format change.

## Gameplay discussion (please weigh in)

**For lifting it**

- The rule is not explained anywhere in the game. `[%0 units too low level]` is the only string that mentions it, one of eight failure reasons in the building panel, and the tutorial and help texts do not cover it (grep over `data/texts.en.txt` and `docs/`). A player with many level-2 towers and few schooled workers gets towers that never fill and has to work out why.
- The "builder" reading of the school is intact: sites keep the gate, so upgrading still needs schooled workers, and schooling still buys chop speed and build speed.
- The punishment is disproportionate to the mistake: one dead schooled worker can leave a level-3 inn unfed while ten idle level-0 workers stand next to it.

**Against lifting it**

- It is a balance lever. Today an upgraded inn or tower is a commitment: you need schooled workers *and* you need to keep them alive and free to keep it running. After this, one schooled worker builds the level-3 inn and the swarm feeds it. Level-3 inns feed faster (`timeToFeedUnit` 24 → 15 → 9), so this makes the upgrade cheaper to sustain.
- AI teams benefit too, and Nicowar upgrades inns freely; the AI bed numbers below show how much.
- Players who *have* internalised the rule lose a reason to build a second school.

My read: lift it. The rule punishes a state the game never explains, and the strategic content (school before you upgrade) survives in the site gate. But this is a feel decision for a maintainer, per CLAUDE.md.

## Measurement

Eight Nicowars on Playground, resumed from a tick-25000 save, run to 40000, ASLR off, `MALLOC_PERTURB_` swept over 0/85/170 (the protocol from #253; this branch does not carry that fix, so the spread across fills is reported rather than hidden). End-of-run totals are in the first comment. Leo's point stands that the effect can only start once the first inns and towers are upgraded, so a per-512-tick series of deliveries against the count of upgraded buildings is being measured and will follow in a second comment.

## Verification

- `level-gate-test` harness (from #256, cases updated here): a level-0 worker is hired by completed level-2/3 inns and a level-3 tower, refused by a level-2 site; level-1/2 workers build level-2/3 sites; `build=1, harvest=0` builds a level-2 site. On master the three completed-building cases fail with `UnitTooLowLevel`.
- `TestsRunner` (187) green.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -159,6 +159,10 @@ jobs:
           scons -j$(nproc) release=1 server=0 hiring-bucket-test
           python3 test/run-savegame-safety-tests.py --check-preferences build/src/HiringBucketHarness .
 
+      - name: Build and run the hiring level-gate regression
+        run: |
+          scons -j$(nproc) release=1 server=0 level-gate-test
+          python3 test/run-savegame-safety-tests.py --check-preferences build/src/LevelGateHarness .
       - name: Build and run the resource-fetch target regression
         run: |
           scons -j$(nproc) release=1 server=0 resource-fetch-target-test
--- a/src/ReplayReader.h
+++ b/src/ReplayReader.h
@@ -25,9 +25,11 @@ static constexpr Uint32 REPLAY_MIN_VALID_ORDERS = 5;
 //! the reader still accepts. Replays older than this are rejected: versions 90, 92,
 //! 93, 94, 95 and 96 changed the simulation (weighted pathfinding and diagonal timing,
 //! hiring-bucket iteration, trapped-colony elimination, fetch-job apportionment,
-//! round-trip routing and hiring, one worker level), so earlier replays would
-//! diverge from what happened.
-static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 96;
+//! 93, 94, 95, 96 and 97 changed the simulation (weighted pathfinding and diagonal timing,
+//! hiring-bucket iteration, trapped-colony elimination, fetch-job apportionment,
+//! round-trip routing and hiring, one worker level, ungated stocking of completed
+//! buildings), so earlier replays would diverge from what happened.
+static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 97;
 
 /// This class is used for reading replays.
 /// The replay stream is kept open and read every time you do retrieveOrder.
--- a/src/ReplayReader.h
+++ b/src/ReplayReader.h
@@ -23,10 +23,11 @@ static constexpr Uint32 REPLAY_MIN_VALID_ORDERS = 5;
 
 //! Oldest replay format (the VERSION_MINOR the replay was written with) that
 //! the reader still accepts. Replays older than this are rejected: versions 90, 92,
-//! 93, 94 and 95 changed the simulation (weighted pathfinding and diagonal timing,
+//! 93, 94, 95 and 96 changed the simulation (weighted pathfinding and diagonal timing,
 //! hiring-bucket iteration, trapped-colony elimination, fetch-job apportionment,
-//! round-trip routing and hiring), so earlier replays would diverge from what happened.
-static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 95;
+//! round-trip routing and hiring, one worker level), so earlier replays would
+//! diverge from what happened.
+static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 96;
 
 /// This class is used for reading replays.
 /// The replay stream is kept open and read every time you do retrieveOrder.
--- a/src/SConscript
+++ b/src/SConscript
@@ -659,6 +659,13 @@ if not env['server'] and 'hiring-bucket-test' in COMMAND_LINE_TARGETS:
     hiring_sources += local.Object('HiringBucketHarness.o', '#test/HiringBucketHarness.cpp')
     local.Alias('hiring-bucket-test', local.Program('HiringBucketHarness', hiring_sources))
 
+# Which workers a building may hire: completed buildings take any worker,
+# construction and upgrade sites keep the schooling-level gate.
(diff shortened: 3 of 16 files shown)


Giszmo:
## Measurement

Eight Nicowars on Playground, resumed from the tick-25000 saves, run to 40000, ASLR off (`setarch -R`), `MALLOC_PERTURB_` swept over 0 / 85 / 170. Both columns are **without** the #253 fix (this branch is on master), so seed 7 is heap-sensitive on both sides and is shown at two fills; every other seed was identical across all three fills on both sides. Master numbers are the same protocol on master 4e00c2b17 (Carol's runs); this branch is 4c0068512 with the same delivery/hunger instrumentation applied locally. Population = units alive at tick 39936.

| seed | deliveries master / this | starvation deaths master / this | population master / this |
|---|---|---|---|
| 7 (fill 0) | 3322 / 3272 | 39 / 53 | 272 / 249 |
| 7 (fill 170) | 3289 / 3153 | 44 / 62 | 244 / 255 |
| 11 | 2839 / 3070 | 23 / 26 | 228 / 253 |
| 12 | 1792 / 1809 | 53 / 36 | 169 / 168 |
| 13 | 2669 / 2917 | 35 / 17 | 282 / 300 |
| 14 | 2939 / 2919 | 21 / 12 | 287 / 268 |
| 15 | 2987 / 2933 | 30 / 24 | 243 / 234 |
| **total (fill 0)** | **16548 / 16920** (+2.2%) | **201 / 168** (−16%) | **1481 / 1472** (−0.6%) |

Read: starvation deaths drop on five of six seeds, deliveries are up on the two seeds where the AI had upgraded inns it could not staff (11, 13) and flat elsewhere; seed 7 goes the other way and is the noisy one. Consistent with the mechanism: Nicowar upgrades inns freely and this lets its unschooled workers feed them. Not a large effect on this bed, and this bed is AI-only; the human-play question above is the one that decides the PR.

`TestsRunner`: 187 OK on 4c0068512.

Giszmo:
## Deliveries over time, master vs this branch

Same bed and protocol as the first comment (eight Nicowars, tick-25000 saves, run to 40000, ASLR off, `MALLOC_PERTURB_=0`; this time both binaries carry a per-512-tick counter). One thing the series settles first: the warm saves already hold upgraded inns and towers at tick 25000 (39 across the six seeds, 51 by tick 26624), so the whole measured range is after the first upgrade and there is no unupgraded lead-up diluting it.

Deliveries per 2048-tick window, summed over the six seeds and all teams; "upgraded" = completed level-2/3 inns + towers, summed over seeds, at window end.

| window start | master | this | delta | upgraded master / this |
|---|---|---|---|---|
| 24576 (from 25000) | 1341 | 1392 | +3.8% | 39 / 38 |
| 26624 | 2611 | 2660 | +1.9% | 51 / 48 |
| 28672 | 2306 | 2393 | +3.8% | 60 / 60 |
| 30720 | 2124 | 2372 | +11.7% | 65 / 75 |
| 32768 | 2267 | 2357 | +4.0% | 75 / 83 |
| 34816 | 2102 | 2140 | +1.8% | 77 / 79 |
| 36864 | 2214 | 1997 | −9.8% | 86 / 80 |
| 38912 (to 40000) | 1486 | 1434 | −3.5% | 78 / 79 |

Per seed, deliveries from the first upgraded building (tick 25088 on every seed) to the end, master / this: 7: 3256 / 3124, 11: 2810 / 3038, 12: 1768 / 1780, 13: 2640 / 2890, 14: 2919 / 2903, 15: 2951 / 2903. Four up, two down; the two down are seed 7 (the heap-sensitive one, see #253) and seed 15 (−1.6%).

Read: the branch runs ahead for the first six windows, by 2–12%, while the colonies are upgrading and unschooled workers can now feed the new inns, and falls behind in the last two windows, when it has more upgraded buildings and more units alive (population +2.3% at the end, starvation −16%) and the AI economies saturate. Fill 85 gives the same shape (+2.5 / +7.0 / +11.2 / +2.0 / +3.2 / −6.9 / −5.7%). Net over the run +2.2% deliveries, as in the first comment. Same conclusion: a modest, front-loaded gain on the AI bed; the decision is the human-play one.

Giszmo:
Closed on Leo's call: the effect is real (up to +12% deliveries per window on the AI bed), which means the schooling gate is a real balance lever and stays. The player-facing problem is that the game does not show *which* units fail the gate; that gets its own PR: red shapes next to the failure counts in the building panel, and the same shapes over the exact units in the map view.

PR #248, opened 2026-09-11
Add Cabino: a resurrected port of the original (2005-2007) Nicowar AI

genixpro:
## What

Resurrects the original 2005-2007 Nicowar — before its 2007-01-31 from-scratch rewrite onto the Echo framework (which kept the `Nicowar` name for what's now `NewNicowar`/`AINicowar.*`) — as a new, separately-named AI: **Cabino** (`AI::CABINO = 8`; `7` intentionally left free, reserved by the in-progress Maxima branch).

Where the current Nicowar is a phase-driven strategy (growth → skilled-work → upgrade → war-prep → war, plus fruit/starvation/explorer-attack sub-phases, all externally tunable via `data/nicowar.default.txt`), the original architecture Cabino resurrects is a set of independent specialist modules — defense, attack, construction, upgrades, unit/swarm management, plus smaller ones (exploration, inns, towers, clearing, happiness/fruit) — each acting on its own initiative every tick rather than following one central plan. Same rough goal (grow fast, then throw large armies of high-level warriors at opponents), different architecture getting there — a "cabinet of departments" instead of a state machine.

## Why real bugs, not just a straight port

Porting ~6,000 lines of C++ that hasn't compiled in ~19 years surfaced routine API drift (`boost::shared_ptr`→`std::shared_ptr`, `Ressource`→`Resource` renames, `Order::getOrder`'s new `versionMinor` parameter, `BrushAccumulator::applyBrush`'s swapped argument order, `OrderAlterClearArea`/`OrderAlterForbidden`'s new required `Map*` argument, `std::random_shuffle`'s removal in C++17) — but headless self-play surfaced several **real, independent bugs** in the original logic itself, fixed here rather than preserved 1-for-1 (explicitly fine per the conversation that asked for this — bug fixes and basic improvements were welcomed, not a strict 1:1 port):

- **Four out-of-bounds team-array reads.** Several loops iterated `ai.game->teams[0..31]`, a holdover from when the live team-slot array actually had 32 entries. It's now sized `Team::MAX_COUNT` (12), so the tail of each loop read past the array into unrelated `Game` fields and could dereference whatever garbage it found there as a `Team*` — the actual cause of a wild-pointer crash a few ticks into every game. Fixed to bound on `Team::MAX_COUNT`.
[...]

PR #256, opened 2026-09-11
Read one worker level where hiring read harvest and the menu read build

Giszmo:
Marked as a bugfix on Leo's call; #255 (any worker stocks completed buildings) will sit on top of this.

## Bug

A worker carries two schooling levels, `level[HARVEST]` and `level[BUILD]`, for one concept. Only the school raises either, and it raises both in the same visit (`upgradeInParallel`), so in play they are always equal. The code, however, reads them inconsistently:

| question | reads |
|---|---|
| may this worker be hired for a building of tier N? (`Building::canUnitWorkHere`) | harvest |
| may the player upgrade a building to tier N? (`Team::maxBuildLevel`, upgrade button) | build |
| fetch ranking (`bringResourcesLevel`) | harvest |
| chop speed / fill speed | harvest / build |

The editor exposes both as separate boxes, and the unit panel shows two rows with the same number. A map with the two set apart can offer an upgrade the colony's workers are then refused for.

Real-engine run on master, one hiring pass, worker four tiles from a level-2 inn *site* carrying wood: `harvest=0 build=1` → refused (`UnitTooLowLevel`), `harvest=1 build=0` → hired. The upgrade menu would say the opposite.

## Fix

- `Unit::workerLevel()` is the one read; `Unit::setWorkerLevel()` sets both halves and their performance together.
- Hiring gate, `Team::maxBuildLevel` and the fetch ranking read through it.
- Load normalises a unit whose two levels differ to the higher one, so existing maps keep working and cannot desync on it.
- The editor keeps the Build box (it now drives both) and drops the Harvest box. The unit panel drops the Harvest row.
- Save format and race tables untouched: both abilities are still stored and still drive their own speeds.
- `VERSION_MINOR` → 96, replay floor with it.

## Verification

- `level-gate-test` harness (new, run in CI): tier gate on completed inns and sites, plus `build=1, harvest=0` hired by a level-2 site and `build=0, harvest=1` refused.
- `TestsRunner` (187).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -175,6 +175,10 @@ jobs:
           scons -j$(nproc) release=1 server=0 hiring-bucket-test
           python3 test/run-savegame-safety-tests.py --check-preferences build/src/HiringBucketHarness .
 
+      - name: Build and run the hiring level-gate regression
+        run: |
+          scons -j$(nproc) release=1 server=0 level-gate-test
+          python3 test/run-savegame-safety-tests.py --check-preferences build/src/LevelGateHarness .
       - name: Build and run the resource-fetch target regression
         run: |
           scons -j$(nproc) release=1 server=0 resource-fetch-target-test
--- a/src/FileFormatVersions.h
+++ b/src/FileFormatVersions.h
@@ -96,6 +96,11 @@ static constexpr int FILE_FORMAT_VERSION_ROUND_TRIP_FIELDS = 95;
 //! (MapIO.cpp:350, 410, 451, 514).
 static constexpr int FILE_FORMAT_VERSION_TOPOLOGY_GENERATION = 97;
 
+//! Harvest and build became one worker level. Saves before this could hold the
+//! two apart — the map editor offered a box for each — so the loader evens a
+//! worker out to the higher of the two (UnitSerialization.cpp:90).
+static constexpr int FILE_FORMAT_VERSION_ONE_WORKER_LEVEL = 98;
+
 // === Save-file section signatures (4-byte ASCII tags) ===
 // Embedded as four chars at the start of each save section so a corrupted
 // stream fails fast. NEVER change these values — old saves on disk depend
--- a/src/ReplayReader.h
+++ b/src/ReplayReader.h
@@ -23,12 +23,12 @@ static constexpr Uint32 REPLAY_MIN_VALID_ORDERS = 5;
 
 //! Oldest replay format (the VERSION_MINOR the replay was written with) that
 //! the reader still accepts. Replays older than this are rejected: versions 90, 92,
-//! 93, 94, 95, 96 and 97 changed the simulation (weighted pathfinding and diagonal
+//! 93, 94, 95, 96, 97 and 98 changed the simulation (weighted pathfinding and diagonal
 //! timing, hiring-bucket iteration, trapped-colony elimination, fetch-job
 //! apportionment, round-trip routing and hiring, Echo building-order ids surviving a
-//! load, route fields invalidated by the map's topology generation), so earlier
-//! replays would diverge from what happened.
-static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 97;
+//! load, route fields invalidated by the map's topology generation, one worker level),
+//! so earlier replays would diverge from what happened.
+static constexpr Uint16 REPLAY_MINIMUM_VERSION_MINOR = 98;
 
 /// This class is used for reading replays.
 /// The replay stream is kept open and read every time you do retrieveOrder.
--- a/src/SConscript
+++ b/src/SConscript
@@ -672,6 +672,13 @@ if not env['server'] and 'hiring-bucket-test' in COMMAND_LINE_TARGETS:
     hiring_sources += local.Object('HiringBucketHarness.o', '#test/HiringBucketHarness.cpp')
     local.Alias('hiring-bucket-test', local.Program('HiringBucketHarness', hiring_sources))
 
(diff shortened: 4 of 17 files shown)


genixpro:
Not a review of the change itself yet — one thing that will bite on rebase, flagging it early since you'll be rebasing anyway.

**The format version this claims is now taken.** master is at `VERSION_MINOR 96` (`#253`, Save the Echo AI's building-order id, took it). This branch bumps 95 → 96, so after a rebase the number collides with a format that has already been published.

That matters beyond the constant: the gated reads in this PR would then be `versionMinor>=96`, and a save written by master at 96 never wrote those fields — so the reads consume the next values in the stream and the load desynchronises. That is exactly the failure you diagnosed on #232 when master moved out from under it, down to the mechanism.

So on rebase this wants 97, and both the bump and every gate that references it need to move together. #232 is currently sitting on 97 as well, so whichever of the two lands second wants 98.

Worth borrowing the pattern #232 ended up with: a named constant beside master's existing ones in `src/FileFormatVersions.h` rather than a bare `>=96` at each read site, so the next time master moves there is one place to change and a grep that finds it.

Happy to review the substance — the one-worker-level unification — whenever you'd like, rebased or not. The design question in it (dropping the editor's Harvest box, normalising existing maps to the higher of the two levels) doesn't depend on the rebase.

Giszmo:
Rebased onto master and renumbered — thanks for flagging it before it bit. Two adjustments to your note, then what I did.

**It wants 98, not 97.** #232 has landed (95c6a607), so master already holds 97 as `FILE_FORMAT_VERSION_TOPOLOGY_GENERATION`. This branch is the one landing second, so it takes 98.

**The desync mechanism you describe doesn't apply to this PR.** It adds no version-gated reads and no new fields to the save stream. The only serialization it touched was one ungated in-memory normalisation at the end of `Unit::load` — the two halves of the worker level evened out after they were read, not an extra read. So a save written by master at 96 or 97 would not have gone off the rails here; the collision was a wrong number on a replay-refusal marker, not a stream misalignment. The bump exists because simulation behaviour changes for maps that have the two levels apart, which is what `REPLAY_MINIMUM_VERSION_MINOR` is for.

**Took the named-constant point anyway, and gave it something to hold.** `src/FileFormatVersions.h` now has

```cpp
//! Harvest and build became one worker level. Saves before this could hold the
//! two apart — the map editor offered a box for each — so the loader evens a
//! worker out to the higher of the two (UnitSerialization.cpp:90).
static constexpr int FILE_FORMAT_VERSION_ONE_WORKER_LEVEL = 98;
```

and the load-time evening-out is now gated on it rather than running unconditionally:

```cpp
if (versionMinor < FILE_FORMAT_VERSION_ONE_WORKER_LEVEL
    && canLearn[BUILD] && level[HARVEST] != level[BUILD])
    setWorkerLevel(std::max(level[HARVEST], level[BUILD]));
```

That is safe because after this change nothing leaves the two apart in a save at 98 or later: every school tier has `upgrade[BUILD]` and `upgrade[HARVEST]` both set (`BuildingTypesUpgrade.cpp`), the editor's build box now calls `setWorkerLevel`, and the non-parallel upgrade path routes through it too. It also gives the constant a real consumer, so the next time master moves there is one place to change and a grep that finds it.

Also updated `REPLAY_MINIMUM_VERSION_MINOR` to 98 and merged its comment with master's list, moved the `Version.h` changelog entry to 98 alongside master's 96 and 97 entries, and dropped the stale `(cherry picked from commit e5b1cf6a)` trailer since the commit is no longer that content.

Verified locally on the rebased head before pushing: `level-gate-test` passes, `savegame-safety-test` and `buffered-file-test` pass, `team-stats-save-test` passes including the version88 and version84 legacy fixtures (stdout matches), and `check_translations.py --strict` / `test_translations.py` report 0 structural errors. Full CI is running on the pushed head.

The substance is unreviewed and the design question you raised — dropping the editor's Harvest box, normalising existing maps to the higher of the two — is untouched by the rebase, so it is still open whenever you want to look.

genixpro (approved):
Approving. Reading one worker level everywhere is the right fix: hiring, `maxBuildLevel` and the fetch ranking now agree, and school upgrades and the editor both go through `setWorkerLevel`. Raising a legacy mismatch to the higher level never makes a worker worse than the map showed it, and gating that on `FILE_FORMAT_VERSION_ONE_WORKER_LEVEL` keeps it to older files.

Before merging, I merged current master into the branch (`8fa86c609`, no conflicts) and pushed one fix (`e301713ed`). #269 renamed `CORN` to `WHEAT` without an alias, and `LevelGateHarness.cpp` still used `CORN`, so a squash-merge from the 9/12 CI run would have broken master's Linux build. With the rename, `level-gate-test` and `SavegameSafetyHarness` pass locally on the merged tree (macOS arm64). Merging once CI is green on the new head.

Heads-up for ordering: #265 (guard-area crowding) also claims `VERSION_MINOR` 98. This lands first, so #265 moves to 99 with the replay floor.

genixpro:
Master moved again before this could merge: #248 (Cabino) took `VERSION_MINOR` 98. I merged master (`6b028a1b6`) and moved this to **99**: `VERSION_MINOR`, `FILE_FORMAT_VERSION_ONE_WORKER_LEVEL`, and `REPLAY_MINIMUM_VERSION_MINOR` (Cabino was additive and left the replay floor at 97). The load-time evening-out now covers saves below 99, including any 98 save written by a Cabino build. `level-gate-test`, `SavegameSafetyHarness`, `ReplayStepCounterTest` and `TestsRunner` (190) pass locally. Merging once CI is green; #265 moves to 100.

PR #259, opened 2026-09-11
Mark the units a selected building could not hire

Giszmo:
Leo's ask after #255: keep the schooling gate, but show the player *which* units a building could not hire, not just how many.

## What

- **Recording.** While the local player has a building selected, the building keeps the gid of every unit behind each failure tally (`Building::unitsFailingByReason`, filled by `noteUnitFailing` at the same places the counts are taken, reset with them at the start of a hiring pass). `GameGUI::setSelection` / `cleanOldSelection` switch the recording on and off with the selection, so an unselected building records nothing. Display state only: never read by the simulation, not saved, not in the checksum; a client that has a building selected runs the same simulation as one that does not.
- **Shapes.** One outline per reason (`src/gui/FailureShapes.cpp`): square (not available), triangle (too low level), cross (can't access building), circle (too far from building), diamond (can't access resource), inverted triangle (can't access fruit), double ring (too far from resource), plus (too far from fruit). Drawn in red.
- **Panel.** `drawBuildingFailureReasons` draws the reason's shape next to its tally; the text moves right by the shape's width.
- **Map view.** `Game::drawUnit` draws the same shape as a small badge in the top-right corner of the tile of every unit of the local team the selected building remembers under that reason (fist-size: just large enough to tell triangle from square).

No simulation change, no version bump.

## Verification

- `failing-unit-markers-test` harness (new, runs under xvfb in CI, scene uploaded as the `failing-unit-markers-<image>` artifact): hiring pass on a level-2 inn with three unschooled workers and one schooled. Not recording → nothing remembered, count still 3. Recording → the three gids under too-low-level, the schooled one absent. Rendered with the inn selected → red marker pixels in each unschooled worker's tile, none in the schooled worker's. Deselect → lists dropped.
- `TestsRunner` (187).
- Scenes from the harness (also uploaded by CI as the `failing-unit-markers-<runner>` artifact):

Map view with the inn selected: the three unschooled workers wear the too-low-level triangle, the hired schooled worker (bottom left, next to the wheat) does not.

map view, inn selected

Building panel with the inn selected: the tally row carries the same triangle.

building panel, inn selected

Same map with nothing selected, for reference:

map view, nothing selected

## Feel

Only visible with a building selected and only when it has a real obstruction (the rows already had that rule). The map markers are small corner badges with a one-pixel dark halo, same red as the panel; no change while nothing is selected.


genixpro:
Reviewed the code. **No correctness defects found**, and the thing most worth checking — that this doesn't perturb the simulation — holds up under inspection. Three notes below, then the part that needs a maintainer rather than a reviewer.

## The simulation separation is done right

This records inside the hiring pass, which is lockstep simulation code, so it's the obvious place to introduce a desync. It doesn't:

- `noteUnitFailing` increments the tally unconditionally and only the gid push is gated on `recordFailingUnits`. A client with the building selected and one without execute identical simulation.
- Neither `recordFailingUnits` nor `unitsFailingByReason` appears in `Building::checkSum` (checked `Misc.cpp:463` onward).
- Neither is serialized in `Lifecycle.cpp`.
- `Building` is never `memset` anywhere in `src/building/`, which is what makes adding `std::vector` members safe here. Worth stating explicitly since that isn't true of every class in this codebase.

Excluding `UnitNotAvailable` from marking is a good call — badging every busy globule in the colony would be pure noise, and the comment says exactly that.

## Three notes

**1. The map lookup is linear, per unit, per frame.** `Game::drawUnit` walks all 8 reason lists with `std::find` for every visible unit. I haven't measured it, so I'm not claiming it's a problem — but the shape is `visible units x 8 x list length` comparisons every frame on the render path, and `unitsFailingByReason[UnitTooLowLevel]` can hold every unschooled worker in the colony. A late-game 512² map is where that would show. A `flag on Unit` set during the pass, or a sorted list with `binary_search`, makes it constant or log time for a fraction of the code.

**2. The panel and the map can disagree, briefly.** `unitsFailingRequirements` is written to the savegame (`Lifecycle.cpp:520`); `unitsFailingByReason` isn't, and correctly so. But that means after loading a save — and in the window between selecting a building and its next hiring pass — the panel shows a tally with its shape while no unit on the map wears one. It resolves itself on the next pass, but a marker feature that sometimes shows nothing is the kind of thing that makes players stop trusting it. Might be worth suppressing the panel shape while the corresponding list is empty, so the two are never in disagreement.

**3. Recycled gids.** The lists hold gids from the last pass. If a marked unit dies and its gid is reused before the next pass, the badge lands on whatever unit inherited it. Transient and cosmetic, but a validity check against the remembered unit would close it.

## For a maintainer

The design question is the shape vocabulary: 7 distinct outlines on the map (8 in the panel, counting the never-marked "not available"). They only appear with a building selected and genuinely obstructed, and the panel always shows the same shape beside its label, so there's a legend in context — that's a sensible containment. Whether seven outline shapes actually read at a glance in play, or whether this wants fewer buckets or a different encoding, is a judgement call that needs someone playing it, and per AGENTS.md a UI feature needs explicit maintainer approval regardless.

Note this branch predates #252, #250, #253 and #262 and will want a merge of master before it goes in.

genixpro:
**Correcting my earlier review**, which was written against e38e57198 and is now out of date in its most visible part: it describes seven distinct outline shapes and raises the vocabulary burden as the open design question. That design is gone. `d6db54cee` draws one badge for every reason, and the argument in the code is better than the one I was going to make:

> several reasons can apply to the same unit at once with only the first one reported, so telling them apart by shape would be a promise the engine cannot keep

That's a correctness argument, not just a legibility one, and it settles the question rather than answering it. I verified it empirically — drew all eight reasons at the real map size and the pixels are byte-identical across every one.

The single badge was reviewed on screen at real size, over grass, sand, water and a red team unit, and reads clearly; the white border is doing the work on red units, which is exactly the case it was added for. Approving on that basis.

I also read the rest of the delta since my review. `shouldShowFailingUnitMarkers` is the right shape of fix for the panel-vs-map disagreement I raised: one gate asked by both sites, so a unit can never wear a badge the panel has no row to explain. The antialiasing rewrite is clean — run-merged spans, a correct sentinel to flush the final run, standard coverage AA, no domain hazards.

## Follow-ups, not blockers

Filing these rather than holding the feature: all four are display-only, with no correctness, lockstep or save-format exposure.

**1. The most common reason marks units you often cannot see.** This is the one I'd most want you to think about. I ran a real 128x128 four-AI game out to ~13,000 ticks and captured the six buildings that were genuinely failing to hire. **All six failed for the same reason: "too far from building"** — which by definition puts the marked unit away from the building you just clicked. In one of the six, the pixel diff between "nothing selected" and "building selected" was exactly **zero**: the marked unit was off-screen entirely.

With distinct shapes gone, the panel text is now the only thing that says *why*, so a badge you cannot find leaves the player back at a bare count — and worse, now hunting for a marker that is not on screen. An edge-of-view indicator, or a panel row that says the marked units are off-view, would close it.

**2. Linear lookup per unit per frame.** `Game::drawUnit` still walks all eight lists with `std::find` for every visible unit. Now short-circuited by the new gate, so it only costs when a building is selected *and* obstructed *and* still asking — but the inner cost is unchanged, and `unitsFailingByReason[UnitTooLowLevel]` can hold every unschooled worker in the colony. Unmeasured; a flag on Unit would make it constant-time.

**3. Recycled gids.** The lists hold gids from the last scan; if a marked unit dies and its gid is reused before the next one, the badge lands on whatever inherited it.

**4. Tallies are saved, gid lists are not.** So just after loading a savegame the panel can show a row while no unit wears a badge, until the next hiring pass repopulates. The new shared gate does not cover this one, because the gate reads the tallies, which survive the load.

PR #261, opened 2026-09-12
Repeat a small map over the torus to fill a large one

Giszmo:
Repeat a small map over the torus until it fills a large one. A 64 x 64 map
repeated 2 x 2 becomes a 128 x 128 map with four copies of the landscape and
one colony per copy, each dealt to a fresh team. The editor now also offers
32 x 32 as a map size, which is the natural input for repeating.

This is Bob's and Junior's `feat/map-tiling` (tip `6fdb347`, last touched
2026-09-06) rebased onto current master. The branch was never proposed; master
has moved 53 commits since it forked.

### Changes

- `Map::tile()` repeats the terrain, discovered mask and undermap arrays over the
  torus, dropping the per-team state that is rebuilt afterwards (`src/map/MapTile.cpp`).
- `Game::tileForPlay()` deals each repeated copy's buildings and units to a fresh
  team, wrapping every colony so a base stays together across the seam
  (`src/Game_tiling.cpp`, `src/MapTiling.cpp`).
- The map chooser gains tile x / tile y / colonies / bases controls, previews the
  repeated map rather than the source, and paints buildings and units in their
  team colour. Choosing Ok writes the repeated map to the temp directory, so tiled
  maps never appear in the maps folder.
- The editor and the custom-game generator offer 32 as a width and height. That is
  one shared control definition on master now, so the new minimum applies to both.
- `-dump-tiled <map> <rx> <ry> <colonies> <bases>` gives a headless check of a
  repeated map's colonies.

### What the rebase had to change

Master renamed a lot of what this branch calls:

- `FormatableString` -> `FormattableString`, `isRessourceTakeable` -> `isResourceTakeable`,
  `Building::ressources` -> `resources`, `MAX_NB_RESSOURCES` -> `MAX_NB_RESOURCES`,
  `Unit::hungryness` -> `hungriness`.
- `Map::cases` -> `Map::tiles` and `Case` -> `Tile` (PR 199).
- `Team::setBaseTeam()` was removed as dead code by PR 190. Its five lines are
  inlined at the one call site that wanted them.
- The editor's size widgets were replaced by `MapGenerationDescriptor::sharedControls()`
  in PR 238, so "allow 32 x 32" is now a minimum of 5 on the `wDec`/`hDec` controls
  instead of an extra `add(32)`.

Each of the twelve commits was adapted in place, so every commit on the branch builds.

### Screenshots

Editor create-map dialog, 32 x 32 now selectable:

Editor 32 x 32

Map chooser, "balanced for 2" at 1 x 1 — 2 bases, 64 x 64:

Chooser 1 x 1

The same map at 2 x 2 — 8 bases, 128 x 128 (64 x 64), preview shows the repeat:

Chooser 2 x 2

The repeated map loaded in the editor; the minimap shows the four copies:

Editor with the tiled map

The screenshot files live in commit `e8c8229e0` and are removed again at the branch tip.

### Validation

- `scons release=1` clean, no warnings introduced.
- `test/TestsRunner`: OK (187 tests). `test/WinningConditionsHarness`: exit 0.
- `./glob2 -dump-tiled maps/balanced_for_2.map 2 2 4 1` reports `128x128, teams=4`
  with each colony's units and buildings wrapped to its own quarter.
- Screenshots above were driven through the real GUI under Xvfb, not mocked up.
- The rebase's first attempt silently broke the translation files: stripping the
  conflict markers also removed a literal `<<<<<<<` line that lives inside
  `data/texts.br.txt` on master, and left a stray blank line in front of the
  appended block in fourteen files. Both desynchronise `StringTable`'s key/value
  pairing and the game asserts on startup. Fixed in every commit, and a pairing
  check now passes at each of the twelve.

### After #240: merge, translations, and review follow-ups

Three commits on top, after #240 landed:

- **Merge master.** The catalogs take master's generator strings, then the tiling block.
  `MapThumbnail` keeps master's `loadFromMap(const Map&)` beside the team-coloured
  overload (both draw through one `render`), `dumpResources` follows the `CORN` → `WHEAT`
  rename. The 32 x 32 minimum no longer lives in the deleted `MapGenerationDescriptor.cpp`.
  Since #240 the size controls are shared by the editor and the custom-game lobby, and
  lowering them put 32 x 32 in the lobby too. There most landscapes cannot seat colonies,
  and CI's setup harness failed on it. So **only the editor offers 32 x 32**:
  `editorSizeControl` lowers the minimum on the new-map screen and in request validation,
  and switching landscapes keeps the size. The lobby is unchanged from master.
- **Translations.** The seven tiling strings are translated in all 32 catalogs.
  `test_translations.py` and `check_translations.py --strict` pass. "bases" is added to the
  shared vocabulary for br, ca, es, fr, nl and pt, where the word is the same as in English.
- **Painted areas and clearing flags survive the repeat.** `Map::tile` clears the per-team
  forbidden, guard and clearing masks, and `tileForPlay` never painted them again, so a
  map's hand-drawn zones vanished from every copy. Each copy now paints its colony's areas
  for the team the colony is dealt to, and clearing flags keep their resource choices. New
  `MapTilingHarness` (`map-tiling-test`, in CI) checks areas, flag choice, buildings and
  units per copy. Without the fix it fails. It also checks the editor-only 32 x 32 size, and
  that such a map generates.
- **LAN transfer of a tiled map.** A tiled map exists only in the temp directory, behind
  `MapHeader`'s file name override, and the override is not serialised. The LAN session
  test gained `--tiled` (in CI): the host tiles FourSquares1 2 x 2, and the guest must
  download it and match the bytes exactly. It passes over two join/leave cycles. The guest
  stores the map under `maps/`, so on the joining side a tiled map does appear in the
  user's map list. The 3 MB map takes most of a minute to transfer, so tiled mode stays in
  the lobby longer.

What a repeat still drops, by design or as a known limit: units inside buildings, a
building's in-progress upgrade, and each team's local resource counters. Only `ALIVE`
buildings and units standing on the map are copied.

### Verified locally (macOS arm64)

- Full client build.
- `MapGeneratorDefaultsTest`, `MapGeneratorGoldenTest --require-rows` (248 rows, 0 failures)
  and `CustomGameSetupHarness` all pass.
- `MapTilingHarness` passes, and fails with the painted-area fix reverted.
- LAN session test passes in plain and `--tiled` modes.
- A full 90,000-tick headless game on the tiled FourSquares1 2x2 map (Nicowar, Cabino,
  Castor, Warrush) ran to the end without errors.
- Every generator at 32 x 32 (1, 2 and 4 colonies, 5 seeds each) either refuses the request
  up front or fails with a diagnostic. None crashes. Some fail for every seed at a given
  colony count: isles with 1 colony, fjord-continent and concrete-islands with 2 or 4,
  contested-commons with 4. In the editor, those combinations show the generator's
  message.

The simulation, save format, replay and network versions are unchanged. Tiling only builds
a new map file, which then loads and plays like any other map.


genixpro:
@Giszmo, we like this one. Plan: #240 (map generators) is about to land and touches the same shared width/height controls and map-setup screens, so we'll wait for it, then take the rebase and fixes on our side. No need for Bob to spend time on it meanwhile.

What we'll cover once #240 is in:
- merge master, including the `CORN` → `WHEAT` rename (#269). The new thumbnail code in `MapThumbnail.cpp` still uses `CORN` and won't compile on master as is;
- translations for the seven new strings, so `test_translations.py` passes;
- a look at what the repeat drops: painted forbidden, guard and clearing areas, flag settings beyond `minLevelToFlag`, units inside buildings;
- checking that a client joining a LAN/YOG game on a repeated map gets the file, since the temp-path override isn't serialised.

On authorship: no rewrite needed. We'll squash-merge, so the commit on master is authored by you, and we'll drop the agent `Co-authored-by` lines in the merge message.

PR #263, opened 2026-09-12
Rasterise interface text for the screen it is drawn on

idevchris:
Text renders pixelated because glyphs are rasterised once, at the fixed pixel
size the font is loaded with (20, 13 and 10 px), and that finished bitmap is
then resampled along with the rest of the frame to whatever size it occupies on
screen. Nothing ever re-renders text for the screen it is actually drawn on, so
at 200 % interface scale one glyph pixel becomes a 2x2 block of screen pixels.

Nor is a raised interface scale the only case. At 100 % it happens too whenever
the display cannot deliver the requested resolution exactly: a fullscreen
3456x2234 on a 16-inch MacBook is presented at 3456x2168, so every glyph is
reduced by 3 % and loses pixel rows.

### Before and after

| before | after |
| --- | --- |
| Screenshot 2026-09-12 at 16 47 14| Screenshot 2026-09-12 at 16 48 10|

### What this does

`TrueTypeFont` keeps a second font handle at `size * scale` and draws from it,
while every reported metric still comes from the authored size, so no widget
moves. The scale is the new `GraphicContext::textRenderScale()`, which is 1 for
the software renderer. Hinting rounds each size differently ("Editor" is 61
logical pixels at size 20 but 118 device pixels at size 40, i.e. 59), so the
raster is drawn at its own size rather than stretched into the measured
rectangle; stretching repeats or drops pixel columns inside single glyphs.

Two fixes it depends on: a new GL context now resets the cached GL state, and the
string caches drop themselves when the context generation changes. Without them
the first draws after a resolution or scale change fill white quads from texture
names of the destroyed context.

`test/TextRasterHarness.cpp`, built and run in CI under Xvfb, requires unchanged
string metrics across a scale change, detail inside the scale-sized pixel blocks,
glyphs matching an independently rasterised reference at a magnifying, a
fractional and a reducing scale, and overlay text still reaching the screen. Each
check fails against the implementation it guards. 187 unit tests and a 600-tick
headless run (checksum `6b785f48`) are unchanged; simulation, saves, replays and
the protocol are untouched.

### What it does not cover

`OverlayScreen` composes itself on its own logical-size surface that is resampled
as one image, so the in-game menu, load/save, message boxes and the editor
dialogs look exactly as before. `DrawableSurface` has no scaled blit
(`// TODO : Implement`), so handing them a raster made for the screen would have
made their text vanish instead. Sharpening them needs either overlay widgets
drawn straight onto the context, or a device-resolution backing for those
surfaces plus that blit: a separate change across all widget drawing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -221,6 +221,14 @@ jobs:
           timeout 60s xvfb-run -a -s '-screen 0 1600x1400x24' ./build/libgag/src/WindowResizeHarness gl
           timeout 60s xvfb-run -a -s '-screen 0 1600x1400x24' ./build/libgag/src/WindowResizeHarness software
 
+      - name: Build and run the scaled text regression
+        env:
+          LIBGL_ALWAYS_SOFTWARE: 1
+        run: |
+          scons -j$(nproc) release=1 server=0 text-raster-test
+          mkdir -p artifacts/text-raster
+          timeout 60s xvfb-run -a -s '-screen 0 1600x1400x24' ./build/libgag/src/TextRasterHarness artifacts/text-raster
+
       - name: Check game rendering across repeated map copies
         env:
           SDL_VIDEODRIVER: dummy
--- a/libgag/include/SDLGraphicContext.h
+++ b/libgag/include/SDLGraphicContext.h
@@ -417,6 +417,8 @@ namespace GAGCore
 		float getUiScale(void) const { return uiScale; }
 		//! the scale the last setRes() was asked for, before the window floor reduced it
 		float getWantedUiScale(void) const { return wantedUiScale; }
+		//! drawable pixels per logical pixel for text drawn on this context, 1 when unscaled
+		float textRenderScale(void);
 		//! the interface scale the desktop asks for, or 0 when nothing reports one
 		static float querySystemUiScale(void);
 		//! the scale actually used for a preference; 0 follows the desktop, GLOB2_UI_SCALE wins
--- a/libgag/include/TrueTypeFont.h
+++ b/libgag/include/TrueTypeFont.h
@@ -38,38 +38,76 @@ namespace GAGCore
 		virtual Style getStyle(void) const;
 		
 	protected:
+		struct CacheKey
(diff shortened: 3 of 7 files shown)


genixpro (approved):
Thanks @idevchris, and welcome! This is a really nice first contribution: a clear diagnosis, a fix that keeps every reported metric at the authored size so no widget moves, and a harness that fails against the implementation it guards. The two underlying fixes, resetting the cached GL state on a new context and dropping string caches on a context-generation change, are good catches in their own right.

**One fix pushed to your branch** (`5cdd355cc`, one file). The harness didn't compile on Linux. `GraphicContextPrivate.h` pulls in the GL headers, which bring `X11/X.h` with its `typedef XID Font`, so `Font` under `using namespace GAGCore` was ambiguous. It now names `GAGCore::Font` explicitly. macOS has no X11, which is why it built for you. CI's Linux jobs would have stopped there, so I fixed it rather than send it back.

**Verified**, with the PR merged onto current master (which now includes #275's `rasterScale()` in the same files; no conflicts):
- macOS arm64, Apple M3, real GPU: `TextRasterHarness` PASS at scale 4; `torus-render-test` at `GLOB2_UI_SCALE=3` PASS; `HighResolutionIntegrationHarness` PASS.
- Ubuntu 24.04 x86_64, Mesa software GL under Xvfb, with the fix: `TextRasterHarness` PASS at scale 2; `WindowResizeHarness gl` and `software` PASS.
- The before/after raster from the harness is clearly sharper, with glyph edges on device pixels instead of 2×2 blocks.

Non-blocking thoughts for later, none needed for this PR:
- `targetScalesText()` treats every draw on the screen context as scaled by `drawableScale()`. Text drawn under the 2D map transform (zoom, now 25–300% after #280) or into the torus atlas is also resampled by `mapScale` or by the atlas's own scale. #275 added `GraphicContext::rasterScale()` for exactly that "pixels of the current target" question, so folding it in could sharpen map-space text too.
- As you noted, `OverlayScreen` dialogs keep the old look until they get a device-resolution backing.

PR #265, opened 2026-09-12
Spread free warriors between guard areas by crowding

genixpro:
## Why

Guard areas were only half usable. Painting one told the warriors where to go, but not how many would go: every free warrior walked to the nearest painted area whatever was already there, so a batch trained in one place all took the same area and any other area stayed empty. How many warriors a defence ended up with depended on where the barracks stood, not on what the player painted, which left the war flag as the only dependable way to station warriors.

This makes guard areas mean what a player expects: warriors spread between painted areas, a bigger area gets more of them, and an area that is over-full thins out into the others.

| Before: everyone takes the near area | After: both areas are held |
| --- | --- |
| before | after |

24 warriors spawned at the top left, a 5x5 area 10 tiles east and another 16 tiles south, 4,500 ticks later, rendered through the real map renderer by the new harness.

## How

Each painted guard tile is seeded with a crowding cost instead of zero: the team's warriors within eight tiles, per painted tile within eight tiles, at four tiles of walking per warrior. A crowded area then reads as farther away than it is, and Dijkstra gives every tile the best of "distance to an area plus that area's crowding". Full design, tuning knobs and what was tried and dropped: `docs/guard-area-balancing/README.md`.

- Both counts are separable row-major box sums, linear in map size whatever the radius or warrior count, and only run for a team that has paint and warriors. The rebuild schedule is unchanged.
- Inside an over-full area another area's field runs over the painted tiles. A warrior there follows it out on one action in 64, so the area drains as a trickle rather than all at once; a leaver keeps counting toward the area for eight tiles, which is more than the one warrior's cost it freed, so it does not turn back.
- A random step taken on a painted tile prefers a free painted neighbour, since stepping off whenever it could is what would empty a packed area in one go. When no painted neighbour is free (a packed, sparse, checkerboard or 1x1 area) the warrior takes the ordinary random step onto any free tile, as before this change, rather than standing with no direction, which both renderers draw as a unit spinning in place. Guard areas are rarely painted exactly, so warriors standing beside the paint still defend it.
- Nothing new is saved. The crowding grids are scratch and the guard field was already saved with its refresh flag.

| Drain: the near area is full when the far one is painted | After, 1,000 ticks later | After, 4,000 ticks later |
| --- | --- | --- |
| before (old engine, 4,000 ticks later: nobody moved) | after 1000 | after 4000 |

## Verification

`test/GuardAreaBalanceHarness.cpp` (target `guard-area-balance-test`, wired into CI, frames uploaded as an artifact) runs the real engine on a blank map with 24 warriors. Warriors at near / far after 4,500 ticks, at the three guard-field refresh cadences a game can have:

| Scenario | Before | After (every tick / ~10 / ~170 ticks) |
| --- | --- | --- |
| spawn: both areas painted, warriors arrive | 24 / 0 | 13 / 11, 13 / 11, 12 / 12 |
| drain: 24 already inside near, far painted later | 24 / 0 | 14 / 10, 15 / 9, 15 / 9 |
| patches: near is two unconnected 3x3 patches | 24 / 0 | 11 / 13, 12 / 12, 9 / 15 |
| size: near 5x5, far 9x9 | 24 / 0 | 4 / 20, 5 / 19, 10 / 14 |
| three areas at 18, 32 and 45 tiles | 24 / 0 / 0 | 5 / 10 / 9 |
| erase: far erased after settling | | 24 / 0 |
| saveload: save at tick 1,500, load, 1,000 more ticks | | identical checksums every tick |
| spins: settled guards, share of on-paint ticks with no direction | 1x1 100%, sparse 100%, checker 81–85%, packed 5x5 72–74% | 1x1 0–24%, sparse 0%, checker 0%, packed 5x5 0.6–2.5% (seeds 1–10) |

The balance rows above were recorded before the random-step fallback. With it, at seed 1, spawn ends 13 / 11, 13 / 11 and 11 / 13 at the three cadences and drain 11 / 13 at all three. What remains on a 1x1 is the one guard boxed in on all eight sides by the crowd pressing round its single tile.

**Accepted trade-off: a small drain overshoot.** A warrior with no free painted neighbour now steps off the paint, where the guard field may send it on to the other area, so the first area in `drain` dips 1–5 of 24 warriors below where it settles before they come back (mean 2.4, seeds 1–10 at all three cadences; 0 on every seed without the fallback). The `drain` check now guards against the herd effect (an area emptying out and refilling) at a quarter of the warriors, and all scenarios pass on all 30 seeded runs.

**The harness is now seeded** (`--seed N`, default 1). It never set the game's random seed, and `GameHeader` defaults to the wall clock, so earlier runs of the same binary differed from one second to the next, and the old "never overshoots" result was the luck of those seeds.

The drain never overshoots (the first area's lowest count is its final count) and in-area wandering is unchanged. The crowding box sum is checked against a brute-force count across the torus seam.

Cost, median of 200 guard-gradient rebuilds: 64x64 map 31 to 32 us before, 36 us after; 256x256 map 444 to 510 us before, 529 to 582 us after, of which the two box sums are 5 us and 73 to 90 us. Full simulation step with 24 warriors unchanged (16.0 vs 15.8 us); with 96 warriors it drops from 182 to 70 us because fewer blocked warriors hit the pre-existing full rebuild on retry. A 20,000-tick headless Warrush game takes 2.09 s before and 2.00 to 2.06 s after.

Cross-platform, on the head merged with current master (`45ca3dec2`, includes #256 and Cabino; later commits touch only the harness and docs): the same recipe below gives byte-identical sidecars on macOS arm64 (twice) and Ubuntu 24.04 x86_64 / g++ 13.3, md5 `2de858478c20786c3adc09ac4cbd6534`, replay md5 `ecc7e95fb2ab6c81aee420d897fbd0a3`.

Before that merge (`e1c7886ac`, with the fallback): a tick-0 Warrush-vs-Warrush state on SmallForTwo (`GLOB2_TEST_SEED=42 glob2 -test-games-nox 1 --map SmallForTwo --matchup warrush,warrush --save-game-as <file>`) run with `GLOB2_CHECKSUM_SIDECAR=1 glob2 --nox <file> 20000 1` on macOS arm64 (Apple clang, twice) and Ubuntu 24.04 x86_64 (g++ 13.3) gives byte-identical per-tick checksum sidecars (md5 `4ba30e8e1db66857a7557f9e330b09df`) and replays (md5 `b85e04f59456740285a434defb320f7e`). The same state on the pre-fallback engine ends at tick 12,739 with a different sidecar, so this game exercises the new step.

Earlier, before the fallback: the same tick-0 state run for 20,000 ticks with `GLOB2_CHECKSUM_SIDECAR=1` on macOS arm64 (Apple clang) and Ubuntu x86_64 (g++ 14) gives byte-identical 188 MB per-tick checksum files (md5 `00f7fa38804129c9e786ece0089f986e`), and the harness prints identical tables and passes on both.

Also run: CppUnit suite (190 tests), immobile-gradient, inn-swap, resource-fetch, fetch-apportionment, hiring-bucket and savegame-safety harnesses, client and server builds.

## Compatibility

No save-format change. The simulation changed, so `VERSION_MINOR` 100, `REPLAY_MINIMUM_VERSION_MINOR` 100, `NET_PROTOCOL_VERSION` 30 and the YOG floor 30 refuse older replays and clients, as for #211. (98 went to Cabino, #248; 99 to one worker level, #256.)

## What to look at in play

- A batch trained together reaches its first area later than before, since the moving crowd counts against the area it approaches and part of it diverts.
- A packed area keeps its extra warriors milling on the tiles beside it rather than frozen in place; guards on a small or sparse area keep moving too.
- While an area drains into a newly painted one, a few warriors overshoot to the new area and drift back.
- An over-full area empties over a couple of thousand ticks, not at once.
- Warrush paints guard areas on enemy buildings it has found; its attack now spreads across them instead of massing on the nearest.
- In the SmallForTwo Warrush-vs-Warrush game above, the fallback alone moves the end from tick 12,739 to past 20,000: Warrush's guard-area attacks behave differently enough to change the match.
- Shares follow painted size roughly, by design: the equilibrium is a band as wide as the distance between areas divided by four tiles per warrior. An exact split would have warriors walking between areas to correct a difference of one.

https://claude.ai/code/session_018dsZ88QoF1D2U68nuP4tsB


Giszmo:
I reviewed this at 514543b by building the branch and running your harness, plus seven scenarios I added to it for the shapes the harness doesn't currently cover. The mechanism is simple and most of it holds up — all eight of your regression scenarios pass here on Linux x86_64 and the headline 13/11 reproduces exactly. Three things need fixing first.

**1. Guards spin in place.** `keepInGuardArea` leaves a warrior with no free painted neighbour at `count == 0`, which sets `direction = 8`. Both renderers animate that by cycling all eight facings, so a crowded guard area now looks like a field of entrapped units. A full 5×5 does it 73% of the time; the sparse brush does it 100% of the time and those warriors never move again.

**2. A war flag beside a guard area drains it.** `computeWarriorCrowding` counts every warrior not inside a building, including a flag's garrison, which is not guarding anything. A flag 6 tiles from one of two areas moves the split from 13/11 to 8/16.

**3. Small areas saturate the cost cap** and stop balancing entirely. Not a bug so much as an undocumented limit, but it is reachable with brush 0.

Plus the mechanical ones: the branch is `CONFLICTING` against master, and no CI run exists for it yet.

Details below.

---

## 1. Guards spin in place

`direction = 8` is "no direction" (`Unit.h:226`). Both renderers draw it as `imgid += 8*(delta>>5)` (`GameRenderUnits.cpp:91`, `GameGUIDrawUnitInfos.cpp:80`) — all eight facings cycled as `delta` advances — and `Unit::step` advances `delta` every tick whether or not the unit is displaced. That is the animation for a unit entrapped in resources.

Before this PR a warrior in a packed area took its random step onto an unpainted neighbour. With `keepInGuardArea` those neighbours are filtered out, so when every *painted* neighbour is occupied the loop ends at `count == 0` and falls into the pre-existing `dx=dy=0, direction=8` branch. That branch used to be rare. It is now the normal state of a crowded area.

Warriors standing on paint, sampled over 500 ticks after the areas have settled (24 warriors, `--resource-gradients`, one 64×64 team):

| Paint | Guards | Never moved | Ticks at direction 8 | Moves/guard |
| --- | --- | --- | --- | --- |
| 1×1 (brush 0) | 1 | 1 | **100%** | 0.0 |
| sparse (brush 5), 25 tiles over 9×9 | 24 | 24 | **100%** | 0.0 |
| checkerboard (brush 4), 25 tiles over 7×7 | 24 | 1 | 79% | 4.5 |
| solid 5×5, 24 warriors | 24 | 0 | **73%** | 7.2 |
| solid 5×5, 8 warriors | 8 | 0 | 0% | 23.5 |
| solid 9×9, 24 warriors | 24 | 0 | 0.1% | 24.2 |

The same build with `keepInGuardArea` forced false gives **0.0% at direction 8 on every row**, so the rule is the whole cause.

None of these rows is contrived. Brushes 4 and 5 are the checkerboard and quarter-density area brushes a player picks in the UI, and your own `erase` scenario ends with 24 warriors on a single 5×5 — the row that spins 73% of the time.

The rule is earning its place, though: with it disabled, nobody stays on a 1×1 or a sparse area at all, which is the drain-all-at-once problem it was written for. So the fix isn't to drop it but to give it a fallback other than "no direction" — allow the unrestricted step when no painted neighbour is free, or route that case through the 1-in-64 leave path. I didn't want to pick for you.

A related note on the same rule: it keys off `tiles[].guardArea` only, so a warrior attached to a war flag that happens to stand on painted ground is confined to the paint rather than to the flag's `unitStayRange`.

## 2. A war flag beside a guard area drains it

`computeWarriorCrowding` excludes only `DIS_INSIDE`, so a war flag's garrison, warriors passing through and warriors fighting all price a nearby area as crowded.

Two 5×5 areas ("near" 18 tiles from the spawn block, "far" 32 tiles), 24 free warriors:

| Setup | near | far | near seed cost |
| --- | --- | --- | --- |
| no flag | 13 | 11 | 52 tiles |
| war flag 6 tiles from near, holding 12 warriors | **8** | **16** | 62 tiles |

The near area loses a third of its guards to a flag that isn't guarding it. That cuts against the stated goal that what a defence gets should follow what the player painted — a player who parks a flag next to their wall silently loses the wall's garrison. Skipping units with `attachedBuilding` in the gather loop looks like it would cover it, but it's your call whether fighting and in-transit warriors should count too.

## 3. Small areas saturate the cap

The seed cost is `GUARD_CROWD_COST_PER_WARRIOR × W × 25 / P`, i.e. 4 × W × 25 / P tiles, capped at 400. The cap is reached at `W/P ≥ 4`: 100 warriors for a 5×5, 36 for a 3×3, and **4 for a 1×1**. Past the cap every area reads the same and the field is a plain nearest-area partition again.

Two 1×1 areas, 24 warriors: one warrior parked on each painted tile, both seeds pinned at 400, and the remaining 22 milling in two blobs split 6/18 — arbitrary rather than balanced. No oscillation, though; I tracked on-paint/off-paint transitions and got exactly one per warrior, i.e. arrival only.

I'd suggest the tuning table in `docs/guard-area-balancing/README.md` say where the mechanism stops working, since `GUARD_CROWD_COST_MAX` is currently documented only as "keeps seeds above the unreachable sentinel".

## Behaviour worth documenting: the field no longer plateaus on the paint

Before this change every painted tile sat at `GRADIENT_AT_GOAL` and nothing else did. Now:

| Case | Tiles at AT_GOAL | Painted tiles |
| --- | --- | --- |
| painted, team has no warriors | 25 of 25 | flat at 0 cost (crowding pass skipped) |
| one 5×5, 24 warriors on it | **0** | flat at 96 tiles of cost |
| over-full area + a freshly painted second one | 25 (the new one) | 0 to 37 tiles of cost; an unpainted tile beats the worst painted tile |

Flat within an area only because anything up to 9×9 in extent fits inside the radius-8 window and so counts the same paint everywhere; a larger area has fewer painted neighbours at its edge than at its centre and therefore slopes inward. That is probably fine — warriors prefer the interior — but it is a real behavioural difference for big painted regions and isn't mentioned in the design note.

`UnitMovement.cpp` was correctly updated for the lost plateau. I grepped every reader of `guardAreasGradient` and that was the only place that needed it. Worth knowing that `getGlobalGradientDestination` now always returns `false` for this field, since it only reports an exact destination when it reaches `0xFFFF`; the one guard-area caller discards the return value, so nothing is broken today.

## What holds up

- All eight regression scenarios pass on Linux x86_64 at the default cadence: spawn, drain, patches, size, three, erase, saveload, crowding. The 13/11 headline reproduces.
- A checkerboard or sparse area of 25 painted tiles splits against a solid 5×5 at 13/11, the same as two solid 5×5s — the share follows the painted-tile count rather than the bounding box, which is the right answer.
- The box sum is O(map) and radius-independent, and it is correctly skipped for teams with no paint or no warriors.
- The `VERSION_MINOR` / replay floor / `NET_PROTOCOL_VERSION` / YOG floor bumps are the right kind: this genuinely changes the simulation, including the `syncRand()` draw added per warrior standing on paint.
- Nothing new is saved, and the save/load continuation check covers it.

The engine change is 122 non-comment lines across six files. It is a simple mechanism and I don't think it is implemented weirdly; the three items above are all local to the new code.

## Reproducing

Scenarios were added to `test/GuardAreaBalanceHarness.cpp` on top of 514543b and run as `report --ticks 4500 --resource-gradients`:

- `spins` — for each paint layout, settle, then sample 500 ticks and report per-guard `direction == 8` ticks and tile moves.
- `divert` — two 5×5 areas with and without a war flag 6 tiles from "near", `unitStayRange` 2, `maxUnitWorking` 12.
- `plateau` — dump tiles at `GRADIENT_AT_GOAL`, the painted range and the best unpainted tile.
- `tiny` / `tiny1` — one and two 1×1 areas, with on-paint/off-paint transition counts.
- `checker` / `sparse` — brush 4 and brush 5 patterns against a solid 5×5.

Happy to push these to a branch if they'd be useful in the harness proper. The A/B was done with a local switch on `keepInGuardArea` in `pathfindRandom`, not a code change I'm proposing.

Giszmo:
If I understand my agent Carol correctly, it considers it a bug that a war flag drains the closest guard area first. I think that's a feature, not a bug, so disregard number 2.

genixpro:
@Giszmo — thanks for the thorough review. At `e1c7886ac`:

**1. Guards spin in place — fixed, with your first option.** When a warrior on paint has no free painted neighbour, `pathfindRandom` now falls back to the ordinary random step onto any free tile, as before this PR. A free painted neighbour is still preferred, so an area with room keeps its guards on the paint. We went this way over holding them still because players don't paint their defences exactly: warriors standing beside the area still defend it, and it's closest to how the game already felt.

I added your `spins` measurement to the harness as an asserting scenario: 500 ticks sampled after 4,000, 24 warriors.

| Paint | No direction, before → after | Held on one tile all 500 ticks, before → after |
| --- | --- | --- |
| 1x1 | 100% → 10–11% | 1 → 0 |
| sparse 9x9 (brush 5) | 100% → 0% | 24 → 0 |
| checkerboard 7x7 (brush 4) | 81–85% → 0% | → 0 |
| solid 5x5 | 72–74% → 1% | → 0 |
| solid 5x5, 8 warriors / solid 9x9 | ~0% → 0% | 0 |

The before column reproduces your table. The check fails on the previous code and passes on this one. What's left on a 1x1 is the single guard boxed in on all eight sides by the crowd round it. A unit with no free tile at all has always stood still, and that isn't new. Every layout still keeps its warriors within 7 tiles (22–24 of 24).

The balance scenarios all still pass at all three cadences. Spawn ends 13/11, 13/11 and 10/14; drain ends 11/13 with no overshoot.

**2. War flag draining — left as is**, per your follow-up.

**3. The cap — documented.** The tuning table in `docs/guard-area-balancing/README.md` now says where `GUARD_CROWD_COST_MAX` stops balancing: four nearby warriors per nearby painted tile, which is 100 for a 5x5, 36 for a 3x3 and 4 for a single tile. Past that point the field is a plain nearest-area partition again. I also added your plateau observation: tiles at `AT_GOAL`, areas up to 9x9 staying flat, larger ones sloping inward, and `getGlobalGradientDestination` no longer reporting an exact destination.

**Mechanical.** Merged master (clean apart from version numbers). Renumbered to `VERSION_MINOR` 98, with the replay floor at 98, because #232 took 97. #256 also wants 98, so whichever lands second moves.

**Cross-platform, on this head.** A tick-0 SmallForTwo Warrush-vs-Warrush state run for 20,000 ticks with `GLOB2_CHECKSUM_SIDECAR=1` gives byte-identical sidecars on macOS arm64 (twice) and Ubuntu 24.04 x86_64 / g++ 13.3 (md5 `4ba30e8e1db66857a7557f9e330b09df`). The same state on the pre-fallback engine ends at tick 12,739 instead, so that game does exercise the new step. That's also a feel point worth looking at in play: Warrush's guard-area attacks change enough to change the match.

**One thing I found on the way that isn't this PR.** The harness's blank in-code map reaches different states on repeated runs of the same binary when it's stepped without periodic checksumming. It does the same with this harness's setup built against `origin/master`, so it's pre-existing, and the real-game path above is byte-identical run to run. It moves the harness counts by a warrior or two, and I'm tracing it separately.

Also passing: CppUnit (190), `ReplayStepCounterTest`, `SavegameSafetyHarness`, client and server builds. CI is running.

Since this changes where warriors go in play, I'd appreciate your approval (or anything that still stands) before it merges.

genixpro:
@Giszmo @kylelutze @stephanemagnenat, this is ready for review and, more importantly, for playtesting. We'd like other players to try it before it merges. It changes where warriors go in every game with guard areas.

Since the last round (`d34da4bfa`):

- **Merged current master** and renumbered to `VERSION_MINOR` 100 with the replay floor, since Cabino took 98 and one worker level took 99. `NET_PROTOCOL_VERSION` and the YOG floor stay at 30.
- **Mac/Linux equivalence re-checked on the merged head.** A 20,000-tick Warrush-vs-Warrush game gives byte-identical per-tick sidecars on macOS arm64 and Ubuntu x86_64.
- **The harness is now seeded.** It never set the game's random seed, so runs followed the wall clock. That hid something worth knowing: with the random-step fallback that fixed the spinning guards, an area draining into a newly painted one **dips 1–5 of 24 warriors below where it settles, and they come back** (seeds 1–10, all cadences; never without the fallback). We've accepted that. It's small and it settles, and the `drain` check now guards only against the herd effect.

What's most useful to look at in play:

1. Paint two guard areas at different distances and watch the warriors spread out.
2. Paint a second area next to a full one: it should thin out over a couple of thousand ticks, with a few warriors overshooting and drifting back.
3. Paint small, sparse or checkerboard areas: guards should keep moving, not spin, and some will stand just beside the paint.
4. Play against Warrush, whose attacks now spread across the guard areas it paints on your buildings.

Please say if anything feels off.

PR #264, opened 2026-09-12
Add Maxima: a standalone AI that develops a colony and attacks relentlessly

genixpro:
Add Maxima: a standalone AI that develops a colony and attacks relentlessly

Maxima is a first prototype, merged so it can be built on and tuned in later PRs. It is selectable in the player menu (rated Hard, listed after Nicowar) and as the AI name `maxima`. It uses the ordinary order interface and **changes no engine code**: existing units, buildings, economy and AIs play exactly as before.

## What it does

A strategy director reads the colony, the terrain and what reconnaissance has seen, then scores competing demands: survival, food, growth, expansion, access, technology, mobility, military and aggression. Policy modules spend those budgets through a private runtime that tracks buildings, gradients, queued orders and lifecycle conditions.

- **Offense.** Every review tick Maxima counts the warriors that are free and trained, reserves one for each open barracks slot, and puts a single war flag directly on the best reachable target with the surplus. A target that stops taking damage is quarantined, and an enemy sealed behind resources gets a corridor dug instead.
- **Food.** Farming protects sustainable food and reserves the routes to it. A food ledger accounts protected farm capacity against what inns and swarms eat, so they are only placed where wheat can back them. Far or starving ones are relocated when the carriers saved repay the rebuild.
- **Staffing.** Each inn or swarm adds or returns one carrier at a time to hold its wheat inside a band.

## Configuration

Every parameter lives in `data/maxima/base.strategy`, with one overlay per match format. Unknown keys, duplicates and inconsistent bounds are rejected with a source location. Tuning inputs are read by Maxima itself from `GLOB2_MAXIMA_BASE`, `GLOB2_MAXIMA_LAYERS`, `GLOB2_MAXIMA_FORMAT`, `GLOB2_MAXIMA_OVERRIDES`, `GLOB2_MAXIMA_TEAM_OVERRIDES`, `GLOB2_MAXIMA_PLAYER_OVERRIDES` and `GLOB2_MAXIMA_TELEMETRY` (see `doc/MaximaStrategyConfiguration.md`), so the game binary has no Maxima options. Saved games record the resolved strategy, so changing a local file cannot change a resumed game.

## Wiring and compatibility

Outside Maxima's own sources, docs, data and tests, the PR only registers the AI: `AI::MAXIMA = 7`, the `AI.cpp` construct/load cases, the `AINames.cpp` roster row and selection order, the `SConscript` source list, translations, and the custom-game setup harness's expected order.

[...]

PR #270, opened 2026-09-12
Relocate Maxima's far or starving inns and swarms instead of only retiring them

genixpro:
Stacked on #268 (which stacks on #267). Phase 3 of the plan: replace the removed zero-capacity swarm rule with a cost-balanced relocation mechanism.

## Intended result

- **Nomination (executor, `Maxima::update_food_relocation`).** A completed inn or swarm whose ledger quality stays ≥ `food.relocation_min_quality_tiles` for `food.relocation_confirm_ticks` is nominated — never an establishing colony, only if a perfect site could still clear a gain floor, and only when nothing is under attack, nobody is critically hungry and no recovery posture is active. There is deliberately no exemption for the first or only building of a kind: the opening swarm and inn are often the ones worth moving. One live nomination; it is exempt from burden retirement while it waits, and relocation runs before retirement each pass.
- **Siting and pricing (planner).** The nomination becomes a `Relocation` intent generated, scored and capacity-checked like a new food building but against a ledger with the old building removed. `Planner::appraiseRelocation` prices in worker-ticks: distance savings on every unit that keeps flowing (scaled by `food.relocation_distance_realisation_percent`, 155, measured on inn carriers and applied to both kinds) plus coverage savings priced at the ledger's unreachable-penalty trip, against the cost of rebuilding every level the old building has. Viable iff payback ≤ `food.relocation_payback_horizon_ticks` (15000) and a gain floor is cleared. A scan that refuses every site is recorded so the executor stops offering at once.
- **Completion.** When the replacement is `Completed`, the executor destroys the old building. Attacks do not block this (the replacement stands); hunger and the coverage-weighted inn-seat rule do; a destroy deferred for a full cooldown is abandoned keeping both.

Full write-up with formulas and the parameter table: `doc/MaximaFoodLedger.md` § Relocation.

## Compatibility

- **Saves:** version 99 → 100 for the nomination state, the actions' `replacesBuildingId` and the planner's refusals; older formats keep defaults (`MINIMUM_VERSION_MINOR` unchanged).
- **Replays / network:** refused against older builds.
- **Feel:** this changes how Maxima manages its economy — inns and swarms now move. On FourSquares every team rebuilds its badly placed starting swarm beside its first farm within the first ~15 000 ticks and relocates inns later; nothing moves without a payback within the horizon. A human play-through is part of review.

## Verification

- `scons` build; all 15 native Maxima regression programs and 76 Python tests pass. New coverage: `Ledger::residualQuality`; the appraisal (stranded inn viable, swarm priced like an inn, realisation knob, off-switch refuses); the ledger exclusion that makes a covered inn's replacement siteable at all; the refusal record; the executor state machine (nominate → offer → refused/lapsed/failed/completed → destroy once → done; deferral bound; off-switch); and the 98/99/100 save-format gates.
- Headless four-Maxima FourSquares, two games: 4/4 teams relocated the starting swarm (nominated at ticks 3 000–5 000 with 0–6% coverage; replacement built and old swarm destroyed by 7 000–15 000), then inns later. Headless Garden 3: 6 nominations → 3 completed, 3 offers lapsed for lack of a construction slot.
- 24 short two-team games with relocation on one side and off the other showed no measurable effect (16–8 wins, +37 population, sd 104): those games end before relocation matters, so this is not evidence either way.

## Limits

- No win-rate evidence yet: 4-player FFA games rarely finish inside the 90k-tick cap and 2-player games end before relocation matters. A longer on-vs-off batch on 3–4 team maps is the right next experiment (`--maxima-player-overrides <player> food.relocation_enabled=false`; player 0 is the passive local slot).
- The calibration's swarm proxy result (quality not predicting swarm trip distance) came from seven games with mostly starving swarms and is recorded as inconclusive, not applied; if a larger sample confirms it, lowering the realisation for swarms is a one-parameter change.
- The first payback formula scaled by 65536 instead of the ledger's `RateScale` (1e6); fixed before any result reported here.

https://claude.ai/code/session_01Xfax2ecD9Y3vnLhocZFYxp


genixpro:
Folded into #264 as commits on codex/ai-maxima-feature (and codex/ai-maxima-dev) rather than a side stack.

PR #269, opened 2026-09-12
Rename the CORN resource identifier to WHEAT

genixpro:
## Why

The in-game resource has always been called **"Wheat"** to players — the `[Wheat]` string-table key is defined in every localization file (including English), and every art asset lives under `datasrc/gfx/.../wheat*`. There is no `[Corn]` key and no `corn`-named asset anywhere in the repo.

But the internal C++ identifier has been **`CORN`** since the project's initial 2001 commit (`git log --follow` on `src/Ressource.h` bottoms out at "Initial revision"). That split runs through variable names, function names, tuning constants, and comments across six AI implementations, map generation, and unit/building code.

Recent AI Cortex work made the collision harder to ignore: it named its own files/types after "wheat" (`CortexWheat.h/.cpp`, `WheatClass`, `scanWheatForbidden`) while still comparing against `CORN` underneath, forcing awkward `wheat (CORN)`-style comments to bridge the two vocabularies inside the same file.

## What this does

- Renames the `CORN` macro (`src/Ressource.h`, value unchanged at `1`) to `WHEAT`.
- Follows through every derived identifier and comment across 80 files: struct fields, tuning constants, AI helper functions (`isCorn`→`isWheat`, `nearestCornDist`→`nearestWheatDist`, `countHarvestableCornWithin`→`countHarvestableWheatWithin`, etc.), and the two design docs (`docs/AI/cortex/DEVLOG.md`, `docs/AI/cortex/upgrade-expand-mechanics.md`) that cite those identifiers.
- Restores column alignment in the handful of struct/table literals where the identifier's one extra character broke it.
- Leaves the player-facing string-table key, translation files, and art asset names untouched — they were already correct.
- Deliberately leaves `docs/gfx/RECOVERED-RUNTIME.md`'s "corn-stalk" mentions alone — those describe an unrelated, unused *art style*, not this resource.

## Compatibility

Pure identifier rename, no logic change. The resource's numeric value (`1`) is unchanged and the symbol name is never serialized, so save/replay/network compatibility is unaffected.

## Verification

- Full client + test-suite rebuild: zero errors.
- `test/TestsRunner` (190 tests) pass.
- All 8 real-engine regression harnesses that reference the resource directly pass: `ResourceFetchTargetHarness`, `FetchApportionmentHarness`, `HiringBucketHarness`, `BuildingExpelHarness`, `InnSwapHarness`, `FailingUnitMarkersHarness`, `TrappedUnitLifecycleTest`, `MapGeneratorStudy`.
[...]

PR #271, opened 2026-09-12
Discussion: cap wheat fetchers at what the map has left, to stop crowds on one tile

genixpro:
**For discussion, not merge.** @Giszmo, this is one possible answer to workers crowding a single tile of wheat. We aren't sure it's a satisfactory one and would like your read before going further.

## The problem

With many inn workers and little ripe wheat, they all head for the same tile. A few get it, it vanishes, and the rest turn around together. The cause is that every fetcher of a team follows the same resource gradient to the nearest wheat, and nothing tells unit #2 that unit #1 already has that grain covered.

## What this PR contains

Three commits, meant to be read separately:

1. **Prototype fetcher cap**, opt-in via `GLOB2_PROTO_SUPPLY_CAP=1`. Without the variable the simulation is unchanged.
   - Each resource gradient rebuild already visits every tile; it now also sums what the goal tiles still hold (`Map::resourceSupply`, per team, resource and swim class).
   - Before hiring, `Team::syncStep` counts the units already on their way to fetch each resource.
   - With the cap on, a building doesn't hire another fetcher for a map resource, and a unit that just delivered doesn't go back out for it, while as many units are already fetching it as the map has left. Markets are still offered.
2. **Measurement tooling**, `GLOB2_HARVEST_METRICS=1`. Per team and resource it prints fetches started, harvests, deliveries, steps with no gradient step (and how many of those stood on a tile the field hadn't caught up with yet), abandoned jobs, destination changes, and the share of fetch time spent heading to a tile already outnumbered by the units heading there. It only reads game state. `GLOB2_SAVE_AT_END=<path>` saves a `--nox` game at its step limit so the moment can be opened in the GUI. Documented in `docs/headless-replays.md`.
3. **`WheatCrowdingHarness`**: a checkerboard farm on a peninsula (forbidden seed squares, water on three sides) beside 6 inns × 8 workers, run for real ticks. It's a measurement scenario, not pass/fail. Documented in `test/README.md`.

## Evidence

**Neutrality.** On FourSquares, 4 AIs, 30,000 ticks, the checksum is `f0bf6f73` on master and on this branch with the cap off, metrics off and on.

**Repro harness.** 12,000 ticks, 10 paired seeds (`WHEAT_SEED=1..10`), cap off vs on, 95% intervals:

| farm | wheat deliveries | abandoned jobs | fetches started | max workers on one tile |
|---|---|---|---|---|
| checker, 7-wide seed square: farm stays bare, each grain rushed | **+15.5% ± 7.5%** (up in 9/10) | 1652 → 1 | −69% | 48 → 28 |
| sparse seeds | +8.9% ± 12% (not significant) | 1437 → 1 | −72% | 48 → 28 |
| checker, full farm: stays stocked | 0.0% ± 1.9% | 0 → 0 | −1% | 48 → 33 |
| dots, full farm: stays stocked | **−2.2% ± 1.8%** | 0 → 0 | −3% | 48 → 32 |

In the bare-farm cases, deliveries rise mainly because grains survive long enough to grow and spread, instead of being stripped the moment they appear by a crowd that also stands on the tiles wheat would grow into.

Reproduce:

```sh
scons -j8 release=1 server=0 wheat-crowding-harness
GLOB2_HARVEST_METRICS=1 WHEAT_SEED=1 ./build/src/WheatCrowdingHarness 12000 checker 6 8 1 7
GLOB2_HARVEST_METRICS=1 WHEAT_SEED=1 GLOB2_PROTO_SUPPLY_CAP=1 ./build/src/WheatCrowdingHarness 12000 checker 6 8 1 7
```

**Real games. This is where we're unsure.** On FourSquares with 4× Nicowar and 4× Econo (seeds 1–4, `GLOB2_TEST_SEED` + `--save-game-as`, master behaviour, tick 30k → end) wheat crowding is severe:

- 16–62 workers heading for one wheat tile, with 65–79% of wheat fetch time aimed at an outnumbered tile (stone: 0–2%).
- It isn't the harness regime: abandoned jobs are rare (64–310 across 4 games against ~28k–43k harvests).
- Most of the waste is workers with no gradient step, 2.2–3.6 per harvest. Only 12–16% of those are on a tile the field hadn't caught up with; the rest are boxed in by other workers and take a random step.

From a Nicowar save at tick 40,000, run 10,000 more ticks: each team has ~1,200 wheat on the map against 12–52 fetchers, so **the cap is essentially never binding in a real game**. The ~1% of ticks where it appears to bind are all right after loading; see limitations. A multi-seed FourSquares A/B with the cap on hasn't been run yet.

## Known limitations

- **Load continuity.** `resourceSupply` isn't saved. After a load it reads 0 until each field rebuilds, and the cap blocks fetching for those ticks, so a loaded game diverges from a continued one when the cap is on. A real version would need to save the count or derive it safely.
- **It limits how many workers go, not where they go.** The fetchers that do go still converge on the nearest tile.
- **The team-wide supply ignores distance.** Wheat on a far farm lets fetchers go to a near farm that can't serve them all.
- **Swim classes.** Supply is per swim class, so a class whose field hasn't been built yet counts as empty.
- **Conflict.** #269 renames `CORN` to `WHEAT`; whichever lands second needs a trivial rebase.

## Feel

In the bare-farm harness the counters say far fewer workers set out for each grain and almost none abandon (not yet looked at by eye). In a FourSquares Nicowar game at ~27 min with the cap on, workers around the farms still looked like they were milling around chaotically rather than going for the wheat that's there. That fits the numbers: the cap barely binds there, and most lost steps are workers boxed in by other workers. In that save the cap-on run also averaged more idle workers (team 0: 7.7 → 12.3), but one pair of diverging runs can't tell whether that's the cap. Not yet played by a human beyond that look.

## Questions for you

1. Is capping hiring at the supply the right layer, or should this be per-tile claims, where each fetcher heads to a distinct ripe tile?
2. Is the bare-farm rush worth solving on its own, given real games mostly suffer from congestion (boxed-in workers stepping at random) rather than over-hiring?
3. If you'd keep any of it: the metrics and harness are independent of the cap except the `GLOB2_IDLE` line, which reads `Map::getResourceSupply`.

https://claude.ai/code/session_01A5GEG5t9ithd47SFm6PTDZ

### DIFF
--- a/docs/headless-replays.md
+++ b/docs/headless-replays.md
@@ -171,6 +171,41 @@ GLOB2_GAME_END ticks=2483 winner_team=1 seed=1777219846 map="Playground" orders=
 The format is intended for grep/regex consumption — fields are
 space-separated key=value, with `map` quoted.
 
+## Resource-Fetch Metrics
+
+`GLOB2_HARVEST_METRICS=1` counts how workers fetch map resources and prints,
+next to the summary line, one line per team and resource that saw any activity,
+plus one per team about idle workers and wheat demand. `GLOB2_HARVEST_METRICS_EVERY=N`
+also prints the cumulative lines every `N` ticks, so a game that ends early can
+still be compared over a window. The hooks only read game state
+(`src/HarvestMetrics.h`), so the final checksum is the same with the metrics
+on or off.
+
+```
+GLOB2_HARVEST tick=30000 team=1 res=CORN commits=3175 harvests=2230 deliveries=2129 steps=17924 reversals=776 retargets=... lost=1051 ghosts=... abandoned=7 trips=2230 tripTicks=355134 goingTicks=345784 doomedTicks=155966 maxCrowd=16
+GLOB2_IDLE tick=30000 team=1 ticks=... idleWorkerTicks=... demandTicks=... idleWhenDemandTicks=... boundTicks=... idleWhenBoundTicks=... supplySum=... fetcherSum=... wantedSum=...
+```
+
+- `commits`, `harvests`, `deliveries` — fetches started, resources taken from
+  the map, resources put into a building
+- `steps`, `reversals`, `retargets` — steps walked towards a resource, steps
+  turning more than 90° from the previous one, steps after which the gradient
+  destination moved
+- `lost` — the gradient offered no step and the unit stepped at random;
+  `ghosts` are the ones standing on a tile the gradient still takes for the
+  resource (a field not yet rebuilt). `abandoned` — the unit gave up the job
+- `trips`, `tripTicks` — fetches that reached a harvest, and ticks from start to harvest
+- `goingTicks`, `doomedTicks`, `maxCrowd` — unit-ticks heading to a resource;
+  of those, the ones heading to a tile already outnumbered by the units heading
+  there; the most units ever heading to one tile
+- `GLOB2_IDLE` — per tick, summed: free workers without a job, whether a
+  building wanted wheat, and (for `GLOB2_PROTO_SUPPLY_CAP`) whether as many
+  wheat fetchers were out as the swim class 0 field counted wheat
+
+`GLOB2_SAVE_AT_END=<path>` writes the game reached at the `--nox` step limit
+with `GameGUI::save()`, so a moment of a headless game can be opened later from
+the GUI's Load Game menu (put the file in the profile's `games/` directory).
+
 For cross-codebase testing, the canonical baselines live in `glob2/tests/baselines/`. See [`docs/replay-verification.md`](../../docs/replay-verification.md) at the workspace root for the full verification workflows and the regeneration procedure.
 
 The `ReplayWriter` records live during gameplay:
--- a/src/EngineRun.cpp
+++ b/src/EngineRun.cpp
@@ -5,6 +5,7 @@
 
 #include "AINames.h"
 #include "ChecksumSidecar.h"
+#include "HarvestMetrics.h"
 #include "DatasetWriter.h"
 #include "Engine.h"
 #include "EngineTiming.h"
@@ -272,6 +273,7 @@ void Engine::printAutomaticEndingSummary()
 	int seconds = (time / GAME_TICKS_PER_SECOND) % 60;
 	int minutes = (time / GAME_TICKS_PER_SECOND) / 60;
 	std::cout << "automaticEndingGame ended: " << time << " ticks, " << minutes << " minutes, " << seconds << " seconds" << std::endl;
+	HarvestMetrics::report(gui.game);
 
 	// Machine-parseable summary line for the AI-trainer pipeline (and any
 	// external driver scraping headless output). One line, key=value pairs,
@@ -537,6 +539,9 @@ void Engine::runOneGameSession(bool& doRunOnceAgain)
 				gui.isRunning = false;
 				automaticGameEndTick = SDL_GetTicks64();
 				printf("nox::gui.game.checkSum() = %08x\n", gui.game.checkSum());
+				// PROTOTYPE: keep the state reached, to open it later in the GUI.
+				if (const char* savePath = getenv("GLOB2_SAVE_AT_END"))
+					saveInitialGameStateOrExit(savePath, "GLOB2_SAVE_AT_END", gui.game.mapHeader.getMapName());
 			}
 		}
 
--- a/src/Game_sync.cpp
+++ b/src/Game_sync.cpp
@@ -3,6 +3,7 @@
 
 
 #include "AICastor.h"
+#include "HarvestMetrics.h"
 
 #include <assert.h>
 #include <string.h>
@@ -180,6 +181,7 @@ void Game::syncStep(Sint32 localTeam)
 
 		Uint64 endTick=SDL_GetTicks64();
 		ticksGameSum[stepCounter&(TICK_PROFILE_BUF_LEN-1)]+=static_cast<Sint64>(endTick) - static_cast<Sint64>(startTick);
+		HarvestMetrics::sample(*this);
 		stepCounter++;
 	}
 }
--- a/src/HarvestMetrics.cpp
+++ b/src/HarvestMetrics.cpp
@@ -1,0 +1,282 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "HarvestMetrics.h"
+
+#include "Game.h"
+#include "GlobalContainer.h"
+#include "Ressource.h"
+#include "RessourceType.h"
+#include "Team.h"
+#include "Unit.h"
+#include "Building.h"
+#include "BuildingType.h"
+
+#include <cstdio>
+#include <cstdlib>
+#include <unordered_map>
+
+namespace
+{
+	const char *const kResourceNames[MAX_RESOURCES] = {
+		"WOOD", "CORN", "PAPYRUS", "STONE", "ALGA", "CHERRY", "ORANGE", "PRUNE"
+	};
+
+	struct Counters
+	{
+		Uint64 commits = 0;     // fetches started
+		Uint64 harvests = 0;    // resources taken from the map
+		Uint64 deliveries = 0;  // resources put into a building
+		Uint64 steps = 0;       // steps walked towards a resource
+		Uint64 reversals = 0;   // steps turning more than 90 degrees from the previous one
+		Uint64 retargets = 0;   // steps after which the gradient destination moved
+		Uint64 lost = 0;        // no gradient step, the job kept
+		Uint64 ghosts = 0;      // of lost: standing where the gradient still sees the resource
+		Uint64 abandoned = 0;   // no gradient step, the job given up
+		Uint64 trips = 0;       // commits that reached a harvest
+		Uint64 tripTicks = 0;   // ticks from commit to harvest, summed over trips
+		Uint64 goingTicks = 0;  // unit-ticks spent heading to a resource
+		Uint64 doomedTicks = 0; // unit-ticks heading to a tile already outnumbered
+		Uint32 maxCrowd = 0;    // most units ever heading to one tile at once
+	};
+
+	struct UnitTrack
+	{
+		Uint32 commitTick = 0;
+		Sint8 lastDx = 0;
+		Sint8 lastDy = 0;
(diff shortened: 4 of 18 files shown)


genixpro:
@Giszmo, a thought on how this could combine with your earlier work.

On FourSquares the cap barely binds. Most of the waste there is congestion: fetchers that find no gradient step because other workers box them in, and step at random instead. That's the same thing your traffic-lanes PR #213 measured as "blocked" calls to `directionByGradient`. In the `lost` counter here, it's everything except `ghosts`, about 85% of lost steps in the 4-AI games. So something along the lines of #213's ant-like directional lanes, which cut blocked ticks by a quarter to a third in your fixtures, might clear up the crowding that this cap can't touch. With the congestion handled, the cap could then be more effective in practice on FourSquares, where it currently makes little difference.

Two caveats, so we don't over-read it:
- #213 is closed, and as written it only fires in passages at most two tiles wide; open ground is unchanged. Farm crowding happens on open ground around the ripe tiles, so it would need the lane idea extended beyond walled passages, e.g. around resource clusters, if that's even the right shape.
- #213's cost (+11–22% wall time as measured there) was the open problem, and any extension would have to solve that too.

Do you think a lane-style direction cost is a plausible way to handle the congestion part, and would you pair it with a fetcher cap like this or with per-tile claims?

Giszmo:
Thanks for the careful write-up. Before choosing between a cap and per-tile claims I want to go one layer down, because both answers assume the crowd around one tile is waste. In the current engine it isn't, and that is the real problem.

**1. A crowd around one grain is rewarded.** In `Unit::handleDisplacement` the `DIS_HARVESTING` branch sets `carriedResource` and only then calls `decResource`, without re-checking the tile. Every worker whose 32-tick harvest completes gets a grain, even if the tile emptied under it. Wheat is granular, 5 grains per full tile, so eight workers around a 1-grain seedling take home eight grains. I added a `phantoms` counter to your metrics (grain granted while the tile was already empty) and ran your harness, cap off, 12,000 ticks, seeds 1–3, mean:

| farm | harvests | of which phantom |
|---|---|---|
| checker, full farm (stocked) | 525 | 109 (21%) |
| checker, 7-wide seed square (bare) | 328 | 159 (48%) |
| sparse | 281 | 151 (54%) |

In exactly the bare-farm regime your cap improves, half of the delivered wheat is created from nothing by the rush. That also explains why the cap helps there: fewer workers per grain means fewer phantom grains, but more grains survive to grow, and the second effect wins. It also means per-tile claims, or anything else that stops workers converging, silently cuts food by 20–50% unless growth is fixed at the same time (see the NO_PHANTOM column below: refusing phantom grains alone costs 10% on the stocked farm and 53% on the bare one).

**2. Why farms vanish without forbidden areas.** Every wheat tile is an equal goal in `updateResourcesGradient`, so the nearest tile, usually a fresh 1-grain seedling at the near edge, is everyone's target. Growth (`Map::growResources`) samples about 1/64 of the tiles per tick, needs a random water probe within ±15 tiles, then a 1-in-3 roll for wheat, and a unit standing on the target tile blocks it. Growing in place (amount+1) is several times more likely than seeding a given neighbour, so a seedling stripped at amount 1 is worth one grain, while the same tile left alone ripens to five and keeps seeding its neighbours. The forbidden checkerboard is players implementing "leave the seeds" by hand. (Fertility is display-only; growth never reads it.)

**3. A ripeness floor in the engine replaces that micromanagement.** Two env-gated prototypes, three small commits on top of your branch in [`proto/wheat-farming-rule`](https://github.com/Globulation2/glob2/tree/proto/wheat-farming-rule) (neutral when unset, same checksums as your head): `GLOB2_PROTO_RIPE_MIN=N` makes a wheat tile a fetch target (gradient goal and touch check) only with at least N grains; `GLOB2_PROTO_NO_PHANTOM=1` refuses the grain when the tile is empty at harvest end and re-targets. Plus a harness pattern `open`: the same checker seeds, nothing forbidden. Deliveries over 30,000 ticks (20 min), 6 inns × 8 workers, sink on, seeds 1–3, mean:

| farm | master | NO_PHANTOM | RIPE_MIN=3 | NO_PHANTOM + RIPE_MIN=3 | wheat left on the farm, master → RIPE_MIN=3 |
|---|---|---|---|---|---|
| checker, stocked | 1340 | 1201 | 1167 | 1154 | 180 → 391 |
| checker, bare | 759 | 350 | 1247 | 1236 | 1 → 457 |
| open, stocked | 1120 | 1064 | 1514 | 1497 | 0 (farm dead) → 338 |
| open, bare | 321 | 261 | 1415 | 1388 | 0 (dead) → 447 |

Abandoned jobs on checker-bare: 4486 → 162. The 12,000-tick runs say the same, with a smaller margin (open-stocked 630 → 593 at 8 min, 1120 → 1514 at 20 min: the floor pays off as soon as the unprotected farm would have been eaten).

What I read from this: (a) an open farm with a ripeness floor out-delivers every forbidden pattern, including the stocked checkerboard on master (1514 vs 1340), and the farm never dies, while on master even a fully stocked open farm is eaten to nothing within 20 minutes; (b) once the floor is in, removing phantom grains costs about 2%, so the two changes belong together; (c) the floor costs 13% on a checkerboard that is already protected by hand, because workers wait for ripeness instead of taking the seedlings the seeds keep producing. A "ripest reachable tile first" preference instead of a hard floor should recover that; not measured yet.


**3b. Letting blocks be eaten through.** A hard floor never empties a tile, so a solid block only yields from its rim and the interior saturates (about 4 grains per tile at the end). Variant `GLOB2_PROTO_SEED_NEIGHBOURS=K`: a tile below the floor is still a target when at least K of its 8 neighbours hold 2+ grains, i.e. its seed is redundant. Same 30,000-tick runs:

| farm | master | hard floor 3 | floor 3, K=3 | floor 3, K=2 |
|---|---|---|---|---|
| checker, stocked | 1340 | 1167 | 1392 | 1349 |
| checker, bare | 759 | 1247 | 1287 | 1158 |
| open, stocked | 1117 (dead) | 1514 | 1566 | 1465 |
| open, bare | 321 (dead) | 1415 | 1400 | 957 (dead) |

K=3 beats master everywhere, keeps the field alive and compact (about 200 grains on 50 tiles at the end instead of 340–450 on 84–113), and blocks get eaten through from the rim. K=2 collapses the bare farms, so the knob is sensitive and needs a real-map check.

**Proposal: a farming area.** The floor should not apply map-wide (outside a farm, stripping a patch bare is what you want). Add a fourth area type next to clearing, forbidden and guard areas, painted with the same brush and orders, one team mask per farmable resource (wheat, wood, alga; `Tile` gets `farmArea[]`, save-format bump like the fertility field). Inside a team's wheat farm:

- wheat is a fetch target only with 3+ grains, or when at least 3 neighbours hold 2+ grains (the numbers above);
- other clearable resources next to wheat count as clearing-area targets for that team, so the farm eats the forest as it grows (one extra predicate at the clearing-area gradient goal and in `tryClaimClearingAreaForHarvesting`, no flag needed);
- growth is untouched; whatever spreads inside the area is protected by the rule, whatever spreads outside is fair game;
- wood and alga farms work the same way (wood is farmable, not just the thing cleared); the brush draws a stylized ear, tree or alga, and refuses to paint where the resource can never grow: wrong terrain (grass for wheat and wood, water for alga), a no-grow tile, or zero fertility, which is exactly the water probe `growResources` needs, so the existing fertility overlay is the painting guide.

Nicowar already paints forbidden checkerboards near water in `update_farming`; it would paint farm areas instead. The phantom grain is a separate bug fix (a harvest completing on an empty tile still hands out a grain) and should land on its own, with the note that 20–50% of the wheat in the rush regime comes from it today, so AI balance will feel it.

**Your questions.**
1. Neither layer. A farming area with the ripeness rule, plus the phantom-grain bug fix. The cap only helps the bare-farm case and doesn't bind in real games, as you found; per-tile claims would fight the growth model.
2. Yes, but as a growth problem, not a hiring problem: see the open-bare row.
3. Keep the metrics and the harness. Add the phantom counter and the `open` pattern, drop the cap commit.

Caveats: one map, 8- and 20-minute runs, unlimited demand (sink), level-0 workers, no human play yet. One thing to know when reproducing your own numbers: `GLOB2_PROTO_SUPPLY_CAP` is presence-checked, so `=0` still turns the cap on; unset it for the baseline.

Reproduce the K=3 row:

```sh
git fetch origin proto/wheat-farming-rule && git checkout proto/wheat-farming-rule
scons -j8 release=1 server=0 wheat-crowding-harness
GLOB2_HARVEST_METRICS=1 WHEAT_SEED=1 ./build/src/WheatCrowdingHarness 30000 open 6 8 1 13
GLOB2_HARVEST_METRICS=1 WHEAT_SEED=1 GLOB2_PROTO_RIPE_MIN=3 GLOB2_PROTO_SEED_NEIGHBOURS=3 ./build/src/WheatCrowdingHarness 30000 open 6 8 1 13
GLOB2_HARVEST_METRICS=1 WHEAT_SEED=1 GLOB2_PROTO_NO_PHANTOM=1 GLOB2_PROTO_RIPE_MIN=3 GLOB2_PROTO_SEED_NEIGHBOURS=3 ./build/src/WheatCrowdingHarness 30000 open 6 8 1 13
```

genixpro:
@Giszmo, agreed on all of it, and thanks for going a layer down. I wasn't attached to the cap. I opened this mainly to get the problem and some numbers in front of you, and the phantom grain is exactly the kind of thing that was missing. If the rush is what's producing half the wheat in the bare-farm case, then any "fetch smarter" fix, whether a cap or per-tile claims, was going to fight a hidden food source. Fixing growth first makes sense.

So, to your three answers:

1. **Neither layer.** Closing this in favour of the farm area, which I see you've already started in #277. I'll review there.
2. **The phantom grain should land as its own bug fix**, called out as a balance change since AI food will drop noticeably in the rush regime. Happy for either of us to take it. Tell me if you'd rather I do it.
3. **Keep the metrics and the harness, drop the cap.** They currently only exist on top of this branch (and your `proto/wheat-farming-rule`). I can split them into a small tooling-only PR with your `phantoms` counter and `open` pattern, minus the cap and the `GLOB2_IDLE` supply read, or you can fold them into #277 if that suits your measurements better. Your call.

Closing this one now; the branch stays up for reference.

PR #272, opened 2026-09-12
Custom-game rules: Economy & resources

genixpro:
## What

Five opt-in custom-game rules in a new "Economy" category of the lobby's Game Rules tab, stored as `GameHeader` fields:

- **No resource growth**: `Map::growResources()` returns early.
- **Scarce resources**: an extra 2x/4x/8x divisor on the grow/extend roll for every resource type.
- **Instant construction**: building sites complete without waiting for deliveries. The rule, not real deliveries, completes the site, so the resource debit is skipped (otherwise the next level would owe a negative balance). The finished building gets its full `hpInit`.
- **Stockpile start**: seeds each team's shared market/exchange pool (+50/+150/+300 of each resource) in `Game::setGameHeader()`. It only re-seeds when the level changes, so loading a save keeps the real accumulated pool.
- **No hunger**: skips the hunger/starvation decrement in `Unit::handleMedical()`.

Every field defaults to today's behavior. The "Standard" ruleset and existing games are unchanged.

### Also in this PR
- `GLOB2_TEST_RULES` (documented in `docs/headless-replays.md`) turns rules on for `-test-games(-nox)` matches as `name=value` pairs, so AI matches and reviewers can exercise the rules headlessly.
- The Rules tab now lists rules grouped by category. Later rules can belong to earlier categories, and a heading no longer repeats.
- `GameHeaderTextSaveLoadTest`'s version-100 boundary check assumed the AI-config block ends the stream; it now accounts for the rule bytes after it.
- `MaximaReconPolicyTest.py`/`MaximaStrategyPolicyTest.py` asserted `VERSION_MINOR == 100` literally. They now assert `>= 100`, which is what Maxima's save gates need.

## Review fixes (rebased onto current master)
- **Version collision:** the branch claimed `VERSION_MINOR`/`FILE_FORMAT_VERSION_ECONOMY_RULES` 98, which master has since used (Cabino at 98; per-player AI configuration at 101). Loading any of those saves would have misread these fields. The rules are now version **102**.
- **Network protocol:** `GameHeader` crosses the network via `loadWithoutPlayerInfo(stream, VERSION_MINOR)`, so older clients would misparse it. Bumped `NET_PROTOCOL_VERSION` and `YOG_MIN_CLIENT_NET_PROTOCOL_VERSION` to 31.
- **Instant construction HP:** a new building finished at the site's 1 HP, and an instant "repair" completed without restoring any HP. AIs such as Castor would then re-order repairs forever. Completion now grants the finished level's `hpInit`.

## Compatibility
- **Saves:** fields are read only when `versionMinor >= 102`, so saves from 58–101 load with every rule off (`GameHeaderTextSaveLoadTest` now checks that a version-101 header loads byte-exactly with the rules off). New saves need a 102+ client.
- **Replays and multiplayer:** version 102 and net protocol 31 refuse older replays and clients, as with previous format bumps.

## Verification
- Client and `test/` build clean. `TestsRunner` passes (203 tests), and `GameHeaderTextSaveLoadTest` round-trips all five fields. `test_translations.py` and the Maxima policy tests pass.
- **Default rules are unchanged:** on the full three-PR stack, 4 matches (`GLOB2_TEST_SEED=101`, 30,000 ticks; FourSquares1, balanced, SmallForTwo, A_big_pond; covering Castor, Warrush, Nicowar, Cortex, Maxima, Cabino, Econo and Numbi) have per-tick checksum sidecars byte-identical to current master.
- **AI matrix:** 84 headless matches on the full stack, 21 rule configurations (each rule alone plus stacked combinations, including every rule at once) × 4 matchups, 30,000 ticks each. All ran to completion with no crash, assert or hang.
- **Save/load:** a tick-0 `--save-game-as` save with seven rules stacked (economy and combat), reloaded via `--nox`, reproduces every per-team, unit and building checksum for all 4,514 ticks.

## Limits / notes for the maintainer
- **Stockpile start** only feeds markets and exchanges. The AIs rarely build those early, so the rule had no measurable effect in AI matches.
- **Lobby persistence:** `CustomGamePreferences` does not persist these fields, so a reopened lobby shows them at defaults.
- **Feel:** all five rules change pacing and economy pressure by design, but only when chosen.

Part of the custom-game rules batch: #272 (this), then #273 (combat, stacked on this), then #274 (victory).

https://claude.ai/code/session_016Z2T8ETQJ2q8DQr2aSAm8p


genixpro:
Companion PRs in this batch: #273 (Units & combat), #274 (Victory & pacing).

PR #275, opened 2026-09-13
Draw the torus view's map markers at the atlas's scale, not the window's

Giszmo:
Reported by Leo: in the torus view at a 300% interface scale the hunger, hitpoint,
workers and units-inside markers blow up, the selected building's ring turns into a
thick band, and the white circles on the units working for it disappear.

## Markers blow up

GL rasterises lines at a width in the pixels of whatever is being drawn into, so
`setScaledLineWidth()` multiplies by `drawableScale()` — the factor the window
stretches the logical surface by. That is right for the window and wrong for the
torus, which renders the whole map into an offscreen atlas at one texel per world
pixel, so every outline lands there three times too thick.

The status bars make it obvious: `drawPointBar()` paints its *filled* segments as
quads, which stay correct, and its *empty* ones as `drawRect()` outlines, which at a
3 px stroke fill the 4x4 cell solid. That is the grey smear in the report. The
selection circle asks for 2 px and gets 6. `drawLine()`, which builds a thick line
from several 1 px passes, has the mirror-image problem: in the atlas its passes
collapse onto one texel, so the area outlines came out too thin.

`rasterScale()` now reports the offscreen target's pixels per logical pixel while one
is set and the window's otherwise, and the atlas pass sets it to the atlas's texels
per world pixel.

selection ring and status bars

## Missing selection markers

`drawOverlayInfos()` skips its whole selection block while the torus is active, and
`drawTorusMap()` mirrored only the selected building's own circle. The white circles
marking the units working for it, and the circle on a selected resource, were never
drawn at all; the building circle was also always blue instead of blue, orange or red
by ownership. They are now drawn onto the atlas, anchored to the map surface, the way
the flat view draws them over it.

worker markers

## Validation

Both screenshots are the same seeded `-test-games` run (`GLOB2_TEST_SEED=4242 --map
balanced`) at 2560x1440 with `GLOB2_UI_SCALE=3`, master on the left and this branch
on the right, same scripted click and `g` toggle. They live in commit e08cb245e, which
is removed again at the branch tip.

- `test/TestsRunner`: OK (190 tests)
- `torus-render-test -g` at 1120x720, and at 1920x1440 with `GLOB2_UI_SCALE=3`
- `torus-render-test -G` (software)
- `point-bar-render-test` gl and software
- `FullscreenAspectHarness` and `WindowResizeHarness`, gl and software

`TorusRenderIntegrationTest` gains two assertions read back from the atlas texture:
the selection circle is at most 3 texels wide, and a building's worker circles add
bright pixels over the worker's tile. Both fail on the unfixed code — the first only
once the interface is scaled, which is why CI gains a `torus-render-test` run at
`GLOB2_UI_SCALE=3`.

## Limits

No simulation, save or replay state is touched; this is presentation only. The torus
star field still sizes its points in window pixels, so the sky is slightly finer on a
scaled desktop than on an unscaled one — not part of this change.


genixpro (approved):
Approving. Line width should follow the target being rasterised into, not the window, and `rasterScale()` puts that in one place. I checked the new torus selection block against the flat view's `drawOverlayInfos()`: the ownership colours, the worker-position interpolation and the `showUnitWorkingToBuilding` / allies gate all match.

Verified on macOS arm64 with a real GPU (Apple M3), not just CI's software GL: `torus-render-test -g -F -m -s 1920x1440` passes at `GLOB2_UI_SCALE=1` and `3`. The selection circle is 2 atlas texels wide at both, and the worker marker goes from 491 to 626 bright pixels at 1 and from 503 to 639 at 3.

Non-blocking notes for a follow-up if you want:
- The flat view draws each marker once per visible map copy (`forEachMapCopy`); the atlas holds one copy of the world, so a ring or worker circle straddling the map seam is cut at the seam on the torus. The old blue ring already had this.
- The ownership-colour and worker-circle logic now lives in both `GameGUIDraw.cpp` and `GameGUITorus.cpp`; a shared helper taking the origin would keep the two views from drifting.
- The screenshots in the description point at `e08cb245e`, which the branch tip removes. I'm squash-merging without deleting the branch so those links keep working.

PR #277, opened 2026-09-14
Farm areas: pool a harvest over the connected field

Giszmo:
**Draft for discussion.** A fourth painted area that changes how a farmed field is harvested, following the analysis in #271.

## The problem

Wheat is granular: a tile holds up to five grains and regrows. Every wheat tile is an equal goal of the resource gradient, so a crowd of fetchers converges on the nearest one — usually a fresh one-grain seedling at the near edge of the field — and strips it. The field is eaten from its rim inwards and then cannot recover, because nothing in the engine seeds an empty region: growth only ever runs on a tile that already holds the resource (`Map::growResources`). Players work around this by hand, painting forbidden checkerboards so workers cannot reach the seed tiles.

## The proposal

A farm area, painted with the same brush and the same add/remove modes as forbidden, guard and clearing areas, and saved per tile. It changes exactly one thing: **where a harvest takes its resource from.**

When a worker's harvest animation completes on a tile inside its own team's farm area:

1. **The field** is every tile holding that resource reachable from a tile of the worker's own 3x3, through 8-connected tiles that also hold it, without leaving the painted area.
2. One unit comes off the tile of that field holding the **most**. Ties break by distance to the worker, then by tile index, so the choice is identical on every client and in every replay.
3. A tile at one grain is that field's **seed** and is never the source. `decResource` clears a granular tile at its last grain, so without this a farm could be worked down to bare ground and would be protecting nothing. A field worked past its surplus stalls at one grain a tile and regrows; the workers who arrive meanwhile go home empty. Seed tiles still carry the flood, so a field does not fragment as it is worked down.
4. If nothing in the field is above its seed, the worker gets **nothing** and goes back to looking instead of walking home empty-handed.

Seeding the flood from the worker's 3x3 rather than from its target tile keeps a harvest working when the target empties during the 32-tick animation, and it is also what refuses to teleport: connectivity runs through the resource, so an empty gap ends the field. A farm painted over two disconnected patches of wheat serves only the patch the worker can actually reach — a worker still has to walk into the area to get to the other one, and painting an area over bare ground creates nothing.

8-connectivity means a checkerboard of wheat is a single farm, which is what the diagonal reach of the existing touch test already implies.

**The brush refuses ground nothing can grow on.** A farm on grass is a wheat farm and a farm on water is an alga farm — the terrain already says which, so there is no extra UI. `Map::canPaintFarmArea` rejects sand, stone, the fruits, papyrus, tiles the map forbids growth on, and tiles too far from water for the growth probe to succeed. That last test is read off `Map::growResources` rather than invented: it draws offsets in [-15,15] and needs `isWater` there, with algae additionally needing `isSand` at twice the offsets, so a tile no draw can satisfy never grows. The refusal lives in `Game::executeAlterFarmArea`, the path a replay and every remote client take; the brush preview and the map editor apply the same predicate so the overlay cannot disagree with what lands. A refused tile stays out of every field, so wheat on it is harvested off its own tile and cannot connect two patches or act as a source that never refills.

## What deliberately does not change

- **The resource gradient.** Workers walk to the nearest tile holding the resource, by the same field, with the same goals. Only the moment the harvest completes is different.
- **Growth.** A plant has no opinion about the farmer's intentions. `Map::growResources` does not read the farm mask, so a field grows and expands by identical rules with or without an area over it. `growthIgnoresTheFarmMask` in the harness runs 20,000 ticks on two copies of the same map from the same seed — a 5x5 patch grows to 124 tiles and 532 grains, so the comparison is not vacuous — and asserts every tile equal.
- **Clearing.** Clearing is not harvesting. `handleMovementClearingResources` and `tryClaimClearingAreaForHarvesting` call `Map::decResource` on the tile they are touching and never go through the new path, so a clearing worker inside a farm empties the tile it is aimed at and cannot reach into the field. Covered by its own case.
- **Everything outside a farm area**, phantom grain included. The global phantom-grain fix stays the separate change #271 concluded it should be.

## Keeping the farm clear

A farm is kept clear of what it does not grow. `Map::isClearingTarget` makes any clearable resource inside a farm area a clearing goal unless it is the crop that tile's terrain grows, so a wheat farm cuts down the wood creeping into it and expands into the forest as the trees come down. It is the predicate the clearing-area gradient and `Unit::tryClaimClearingAreaForHarvesting` already used, widened by one case, so a farm needs no clearing flag of its own and the existing claim bookkeeping still stops two workers taking the same tile.

## Which resources

The rule applies where a resource is granular, shrinkable and expendable: **wheat and algae**.

Algae are not the exception they look like. `ALGA` is `granular=1, expendable=1, shrinkable=1` and `growResources` has an explicit `ALGA` branch, so algae grow in place and seed neighbours exactly as wheat does — they need water within ±15 and sand within ±30, so they only live on coasts. A coastal strip never comes back after being harvested out not because algae cannot regrow, but because the strip is one tile deep and nothing reseeds an empty region. A wheat field survives the same treatment only because its interior outlives its rim.

**Wood** is `granular=0`: one harvest clears the whole tile, so a wood tile is never part of a shared stock and a farm painted over a forest changes nothing about which tree falls. Stone and the fruits are eternal.

## Effect on play

- The rim tile a worker stands next to is no longer the tile that loses a grain, so edge seedlings survive and ripen.
- Workers therefore **stop at the outer rim and stop walking into the field.** A ground unit cannot stand on a resource tile at all (`isFreeForGroundUnit` rejects any resource), and it does block growth into the empty tile it is standing on (`incResource`), so the field stops being trampled.
- A harvest that finds nothing hands out nothing, inside a farm.

The cost of harvesting hard is paid in the field's **footprint**, not its stock. A sampled source adds one grain either in place or to a neighbour, with spread probability `amount/8`, so a field flattened by heavy harvesting expands at a fifth of the rate of a ripe one — and the ring of workers standing around it occupies the very tiles it would expand into. More workers, less expansion. That trade only exists if the painted area is larger than the current patch; painted tight around a saturated field it is free yield. **This is read off the growth code and is not measured** — a yield-versus-worker-count curve is the obvious next step and would say whether the knee is anywhere interesting.

## Compatibility

```
gd-large-4ai.game    20,000 ticks   master d4a46489   branch d4a46489   sidecars byte-identical
gd-archipelago.game  20,000 ticks   master b7e80200   branch b7e80200   sidecars byte-identical
```

`GLOB2_CHECKSUM_SIDECAR=1` per-tick, per-unit and per-building streams, `cmp`-equal between a master build at `de46aa05a` and this branch. With no farm painted, the simulation is master.

Save format goes to **98**, read behind `FILE_FORMAT_VERSION_FARM_AREA`, so every older save still loads with no farm painted (v84 and v88 fixtures verified). `REPLAY_MINIMUM_VERSION_MINOR` deliberately **stays at 97**: no older replay can contain the order that paints a farm, and a v97 map loads with every mask at zero, so those replays still reproduce exactly. `Tile::farmArea` joins `Map::checkSum` — zero everywhere until a farm is painted, so no existing checksum moves, and a desync caused by a farm area becomes catchable.

## UI

The zone strip could not take a fourth button at its old 40px pitch (three from panel-x 8 already end at 136 in a 160px panel). It is now 38 from 6, and both the draw path and the click path take every position from `zoneStripButtonX` / `zoneStripButtonAt`, so the drawn grid and the clickable grid cannot drift. The map editor gets the same layout.

| | |
|---|---|
| zone strip | tooltip |

farm area painted in game

One rendering fix rides along, in its own commit: SDL_ttf defaults to full hinting, which at the 10px panel size merges an `r` into a following `m`. "Farm area" rendered as "Famm area" — and so does every one of the twelve English strings containing "Swarm". `TTF_HINTING_LIGHT` keeps the vertical snapping that makes small text crisp and leaves horizontal metrics alone. Confirmed in the real GUI at native scale, so this was never a display-scaling artifact. It changes how every string rasterises, so it can be taken on its own.

kerning before and after

Painted live under Xvfb with the real GUI (`-test-games --map balanced`, brush driven with xdotool). `[farm area]` is translated in all 32 catalogs; `check_translations.py --strict`, `test_translations.py` and `test_font_coverage.py` pass. Default binding `<a>-<w>`. The screenshots live in commit 149afb342, removed at the branch tip so the working tree stays clean.

## Size of the change

84 files and +895 lines looks large for one rule. It breaks down as:

| lines | files | |
|---:|---:|---|
| 144 | 2 | **the rule itself** — 107 of them non-comment |
| 96 | 5 | refusing to paint ground that cannot grow |
| 178 | 17 | plumbing for a fourth area: tile mask, order, save format |
| 120 | 16 | UI: panel layout, overlay, map editor |
| 272 | 2 | tests |
| 110 | 2 | documentation |
| 67 | 34 | `[farm area]` in 32 catalogs plus the key list |
| 3 | 1 | an unrelated build fix, below |

The mechanic is about a hundred lines. The rest is the cost of adding an area type at all, and would be the same for any fourth area.

## Tests

```sh
scons -j8 release=1 server=0 farm-test && ./build/src/FarmAreaHarvestHarness
```

Sixteen cases: the ripest-tile source, the empty gap that must not be crossed, a worker out of reach of any wheat, the exhausted field, the target emptying mid-animation, wood, algae, another team's area, off-farm behaviour unchanged, a worker harvesting from one spot flattening what it can touch and then having to walk, growth being identical with and without the mask, clearing staying on its own tile, the brush refusing sand, stone, no-grow and waterless ground, a refused tile failing to connect two patches, a farm giving up its surplus while keeping every tile alive where the same unfarmed field loses its rim to bare ground, and wood inside a farm becoming a clearing target while the crop never does.

Also green: `TestsRunner` (190), every `test/` harness, `savegame-safety-test`, `terrain-test`, `global-gradient-test`, `selection-test`, `team-stats-save-test` including the v84/v88 legacy fixtures.

## One unrelated commit

`Follow the CORN to WHEAT rename into tools/MenuColonyHarness.cpp`. #269 renamed the resource identifier but missed `tools/`, so `menu-colony-harness` has not compiled since it merged and both Linux CI jobs are red on master itself at `de46aa05a`. Every CI step after that one is skipped, so no branch could be validated past it. It rides here only so this pull request's CI means something, and belongs on master on its own.

## Open questions

1. **The trade may be too forgiving.** If the measured yield curve is flat, two candidate knobs, in order of preference: pool within a radius of the worker rather than over the whole field, so a large farm cannot be drained from one corner; or scale harvest time with the field's fill ratio, so a stripped field is slow to work.
2. **The AIs are untouched.** Nicowar still paints forbidden checkerboards in its farming phase. Swapping that for a farm area is a balance change that wants its own measurement.
3. **Not played by a human yet.**

`docs/farm-areas.md` carries the full rule and a map of where each piece lives.


genixpro:
@Giszmo, I like this conceptually. It gives players a direct tool for exactly what #271 diagnosed, and it's a nicer answer than hand-painted forbidden checkerboards. For a new mechanic like this the decision comes from playing it, so we're planning to playtest it in practice before going further. The rule itself reads carefully built: a small core, no change unless a farm is painted (the byte-identical sidecars are convincing), and good harness coverage.

What we'll be watching for in play:

- **Exhausted farms.** Seed tiles are still resource-gradient goals, so once a field is worked down to one grain a tile, fetchers may keep walking to it and completing empty harvests while wheat outside the farm goes untouched. Their bodies also stand on the tiles the field would regrow into. Does that show up, and how badly?
- **Readability.** Grain vanishing from the interior while workers stand at the rim is logical, but is it clear to a player? That ties into your open question 1: a large field drained from one corner.
- **Whether farming feels like less babysitting** than checkerboards, and whether the brush's refusals on sand, stone or waterless ground are clear enough.

Alongside the playtest we'd like yield numbers for this rule specifically. The #271 measurements were of the ripeness floor, and pooling isn't measured yet. A paired multi-seed run in the wheat-crowding harness (farm area vs checkerboard vs open field vs ripeness floor) would say whether pooling holds up.

A few mechanical items that don't depend on the playtest, if Bob wants to pick them up in the meantime:

1. **Network gate.** The PR adds `ORDER_ALTER_FARM_AREA` and changes the simulation once a farm is painted, but leaves `NET_PROTOCOL_VERSION` and the YOG client floor alone, so an old client in a mixed game would get an order it doesn't know. Leaving the replay floor at its current value is fine; the protocol gate needs to move.
2. **Version number.** 98 has since gone to Cabino (#248), #256 is taking 99 and #265 is queued for 100, so this wants the next free number after those, and it conflicts with master now.
3. **Split out the hinting commit.** `TTF_HINTING_LIGHT` is a good fix, but #263 has just merged and reopens fonts at the screen's scale (`renderFont`). As written the hinting would only reach the authored-size font. It deserves its own small PR on top of #263 that sets it on both.
4. **Drop the `MenuColonyHarness` CORN commit.** #278 already put that fix on master.

genixpro:
@Giszmo, one more note on #277, about the art. The green is right, but the wavy lines read as water to me. Over a painted wheat field the area looks flooded, and they're much denser than the other three overlays. The forbidden, guard and clearing sprites all draw a few sparse symbols per tile. The farm tile fills itself with lines, about 520 visible pixels against about 180 for clearing's crescents.

We mocked up a few alternatives on real wheat from your in-game screenshot, at native 32px tiles with the engine's per-tile frame offsets and area border. Wheat ears, seedlings and crop rows came first, then lighter versions of the seedlings. **B1 has the balance I like best**: two small seedlings per tile that grow and open across the 8 frames, about 180 visible pixels, in your `(0,200,80)`. The button is one large seedling in the square outline, matching the other zone buttons.

sprout variants over wheat, 2x

B1 is top right. Same four at 1:1: [sprout-variants-100pct.png](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/previews/sprout-variants-100pct.png). First round, with ears and rows: [first-round-candidates-2x.png](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/previews/first-round-candidates-2x.png).

**Raw files**, named as the engine loads them, drop-in replacements for `data/gfx/` on your branch:

- overlay frames: [area-farm0](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm0.png) · [1](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm1.png) · [2](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm2.png) · [3](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm3.png) · [4](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm4.png) · [5](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm5.png) · [6](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm6.png) · [7](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/area-farm7.png)
- zone button: [gamegui58.png](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/data/gfx/gamegui58.png)
- generator: [generate.py](https://raw.githubusercontent.com/Globulation2/glob2/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area/generate.py). It reproduces these files exactly. `SPOTS`, `ALPHA` and `SIZE` at the top are the knobs if they need nudging once it's seen in motion.

All of it is on branch [`proposal/farm-area-sprout-art`](https://github.com/Globulation2/glob2/tree/2f9fb2c6075fcb538d349b8054b0d72cf4949c76/art-proposals/farm-area), outside `data/` so it doesn't touch the game. Copy the nine PNGs over yours. These are mock-ups meant to settle a direction, so feel free to redraw them by hand in the same spirit. The playtest will tell us whether they read at a glance in motion.

genixpro:
To be clear about where we are: none of this is from playing or testing the farm area yet. We haven't run the branch or exercised the mechanic; the notes above are a read of the code and PR, plus feedback on the art style. The playtest is still to come, and the gameplay questions (exhausted farms, readability, yield) stay open until then.

Giszmo:
This change has merits in some aspects but clearly, no AI would use the mechanism and no map would have the right amount of wheat but my main concern is that now the "farm" is either producing too much or too little. In the too little case, it depleats very suddenly, with all the 1-wheat getting harvested in quick succession the moment the colony tips from abundance to shortage against to what I had in mind for the feature but protecting the last corn per tile is probably even worse as then the player could be completely oblivious about farms. This PR is still in Draft mode. @genixpro thank you for your feedback. I agree on the graphics. I'll give it some more rounds of polish but right now I fear it takes away more than only annoying forbidden-area placement.

PR #280, opened 2026-09-14
Let the 2D map zoom out to 25%

genixpro:
The 2D map could only zoom out to 50%, so the widest view showed twice the map it does at 100%. On a large map that isn't enough to take in a front or an island group at once. The floor is now **25%**: a tile is 8 px, and the view holds four times the map it does at 100%. This covers gameplay, replays and the editor, which all share `MapCamera`. The torus view keeps its own range and is unchanged.

## Change

- `MapCamera::setZoom` clamps to `MapCamera::MIN_ZOOM` (0.25) and `MAX_ZOOM` (3.0) instead of literal `.5` and `3.0`. Mouse wheel and the `-` button both reach the new floor through it. The zoom label already prints any percentage.
- Nothing else was sized around 50%. The viewport, map copies, minimap rectangle and picking all derive from `visibleW()`/`visibleH()`.

## Before / after (HD artwork, 1024×768 logical window)

| | 50% (old floor) | 25% (new floor) |
| --- | --- | --- |
| Gameplay, `gd-small-2ai` |  |  |
| Editor, Archipelago |  |  |
| Dense 64×64 test map, 4 teams |  |  |

At 25% a small map repeats across the view, as the torus wraps, and the minimap's view rectangle is wider than the map.

## Verification (macOS arm64, Apple M3)

- `MapCameraTest` covers the new floor: conversions at 25%, wheel clamping to `MIN_ZOOM`, and a 512×512 map at 25% showing 3840×2880 logical pixels with correct tile picking across repeats.
- `HighResolutionIntegrationHarness` (`highres-integration-test`) passes with 25% added to its gameplay, editor and dense-map loops. At 25% it checks:
  - mouse-to-tile conversion and that a building placement lands on that tile,
  - Alt-wheel isolation,
  - editor brush painting,
  - repeated small maps,
  - that drawing leaves the simulation checksum unchanged.

## Cost to review in play

Zooming out draws more tiles, and the 25% frame costs about three times the 50% one. Frame time from the harness, 1024×768 logical window, mean of 10–20 draws:

| Scene | 25% | 50% | 100% |
| --- | --- | --- | --- |
| Gameplay, original art | 19.2 ms, 11,692 calls | 6.0 ms, 3,468 calls | 2.8 ms, 1,111 calls |
| Gameplay, HD art | 20.3 ms, 11,724 calls | 6.8 ms, 3,492 calls | 3.3 ms, 1,135 calls |
| Dense map, HD art | 22.1 ms, 14,009 calls | 6.3 ms, 3,889 calls | — |

Cost grows with the logical window area, so a larger window or a slower GPU will feel this more at 25%. GPU texture memory is unchanged. A maintainer should try the new floor on a large map in a big window and judge whether 25% is the right floor or something like 33% is.

## Limits

- Presentation only: no simulation, save, replay or network change.
- Tested on one macOS machine; not run on Linux/Windows GPUs or the software renderer, which has no zoom at all (`canDrawStretchedSprite`).

https://claude.ai/code/session_01WePqccHKmAjY6cwoqEcvBE


genixpro:
Also verified on Linux (Ubuntu 24.04 x86_64, g++ 13.3, Mesa software GL under `xvfb-run`, `LIBGL_ALWAYS_SOFTWARE=1`), on a fresh clone of `c24b51c0`:

- `MapCameraTest`: PASS
- `HighResolutionIntegrationHarness`: all checks PASS at 25%, including tile picking, placement, editor brushes, repeated small maps and unchanged simulation checksums.

Draw calls match macOS exactly. The software rasteriser is slow in absolute terms, so only the ratio carries over. It's the same ~3× cost at 25% versus 50%:

| Scene | 25% | 50% |
| --- | --- | --- |
| Gameplay, original art | 83.0 ms, 11,692 calls | 26.5 ms, 3,468 calls |
| Gameplay, HD art | 88.3 ms, 11,724 calls | 26.3 ms, 3,492 calls |
| Dense map, HD art | 51.6 ms, 14,009 calls | 15.3 ms, 3,889 calls |

PR #306, opened 2026-09-15
Add the Continents map generator, drawn from real geography

genixpro:
A new landscape, **Continents**: a real continent floating in an ocean that wraps round the torus, chosen at random or by name (North America, South America, Africa, Europe, Asia, Oceania), with the coasts, lakes, great rivers, deserts, mountain ranges and biomes a player knows. Lakes and sea are water, deserts and ice caps sand, ranges stone scree, forests wood cover, plains wheat country; the great rivers are lines of water with fords, and islets fill the open ocean. It is a toy rather than a tournament map: the geography is fixed, so no two colonies get the same ground and fairness is measured after the fact. What it promises is that every colony can start and that the continent looks like itself.

## Screenshots

| North America, 256x256, 4 colonies | Europe, 256x256, 4 colonies |
| --- | --- |
| North America | Europe |

| Africa | South America |
| --- | --- |
| Africa | South America |

| Asia | Oceania |
| --- | --- |
| Asia | Oceania |

Asia at 512x512 with 8 colonies:

Asia 512

South America turned to fit a 512x128 map, and Europe at 128x128:

South America 512x128
Europe 128

The source rasters the generator fits onto the map (from `tools/world_atlas.py --preview`):

| Atlas North America | Atlas Europe |
| --- | --- |

## What is in the change

- **Data.** `tools/world_atlas.py` bakes Natural Earth 1:50m (public domain) and the Köppen-Geiger climate classification (Kottek et al. 2006, credited in `data/authors.txt`) into `src/map/generator/shared/WorldAtlasData.cpp`: six run-length encoded rasters, 512 pixels across each landmass, about 217 KB of generated source. Nothing reads a file at run time. Documented in `docs/map-generators/WORLD_ATLAS.md`.
- **Shared toolkit** (most of the new code): `WorldAtlas`, `Raster` (a picture fitted onto the torus inside a sea margin, turned to fit a rectangle, resampled by majority or by flag), `Landmass` (coasts cleaned of slivers, specks and islets nothing fits on; the mainland); `farthestSites` and `recentreSites` in `Points`; `windowCount` and `bridgeDiagonals` in `Morphology`; `noisyShare` in `LatticeNoise`; `digPond` and `meanFertilityAround` in `Growth`; `clearRoute` and a lane radius plus protected mask for `openColonyRoutes` in `Roads`; biome kits for real kinds of land (`farmland`, `woodland`, `savanna`, `barrens`, `highland`), stone or wood cover, a dry reserve, `scaledBiome` and a no-cover clearing in `Biomes`. Existing callers keep their behaviour (new parameters default to the old one); no other generator's golden rows changed.
- **Generator** `ContinentsGenerator.cpp`, composed from those stages, with every number and fallback explained in its comments. Controls: continent, orientation (turn to fit or upright), mountain ranges, great rivers, islets, oases, and the five resource amounts.
- `tools/new_map_generator.py` now skips the retired legacy ids (it had handed out 33, which the framework doc reserves); Continents is 44 (35 to 43 went to the generators merged before it).
- Translations of the 18 new strings in every table, written by one agent and reviewed by a second; docs and the map-design skill's references updated.

## Verification

- `MapGeneratorDefaultsTest`, `CustomGameSetupHarness` and `MapGeneratorGoldenTest --require-rows` pass; the golden `--sweep` reports 0 failing combinations. Golden rows for `continents` are recorded for **linux-x86_64 and macos-arm64 and are identical** (the whole pipeline is integer arithmetic).
- **Bulk generation**, local, 624 requests: every size and shape from 64 to 512 on each axis, 2 to 12 colonies, every continent, plus 240 random draws over every control. 562 generated; all 62 refusals were requests with a 64-tile side that the continent cannot hold (40 Oceania, 12 Asia), refused up front by `validateRequest` so the lobby never rolls them. Every combination with both sides at 128 or more generated for every continent and colony count. At 256x256 with four colonies, fairness (weakest over strongest) runs 0.78 to 0.97 by continent and seed.
- **AI playtest on the maintainer's hosts** (`tools/tournaments`, Linux bundle): 56 four-colony free-for-all games of Nicowar against Maxima, every continent, two map seeds, all four rotations, 40 000 ticks. No colony of 224 failed to grow past twelve units; both AIs build and fight on every continent. The colonies eliminated first had started with the least building room (25 to 60 sites within their catchment: a site inside a range that scree then covered, or a cramped cape) or the least fertile ground (under about 2000 on the growth scale). That drove the tuning: sites are scored by the fertile grass round them, kept clear of rivers and preferred for reachable catchment; every home keeps an 11x11 clearing free of ambient deposits and a 17x17 free of cover; a dry start is dug up to three ponds until its fertility meets a floor. On the twelve playtest maps the worst colony's building sites went from 25-58 to 53-367 while fertility stayed level. A confirmation run on the tuned build (28 games, every continent, one map seed, all rotations, same AIs and tick cap) again had no colony fail to grow; the worst colony's building sites per map rose from 25 to 58 before tuning to 61 to 310, mean peak units per colony from 89 to 98 and mean final buildings from 9.7 to 10.2. Survival to the tick cap fell from 72% to 65% because more games ended by elimination (7 of 28 against 7 of 56): with better starts the colonies fight sooner, which is the war game working rather than the map failing.

## Limits

- Fairness is best effort, by design: the lobby keeps the best of its rolls and the scorer measures the finished map, but a continent's geography decides who gets a coast and who gets a desert.
- A 64-tile side cannot hold most continents with more than two colonies; the request is refused with a message rather than generating a broken map.
- Rivers are omitted below a tile per 0.3 atlas cells (128 maps), where their three-tile scars would take more land than the lakes.
- Nobody has played it by hand yet; that is the review. The maps this touches are only the new one (purely additive), so per AGENTS.md this is low-risk, but a maintainer playing Africa or Asia is the real test of whether the toy is fun.

## Maintainer feedback

The ring of round islands round every continent was invented, not geography: `raiseIslands` drops them in the open sea as swimming prizes. The maintainer's first look at Europe asked where they came from, so the `islets` toggle is now off by default and the atlas alone draws the coast; anyone who wants the prizes can turn them on. Revision 2, default fingerprints regenerated on both platforms.

Europe, 256x256, 4 colonies, revision 2

## Review changes

- Legacy id **44**; catalog row in the framework doc; registry and toolkit tables merged with the generators stacked before it.

## Rotation tournament

Six 256×256 maps, four colonies, every cyclic team rotation, four Nicowars, 45,000 ticks (revision 1, before the islets went off; the islets are out in the sea and touch no start): the geography decides, as a toy's should. Pooled per-start peaks from 90 to 200 units, a root-mean-square position bias of 38 points, and on four of six maps one start that never passed about 30 units and was eliminated in every rotation by a rival 55 to 120 tiles away, with more combat than starvation deaths. The start metrics (fertility 0.3 to 0.9, 180 to 500 sites) do not flag those starts, so what dooms them is who reaches them first, not what they hold. Recorded in the framework doc; the retained maps are the material for anyone who wants to move the never-grow starts.


genixpro:
Love it. although i don't know where all the random islands that surround the continents are - are these actual islands but which are just way smaller or disproportionate in size within glob? I'm leaning towards just deleting all these outlying islands and focusing in on making that mainland accurate. Except in oceania where the islands are basically the whole deal. We can keep them there. Already pretty good job!

PR #308, opened 2026-09-15
Add Savannah map generator with open plains and watering holes

genixpro:
Savannah adds an open grassland map with sustainable colony starts, separated renewable crop plots, small home ponds, and neutral watering holes that invite expansion. A 128×128 four-colony map now has clear land approaches and buildable home ground; its watering-hole count can fall below the optional Many target when homes and five-tile routes consume the available space.

The generator is catalog ID 45, revision 2. It targets Sparse/Normal/Many neutral ponds, supports 0–20% dry patches and the existing five resource sliders, and accepts supported squares and 2:1 rectangles from 128 tiles upward. Every home keeps minimum renewable wheat and wood plus quarry stone even at zero abundance. Reusable bounded site jitter and contained-plot operations were added to the existing map toolkit, with defaults for existing callers preserved. [Design and numerical choices](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/docs/map-generators/SAVANNAH.md) explain the placement budgets, validation, errors, saturation and tuning.

### Map previews

128×128, four colonies, Normal ponds (seed 23029):

Savannah 128 square

256×128, eight colonies, Many ponds (seed 23105):

Savannah wide rectangle

128×256, eight colonies, Many ponds (seed 23168):

Savannah tall rectangle

512×512, twelve colonies, default controls (seed 23464):

Savannah large square

### Verification and review evidence

- Optimized macOS arm64 build, defaults harness (including 4,096 unattended resource-growth calls), and all 256 golden rows passed. Existing generator hashes stayed unchanged. Translation catalog audit and regression tests pass: all 32 shipped non-English tables have the six new labels, with no missing or untranslated entries.
- Linux x86_64 bulk generation on `pharaoh-dev-3.local` passed **872/872 valid requests**: 130 stress-pilot cases and 742 fresh-seed full cases. The matrix covers every legal dry/pond value pair on the tight square and both rectangle orientations, every individual resource-slider value, worker extremes and 240 independent mixed settings across 15 shapes. No seed was retried. Worst observed starter wheat/wood gathering path was 14 steps; no map lacked the essential neutral pond.
- Neutral pond requests placed 4,021/4,092 targets. **55/742 full-run maps** omitted an optional pond, mostly crowded Many requests; all Sparse targets were placed. Finite high-abundance crop plots saturated on 264 maps, while none of 292 all-default-abundance maps saturated. No parameter or geometry change was needed after this zero-failure bulk run.
- Light populated-game tests exercised Nicowar, Numbi and Maxima. Two matched 24,000-tick games with the final plots retained more wheat and had fewer starvation deaths than the baseline on those two maps; combat occurred in all four compared games. A separate Linux four-colony game still recorded food pressure, mainly for Maxima. These outcomes are seed-specific play evidence, not a general balance guarantee.
- [Bulk audit and reproduction commands](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/artifacts/savannah/bulk/SUMMARY.md), [full raw job](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/artifacts/savannah/bulk/full.tar.gz), [pilot raw job](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/artifacts/savannah/bulk/pilot.tar.gz), [eight saved sample maps](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/artifacts/savannah/bulk/review-sample-maps.zip), [playtest summary](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/artifacts/savannah/playtest-round1/SUMMARY.md), and [one game save/replay package](https://github.com/Globulation2/glob2/blob/574f8d645f5ed9c03651480ce2a8e844a3586255/artifacts/savannah/review-playtest-seed1.zip) are attached in this branch. Per-case requests, reports and logs are inside the raw-job archives; each screenshot has its own request manifest.

### Feel and compatibility limits

The new option creates more open lanes and pushes optional resource competition toward neutral ponds. It does not change simulation rules, save format, replay or network versions. Human play is still needed to judge whether those ponds actually attract conflict and whether flanking feels useful. Windows execution and cross-platform per-tick checksums were not run. A Linux-generated map for seed 3005 differed in tile layout from the same-seed macOS generation; loading that saved Linux map on macOS preserved terrain, resources, starts and movement reports, and Linux game saves loaded on macOS. The 742-case bulk matrix samples mixed controls but cannot enumerate all 13^5 resource combinations.

## Maintainer feedback

The plain read as empty and dry. Revision 2 adds what the concept names: neutral watering holes go one per 128-square higher at every setting (Sparse 2, Normal 3, Many 4), ponds grow from radius 4.5 to 5.5, small pools dot the plain about forty tiles apart, and lone trees (singles and clumps of two or three) stand on the plain wherever the ground's crop growth chance is zero, where the engine's water probe never lets them spread. The containment proof still holds: the shared check admits a tree outside a plot only on such dry ground, and the pools water ground where nothing is planted. Fingerprints regenerated on both platforms.

Savannah 256x256, 4 colonies, revision 2

## Review changes

- Legacy id **45**; the bulk archives and scripts moved to the `evidence/savannah` branch, previews and notes stay; contract check in `test/MapGeneratorContracts.h`; catalog row in the framework doc.

## Rotation tournament

Six 256×256 maps, four colonies, every cyclic team rotation, four Nicowars, 45,000 ticks, on revision 1: even and flat. Every start held about 25 units for the whole game with about 66 wheat harvested per colony, no eliminations and no position bias worth the name; a bigger home pond changed nothing (the AI works little of a sealed plot whatever its size), so that tweak was not kept. Revision 2 on the same protocol: pooled per-start units 27 to 33 (peaks 28 to 34) against 22 to 25, wheat harvested per colony 88 against 66, wood 382 against 310, births 30 against 24, still no elimination in 24 games and a bias of 8 points. The water lifts the economy by about a fifth without changing the map's unhurried character; recorded in the design doc.


genixpro:
Can we sprinkle some trees out on the open plain? Also feeling like there is just too little water for this to be playable - increase that somewhat/

PR #309, opened 2026-09-15
Add Rice Terraces map generator with contour farms and stair defense

genixpro:
A Rice Terraces game now starts on dry hill summits surrounded by concentric, lobed crop and water bands. Sand stairs cut through each hill, summit towers cover their upper approaches, and open valleys carry fruit and an optional river. The new shared contour-farm, placement, tower-cover, and masked crop-rescue operations keep most of the layout logic reusable.

Revision 5 generated map, seed 7

Revision 4 AI playtest at 60,000 ticks

### Validation

- Bulk generation: 706 accepted revision-5 jobs; 356 generated maps, 350 explicit geometry-fit rejections, zero unexpected failures, timeouts, missing reports, or structural violations. All seven revision-4 summit-spill cases now generate with reachable starter wood. [Full job records and analysis](https://github.com/Globulation2/glob2/blob/feature/rice-terraces-generator/docs/map-generators/evidence/rice-terraces/BULK_GENERATION.md).
- Linux x86_64 and macOS arm64: defaults/toolkit contracts passed, 256 golden rows compared on each with zero failures. The revision-5 seed-7 native map bytes match across platforms. Existing generator golden rows are untouched.
- AI playtesting: the tuned revision-4 opening had zero early eliminations in 12 prospective games across 48 colonies; later AI hunger and passive valley play remain balance review points. [Playtest evidence](https://github.com/Globulation2/glob2/blob/feature/rice-terraces-generator/docs/map-generators/evidence/rice-terraces/PLAYTEST.md).
- Six new labels and controls are localized in all 32 language files; five translation regression tests passed.

This optional map changes the feel of its own games: defending stairs and harvesting the terrace bands concentrate play around each summit, while river valleys offer expansion and fruit. Human gameplay review remains useful for tower strength and valley incentives. Windows generation, revision-5 AI games, and cross-platform per-tick simulation checksums were not verified. No save-format, replay, network-version, or core simulation change was made.

## Maintainer feedback

The circles read as a centre-pivot farm. Revision 6 lobes them: the shared contour farm gains `ContourWobble`, three harmonics round each hill with their own phases that shift every band in or out by the same amount at a heading (so band widths hold), ramped in past the summit's cap so the town stays round and the mapping along a ray stays monotone. The generator spends as much of four tiles of lobe as leaves half the valley floor open at the closest pair of hills, so the fitted band count is unchanged, and tells hill ground from valley by the same nominal radius. A landscape check covers the lobed farm. Fingerprints regenerated on both platforms.

Revision 6, seed 7

## Review changes

- Legacy id **46**; the bulk archives moved to the `evidence/rice-terraces` branch; catalog row in the framework doc; toolkit and landscape checks merged with the stacked generators.
- CI gives `MapGeneratorDefaultsTest` two minutes instead of one: with a contract check for every new generator it runs a little over a minute on the runners and was being cut off after its last check had printed.

## Rotation tournament

Six 256×256 maps, four colonies, every cyclic team rotation, four Nicowars, 45,000 ticks, on revision 5: even and hard. Root-mean-square position bias zero, pooled per-start peaks of 124 to 146 units, 20 to 34 warriors, 24 eliminations in 96 colony-games, and 60 to 67 starvation deaths per colony, the highest of any map that still grows: the capped bands are worked (about 145 wheat harvested per colony) but not fast enough for the armies the stairs invite. The lobes move no band's area, so the economy is the same; whether that hunger is the intended feel is a maintainer's judgment, recorded in the design doc.


genixpro:
Can you make this one more wavy and less circular? Its the right idea but real rice terraces aren't perfect circles like this. This seems more like a saudi central pivot farm.

